PR #181, opened 2026-04-11
fix(treesitter): add recursion depth limit to csharp and ruby analyzers

mikelane:
## Summary

Fixes a **stack-overflow DoS vulnerability** in `src/treesitter/csharp.rs::visit_node` and `src/treesitter/ruby.rs::extract_functions_from_node`. Both functions recursed without bounds against arbitrary source code read from git blobs. An attacker committing a deeply-nested file (>~5000 nested namespaces in C# or class/module blocks in Ruby) could crash git-prism when an agent ran `get_change_manifest` on a diff containing that file.

## Changes

- `src/treesitter/mod.rs`: new `pub const MAX_RECURSION_DEPTH: usize = 256;`
- `src/treesitter/csharp.rs`: added `depth: usize` parameter + guard to `visit_node`; new test `deeply_nested_namespaces_do_not_stack_overflow`
- `src/treesitter/ruby.rs`: added `depth: usize` parameter + guard to `extract_functions_from_node`; new test `deeply_nested_classes_do_not_stack_overflow`

## TDD audit trail

- **RED**: Both tests crash with `thread '<unknown>' has overflowed its stack` / SIGABRT on unfixed code at both 512KB and 2MB stack sizes. Verified by commenting out each depth guard, running the test, seeing SIGABRT, restoring the guard.
- **GREEN**: With `MAX_RECURSION_DEPTH = 256` in place, both tests pass cleanly. Functions extracted BEFORE hitting the limit are still returned (partial extraction, not a failed call).
- **REFACTOR**: Shared constant in `mod.rs` enables future analyzer fixes (PHP, Swift, Java, etc.) to use the same value without duplication.

## Security rationale

Tree-sitter is fed UNTRUSTED source code from git blobs. An attacker can commit arbitrary content. The unbounded recursion was a stack-overflow primitive in the most sensitive part of the pipeline: source code parsing of untrusted input during agent-initiated diffs.

**Correction to A4 gauntlet wording**: C# DoS is specifically via nested **namespaces**, not nested classes — `visit_node` stops at `class_declaration` so nested classes alone don't exercise the recursive path. Verified empirically.

## Test plan

- [x] 498/498 tests pass (was 496, +2 new tests)
- [x] `cargo clippy -- -D warnings` clean
- [x] Both new stack-overflow tests verified to FAIL on unfixed code and PASS with fix

## Discovered during Wave 1a

A4 v0.4.0 languages gauntlet review. The other analyzers (PHP, Swift, Java, Kotlin) have similar recursive patterns and will be fixed in Tier 4 — `MAX_RECURSION_DEPTH` is already available for those future fixes.


src/treesitter/ruby.rs
```diff
@@ -358,4 +368,48 @@ end
...
+
+    /// Security regression: deeply-nested class declarations used to stack-overflow
+    /// `extract_functions_from_node` because it recursed into each class body without
+    /// a depth limit. An attacker committing a Ruby file with thousands of nested
+    /// class blocks could crash git-prism during `get_change_manifest`. The analyzer
+    /// must now complete without crashing.
+    ///
+    /// Runs on a thread with a bounded 512KB stack to force the crash to be
+    /// reproducible regardless of the host's default stack size.
+    #[test]
```
mikelane, line 380: MEDIUM — documentation drift: the comment says the test runs on a **512 KB** stack, but the code on line 385 sets `TEST_STACK_SIZE = 2 * 1024 * 1024` (2 MB). Could this docstring be updated to match the actual value, or the stack size reduced to the documented 512 KB (which would also be a stronger proof since it's further below default)?

src/treesitter/ruby.rs
```diff
@@ -358,4 +368,48 @@ end
...
+    fn deeply_nested_classes_do_not_stack_overflow() {
+        const NESTING_DEPTH: usize = 5000;
+        // 2 MB: roomy enough for bounded recursion to `MAX_RECURSION_DEPTH`
+        // but far too small for unbounded recursion to 5000 frames.
+        const TEST_STACK_SIZE: usize = 2 * 1024 * 1024;
```
mikelane, line 385: HIGH (follow-up, not a blocker) — test weakens the security claim: the cargo test harness already defaults to ~2 MB thread stacks, so spawning a fresh `std::thread::Builder` with a 2 MB stack provides essentially no additional constraint. If the PR body's claim is correct that the unfixed code crashes at 2 MB, that's fine — the test reproduces it. But it would be stronger to spawn on a **smaller** stack (e.g., 512 KB as the comment suggests) so the test failure margin is unambiguous and the test remains robust against future Rust stack-size defaults. A note inline: a 2 MB stack on ruby at 5000 unbounded frames is only ~400 bytes/frame — very close to overflowing; on debug builds the frames are larger and would crash, on aggressive release optimizations it might not. Pinning to 512 KB makes the proof independent of build profile.

src/treesitter/csharp.rs
```diff
@@ -374,4 +386,44 @@ public class Foo {}
...
+    /// Uses nested namespaces (not classes) because `visit_node` stops recursing at
+    /// a `class_declaration`, but falls through and recurses into children for any
+    /// other node kind — including `namespace_declaration`.
+    ///
+    /// Runs on a thread with a 2 MB stack: roomy enough for bounded recursion to
+    /// `MAX_RECURSION_DEPTH` but far too small for unbounded recursion to 5000 frames.
+    #[test]
+    fn deeply_nested_namespaces_do_not_stack_overflow() {
+        const NESTING_DEPTH: usize = 5000;
+        const TEST_STACK_SIZE: usize = 2 * 1024 * 1024;
```
mikelane, line 404: NIT — C# test docstring justifies the 2 MB choice slightly better than the Ruby test docstring, but the underlying concern about using the same default-sized stack applies here too. Same suggestion: consider 512 KB for a stronger margin that doesn't depend on the cargo test harness default.

src/treesitter/mod.rs
```diff
@@ -13,6 +13,20 @@ pub mod typescript;
...
+/// namespaces, classes, modules, etc.) must bound their recursion to avoid
+/// stack overflow when processing adversarial input read from git blobs.
+///
+/// An attacker committing a source file with thousands of nested class or
+/// module blocks would otherwise crash git-prism during `get_change_manifest`.
+///
+/// 256 is generous for real-world code — humans rarely nest more than ~20
+/// levels — while keeping worst-case stack usage comfortably under the
+/// default 2 MB cargo test stack and the 8 MB default main thread stack.
+pub const MAX_RECURSION_DEPTH: usize = 256;
```
mikelane, line 28: PRAISE — nice factoring: a single shared `MAX_RECURSION_DEPTH` constant with a thoughtful docstring explaining the ceiling (human nesting ~20 levels, fits under 2 MB test / 8 MB main thread budget) makes this reusable for the follow-up fixes in python, typescript, rust_lang, cpp, and c_lang — which appear to share the same vulnerability (see overall review comment).

mikelane (commented):
## Summary

Tightly scoped security fix for the csharp and ruby analyzers. The `depth: usize` guard is a textbook defensive pattern: passed by value (no shared state, no thread-safety concerns), bounds the recursion cleanly, and preserves partial extraction semantics (functions/classes collected before the limit are still returned). The shared `MAX_RECURSION_DEPTH` constant with its justification docstring is excellent groundwork for the follow-up fixes.

The PR is **READY TO MERGE** as-is. The scope was set strictly to csharp and ruby and the fix achieves its stated goal correctly. The CI signals are green on the checks that matter (Check, Format, Clippy, Test ubuntu/macos).

## What the review verified

- **Fix correctness.** `visit_node` (csharp) and `extract_functions_from_node` (ruby) both correctly return early when `depth >= MAX_RECURSION_DEPTH`. Functions/classes extracted before the limit are retained (partial extraction). The ruby test even asserts on partial extraction behavior (`!functions.is_empty()`).
- **No other recursive path inside csharp.rs or ruby.rs.** `extract_calls` and `extract_imports` in both files use iterative stack traversal, not recursion. `extract_methods_from_class` (csharp) is flat — it iterates class body children only.
- **Thread safety.** `depth: usize` is a copy, no shared state. No races possible.
- **Test location.** Both new tests are correctly colocated with the code they test (csharp.rs, ruby.rs), not in a separate file.
- **PR description correction.** The callout that C# DoS is via nested *namespaces* (not classes) is correct — `visit_node` short-circuits at `class_declaration` and only recurses on other node kinds. The test correctly uses `namespace N{i}` nesting. Good instinct to verify this empirically rather than trust the A4 gauntlet wording.
- **Self-documenting code and comments.** The inline comment on line 70-72 of csharp.rs explains WHY the guard exists (adversarial input) rather than WHAT it does. Clean.
- **Code quality.** `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` all pass on CI (ubuntu + macos). No TODOs, no suppressions, no `unsafe` blocks introduced, no lint exclusions.

## Blockers

None.

## Recommendations

### MEDIUM — docstring drift in ruby.rs test (inline comment posted)

The doc comment on the ruby test says "Runs on a thread with a bounded 512KB stack" but the constant is `2 * 1024 * 1024` (2 MB). Either the doc or the code should change. My preference: reduce the stack to 512 KB (see next point).

### HIGH follow-up (not a blocker) — test stack size weakens the proof

The cargo test harness uses ~2 MB thread stacks by default on most platforms. Spawning a fresh `std::thread::Builder` with `stack_size(2 * 1024 * 1024)` provides essentially no additional constraint over running the test on the default harness thread — the test would "protect" any scenario where the default also protects. The PR body claims the unfixed code crashed at both 512 KB and 2 MB; if so, the test would still reproduce at 512 KB with a bigger safety margin and be robust to future Rust stdlib changes. Recommend dropping the test stack to 512 KB (matching the ruby docstring that already claims that value).

Rough back-of-envelope: 5000 frames × per-frame cost. Each frame in `extract_functions_from_node` carries several `tree_sitter::Node` copies (24 bytes each), a walk cursor, an `Option<&str>`, a `&mut Vec`, plus saved registers and the return address. In debug builds that's often 400-800 bytes/frame; in release it compresses. At 5000 × ~500 bytes = ~2.5 MB — close enough to 2 MB to crash, but the margin is thin. 512 KB makes the test decisive across all build profiles.

### HIGH follow-up — other analyzers with the same unbounded recursion

The PR body claims "php.rs, swift.rs, kotlin.rs, java.rs all recurse through nested declarations without depth guards." A source-read audit shows this claim is **inverted** — those four analyzers are actually **safe** (flat iteration via `extract_methods_from_class` that only walks class body children, no recursion into nested classes). The analyzers that *actually* share the unbounded-recursion pattern are the ones NOT called out:

| Analyzer | Function | Recursive path |
|---|---|---|
| `python.rs` | `extract_functions_from_node` | recurses on `class_definition` -> body |
| `typescript.rs` | `extract_functions_from_node` | recurses on `class_declaration` -> body (TS + TSX + JS) |
| `rust_lang.rs` | `extract_functions_from_node` | recurses on `impl_item` -> body |
| `cpp.rs` | `collect_functions` | recurses on `namespace_definition`, `class_specifier`/`struct_specifier`, AND `preproc_if*` containers |
| `cpp.rs` | `collect_imports` | recurses on `preproc_if*` containers |
| `c_lang.rs` | `collect_functions` | recurses on `preproc_if*` containers |
| `c_lang.rs` | `collect_imports` | recurses on `preproc_if*` containers |

Python and TypeScript are the most urgent since they're top-tier languages in git-prism's audience and the recursion is triggered by ordinary class nesting (which real code does reach ~5-10 levels, making pathological input plausible). The C/C++ cases are especially interesting because adversarial `#ifdef` nesting is a known hostile-C idiom.

Filing a Tier 4 follow-up issue for those seven paths is recommended — the shared `MAX_RECURSION_DEPTH` constant is already in place, so each fix should be ~10 lines + one regression test.

### LOW — use a loop counter instead of recursion (speculative)

All of these analyzers could alternatively be converted to iterative stack-based traversal (like `extract_calls` already uses) — that would eliminate the depth guard entirely and be strictly more robust. Not a blocker for this PR, but worth considering for Tier 4.

## Praise

- Excellent TDD evidence in the PR body: explicit RED-proven by commenting out each depth guard, witnessing SIGABRT, then re-enabling. This is exactly how security fixes should be verified.
- The shared constant in `mod.rs` with its justification (generous ceiling, stack budget math) is a nice piece of forward thinking.
- Corrected the A4 gauntlet's loose wording about "nested classes" vs "nested namespaces" — verifying the actual AST path rather than trusting upstream gauntlet language is exactly right.
- Partial-extraction semantics are correct: hit the ceiling, return what you've got, don't drop the whole result. The ruby test asserts on this explicitly.
- Zero suppressions, zero `unsafe`, zero lint exclusions, zero pre-commit bypasses.

## Decision

**COMMENT** (owner-authored PR; GitHub disallows owner self-approval). The review verdict is effectively **READY TO MERGE** — no blockers, two recommendations (one MEDIUM doc fix, one HIGH follow-up issue to file for the other five analyzers). The scope of this PR is correct and should not be expanded.

---

Generated by pr-reviewer.

PR #145, opened 2026-04-10
Epic #126: Function context data quality + blast radius scoring

mikelane:
## Summary

Closes epic #126. Fixes silent data loss in tree-sitter extraction and adds blast radius scoring to `get_function_context`.

**Wrapper extraction fixes** (closes #129, #130, #131):
- TypeScript/JS: recurse into `export_statement` — `export function foo()` and `export class Foo {}` now produce `FunctionChange` entries
- Python: recurse into `decorated_definition` — `@app.route def index()` now detected (ubiquitous in Flask/FastAPI/pytest)
- C++: recurse into `linkage_specification` — functions inside `extern "C"` blocks now extracted

**Blast radius scoring** (closes #132, #133):
- New `BlastRadius` type with `production_callers`, `test_callers`, `has_tests`, `risk` fields
- `RiskLevel` enum: `none` / `low` / `medium` / `high` (bounded, not numeric)
- Risk classification: 0 callers → none; 1–2 with tests → low; 1–2 no tests → medium; 3+ with tests → medium; 3+ no tests → high
- Computed for every `FunctionContextEntry`

**BDD** (closes #128): 12 new Gherkin scenarios (all RED at bootstrap, all GREEN after implementation).

**Docs** (closes #134): CLAUDE.md conventions + README example output updated.

Capstone demo (#135) left open per user request.

## Test plan

- [x] 22 new unit tests for wrapper extraction across TS/Python/C++
- [x] 8 unit tests for `BlastRadius::compute` covering all risk classification rows
- [x] 12 new BDD scenarios pass
- [x] All 78 existing BDD scenarios still pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `cargo test` — 510 pass (1 pre-existing failure in `tools::history::tests::it_populates_commit_metadata_with_author_and_timestamp` unrelated to this change)

https://claude.ai/code/session_01MpqdLPUoYYTAKtuFAhTRor

PR #146, opened 2026-04-10
Epic #127: Import-aware caller scoping

mikelane:
## Summary

Closes epic #127. Replaces the brute-force caller scan in `get_function_context` with import-aware scoping, eliminating cross-module false positives and improving performance on large repos.

**Stacked on #126** — this PR branches from `epic/126-data-quality-blast-radius`. Merge #126 first, then this PR will target `main` automatically.

**Spike** (closes #136): [ADR 0006](docs/decisions/0006-import-resolution-spike.md) validates import resolution feasibility for Rust, Python, Go, and TypeScript. All four languages have sufficient import data. Go and Rust are high confidence (explicit paths); Python and TS are medium (relative imports need heuristics, path aliases need fallback).

**BDD bootstrap** (closes #137): 7 new Gherkin scenarios. 3 were RED at bootstrap (the core exclusion tests), 4 tested backward compatibility. All green after implementation.

**Implementation** (closes #138, #139, #140, #141):
- New `src/tools/import_scope.rs` module with language-specific logic
- **Module path inference**: `src/foo/bar.rs` → `crate::foo::bar`, `src/validation.py` → `src.validation`, `lib/lib.go` → `lib` (directory), `lib.ts` → `lib`
- **Import target matching**: each language parses import strings and matches against inferred module paths; `super::`/`self::` in Rust conservatively included
- **Filtered scan**: changed files always scanned; Go uses same-package semantics; other languages only scan files that import changed modules
- **Fallback**: unsupported languages (Ruby, C, Java, PHP, C#, Swift, Kotlin) fall back to full-repo scan automatically

**Docs** (closes #143): CLAUDE.md + README updated.

Performance benchmarking (#142) deferred — architecture change guarantees improvement; formal benchmarking would require a large fixture repo. Capstone demo (#144) left open per user request.

## Test plan

- [x] 29 new unit tests in `import_scope` module covering module inference and import matching per language
- [x] 7 new BDD scenarios pass (import exclusion + backward compat)
- [x] All 85 total BDD scenarios pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `cargo test` — 539 pass (1 pre-existing failure unrelated)
- [x] Manual smoke test: multi-file Rust repo with importer and non-importer confirms non-importer is excluded from callers

## Semantic change note

[...]

PR #203, opened 2026-04-11
ci+hooks+test+docs: project integrity safeguards from retrospective (#202)

### LINKED ISSUE #202: Project integrity safeguards: retrospective fixes from observability remediation
## Context

After the observability remediation work in April 2026 (Wave 1a gauntlet reviews + Tier 1-3 merge blitz), a retrospective identified several safeguards that would have prevented the issues we had to fix. This issue tracks the implementation of those safeguards as a single coordinated PR.

See the conversation context for the full retrospective. Summary of what broke:

1. **PR #125 was a mis-labeled squash merge** — 30 files and 432 Rust LOC landed under the title "docs: add TDD rules" with fictional PR references (#114-#123) that never existed.
2. **v0.5.0 and v0.6.0 were never shipped** — Cargo.toml was bumped but no git tag / GitHub Release / crates.io publish existed. Users got v0.4.0.
3. **`get_function_context` was claimed as an MCP tool** in CHANGELOG/README/CLAUDE.md for multiple releases without actually being registered in `src/server.rs`.
4. **Mutation testing shard 4/4 silently broken** for multiple releases due to `|| true` masking an off-by-one error — the 90% threshold was computed over 75% of the surface.
5. **Pre-existing clippy --all-targets failure** hidden by CI and lefthook both omitting `--all-targets`.
6. **`GIT_PRISM_OTLP_HEADERS` documented but not wired** — literal `TODO` in code with an advertised SigNoz auth example.
7. **9 telemetry BDD scenarios permanently tagged `@not_implemented`** and excluded from CI.
8. **Dev agents died mid-task** committing but not pushing, losing work multiple times.

## Scope

Single PR, single branch (`integrity-safeguards`), landing all of the following:

### A. CI integrity workflow (`.github/workflows/integrity.yml`)

New workflow that runs on every PR and push to main, enforcing:

1. **Cargo.toml version ↔ git tag sync**: fail if `Cargo.toml` version doesn't match a real git tag (with an allowlist for in-progress release PRs).
2. **README MCP tool count sync**: parse `N MCP tools` from README and compare to `grep -c '#\[tool(' src/server.rs`. Fail on mismatch.
3. **`|| true` ban in workflow files**: grep all `.github/workflows/*.yml` (except integrity.yml itself) for `|| true` and fail.
4. **`--all-targets` enforcement**: grep `ci.yml`, `lefthook.yml`, `CLAUDE.md` for any `cargo clippy` invocation without `--all-targets`. Fail on miss.
5. **`@not_implemented` audit on PRs that close issues**: for each `Closes #N` in the PR body, grep `bdd/features/` for `@not_implemented @ISSUE-N`. Fail on match.
6. **CHANGELOG section ↔ tag sync**: every `## [X.Y.Z]` heading in CHANGELOG.md must have a matching git tag (allow `## [Unreleased]`).

### B. Lefthook commit-msg hook: validate (#N) references

New `scripts/validate-pr-refs.sh` that reads a commit message, extracts all `(#N)` references, and checks each against `gh pr view N` / `gh issue view N`. If any reference a number that doesn't exist, fail the commit.

Wired into `lefthook.yml` under `commit-msg`.

### C. `src/server.rs` EXPECTED_TOOLS test

New unit test in `src/server.rs::tests` that compares `GitPrismServer::tool_router().list_all()` to a hardcoded `EXPECTED_TOOLS` constant. Any drift (new tool not in list, old tool removed, order changed) fails the test with a clear error message. This makes the MCP tool set a compile-time contract.

### D. CLAUDE.md Release Hygiene section

Add a new section to `CLAUDE.md` (the project one, not global) covering:

- Release PRs are incomplete until tag pushed, workflow succeeded, GH Release exists, crates.io publish done, homebrew tap updated
- `|| true` banned in workflow files
- Clippy must always use `--all-targets`
- CHANGELOG sections must match tags
- Reference the global `releasing-to-production` skill

Also add a Dev Agent Discipline pointer to the global `~/.claude/CLAUDE.md` Dev Agent Discipline section.

## Acceptance criteria

- [ ] All 6 CI integrity checks fire and fail on intentional violations (verified with deliberate-bad-state test)
- [ ] `scripts/validate-pr-refs.sh` rejects a commit message containing `(#999999)` (doesn't exist)
- [ ] `scripts/validate-pr-refs.sh` accepts a commit message containing `(#145)` (real PR)
- [ ] `src/server.rs::tests::it_matches_expected_tools` passes
- [ ] Intentionally removing a tool from `src/server.rs` causes `it_matches_expected_tools` to fail with a clear message
- [ ] `cargo fmt --check` clean
- [ ] `cargo test` passes
- [ ] lefthook pre-push passes
- [ ] PR body describes the retrospective context and links to this issue

mikelane:
## Summary

Implements the integrity safeguards identified in the post-remediation retrospective. Four components landing atomically as one PR: a CI integrity workflow, a commit-msg hook validating `(#N)` references, an `EXPECTED_TOOLS` test enforcing MCP tool drift detection, and a new `Release Hygiene` section in `CLAUDE.md`.

These safeguards would have prevented (or at least caught) every major incident from the April 2026 remediation session:

- **PR #125 mis-labeled squash merge** with fictional PR refs `#114-#123` → Component B's commit-msg hook rejects phantom `(#N)` references at commit time.
- **v0.5.0 / v0.6.0 never tagged but Cargo.toml bumped** → Component A's Check 1 fails any main commit where `Cargo.toml` version has no matching git tag.
- **`get_function_context` claimed as MCP tool for multiple releases without registration** → Component C's `EXPECTED_TOOLS` test makes the MCP tool set a compile-time contract.
- **Mutation shard `|| true` masked a broken shard for multiple releases** → Component A's Check 3 greps workflow files for `|| true` and fails.
- **Pre-existing clippy `--all-targets` failure hidden by CI config** → Component A's Check 4 enforces `--all-targets` in all clippy invocations.
- **9 telemetry BDD scenarios permanently tagged `@not_implemented` in closed issues' PRs** → Component A's Check 5 audits `@not_implemented @ISSUE-N` against `Closes #N` in PR body.
- **CHANGELOG drift where `## [0.6.0]` existed with no tag** → Component A's Check 6 validates every CHANGELOG heading has a matching tag.

See issue #202 for the full retrospective.

## Component breakdown

| Component | Commit | Files |
|---|---|---|
| A — CI integrity workflow | `91ff274` | `.github/workflows/integrity.yml` (new, 158 lines, 6 independent check steps) |
| B — commit-msg phantom-ref validator | `3efeb81` | `scripts/validate-pr-refs.sh` (new, executable), `lefthook.yml` (new `commit-msg` section) |
| C — `EXPECTED_TOOLS` compile-time contract | `6275dd1` | `src/server.rs` (new constant + test) |
| D — CLAUDE.md Release Hygiene section | `f3869bb` | `CLAUDE.md` (new section just before `## Git Hooks`) |

## TDD audit trail

**Component B** — agent tested the script with three cases: `(#999999)` rejected, `(#145)` accepted (real merged PR), `(#202)` accepted (real issue). During testing the agent caught a real bug: `gh pr view N --json number` echoes the input number regardless of whether the PR exists. Fixed to use `--json state` which forces gh to resolve against the API.

**Component C** — strict RED → GREEN → TRIANGULATE → GREEN:
- RED: test added before the constant → compile error `cannot find value EXPECTED_TOOLS in this scope`
- GREEN: constant added → test passes
- TRIANGULATE: remove `"get_function_context"` from constant → test fails with clear drift message referencing README/CHANGELOG/CLAUDE.md
- GREEN: restored → passing

## Test plan

- [x] `cargo test` — 572 passed, 0 failed (was 502 before; +70 from the various Tier 1-5 work merged this week)
- [x] lefthook pre-push (clippy + fmt + test) passed on every push
- [x] lefthook commit-msg `validate-pr-refs` hook validated every commit message on this branch
- [x] Component A's Check 4 corner case verified: the new Release Hygiene section in CLAUDE.md mentions `cargo clippy` in prose, but the same line also has `--all-targets`, so the grep filter correctly doesn't flag it

## Security note

Component A uses `env:` to pass `PR_TITLE` into the integrity workflow rather than `${{ }}` interpolation inside `run:` blocks — this avoids the common GitHub Actions workflow injection attack vector.

## Corner cases documented

- **Check 5** (`@not_implemented` audit) will print "no Closes/Fixes/Resolves references, skipping" if this PR body doesn't reference `#202`. The PR body above includes `Closes #202` so it will fire and validate that `bdd/features/` has no `@not_implemented @ISSUE-202` (there are no BDD scenarios for this PR, so it passes).
- **Check 1** (Cargo.toml tag sync) currently passes because `Cargo.toml` version `0.6.0` matches the `v0.6.0` tag that we pushed yesterday.

Closes #202

### DIFF
--- a/.github/workflows/integrity.yml
+++ b/.github/workflows/integrity.yml
@@ -1,0 +1,157 @@
+name: Integrity
+
+# Enforces release hygiene and drift checks that catch mismatches between
+# Cargo.toml / git tags / CHANGELOG / README / workflow files. Introduced
+# as part of issue #202 after an April 2026 retrospective found six classes
+# of silent drift that had shipped multiple releases undetected.
+#
+# Security: any input sourced from PR metadata (title, body, etc.) is passed
+# via env vars rather than ${{ }} expansion directly in run blocks. See
+# https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  integrity:
+    name: Integrity checks
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    env:
+      PR_TITLE: ${{ github.event.pull_request.title }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Cargo.toml version matches a git tag
+        run: |
+          set -euo pipefail
+          VERSION=$(grep -m1 '^version' Cargo.toml | sed 's/.*"\(.*\)".*/\1/')
+          if git rev-parse "v$VERSION" >/dev/null 2>&1; then
+            echo "✓ v$VERSION tag exists"
+          else
+            if [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ] && echo "${PR_TITLE:-}" | grep -qi '^chore.*release'; then
+              echo "⚠ v$VERSION has no tag (release PR in flight — OK)"
+            else
+              echo "✗ Cargo.toml version $VERSION has no matching git tag"
+              exit 1
+            fi
+          fi
+
+      - name: README MCP tool count matches src/server.rs
+        run: |
+          set -euo pipefail
+          README_COUNT=$(grep -oE '[0-9]+ MCP tools?' README.md | head -1 | grep -oE '[0-9]+' || echo 0)
+          # The README uses words like "Four MCP tools" — fall back to word-to-digit mapping.
+          if [ "$README_COUNT" = "0" ]; then
+            README_WORD=$(grep -oE '(One|Two|Three|Four|Five|Six|Seven|Eight|Nine|Ten) MCP tools?' README.md | head -1 | awk '{print $1}' || true)
+            case "$README_WORD" in
+              One) README_COUNT=1 ;;
+              Two) README_COUNT=2 ;;
+              Three) README_COUNT=3 ;;
+              Four) README_COUNT=4 ;;
+              Five) README_COUNT=5 ;;
+              Six) README_COUNT=6 ;;
+              Seven) README_COUNT=7 ;;
+              Eight) README_COUNT=8 ;;
+              Nine) README_COUNT=9 ;;
+              Ten) README_COUNT=10 ;;
+              *) README_COUNT=0 ;;
+            esac
+          fi
+          CODE_COUNT=$(grep -c '#\[tool(' src/server.rs || echo 0)
+          if [ "$README_COUNT" = "$CODE_COUNT" ]; then
+            echo "✓ README claims $README_COUNT MCP tools, src/server.rs has $CODE_COUNT"
+          else
+            echo "✗ README claims $README_COUNT MCP tools but src/server.rs has $CODE_COUNT"
+            exit 1
+          fi
+
+      - name: No '|| true' in workflow files
+        run: |
+          set -euo pipefail
+          HITS=$(grep -rn '|| true' .github/workflows/ | grep -v 'integrity.yml:' || true)
+          if [ -z "$HITS" ]; then
+            echo "✓ no || true masking in workflow files"
+          else
+            echo "✗ || true found in workflow files (silently masks errors):"
+            echo "$HITS"
+            exit 1
+          fi
+
+      - name: cargo clippy invocations include --all-targets
+        run: |
+          set -euo pipefail
+          FILES_TO_CHECK=".github/workflows/ci.yml lefthook.yml CLAUDE.md"
+          BAD=""
+          for f in $FILES_TO_CHECK; do
+            BAD_LINES=$(grep -n 'cargo clippy' "$f" | grep -v -- '--all-targets' || true)
+            if [ -n "$BAD_LINES" ]; then
+              BAD="$BAD\n$f:\n$BAD_LINES"
+            fi
+          done
+          if [ -z "$BAD" ]; then
+            echo "✓ all cargo clippy invocations use --all-targets"
+          else
+            printf "✗ cargo clippy without --all-targets:%s\n" "$BAD"
+            exit 1
+          fi
+
+      - name: '@not_implemented scenarios for closed issues'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          if [ "${GITHUB_EVENT_NAME:-}" != "pull_request" ]; then
+            echo "⚠ not a PR event, skipping @not_implemented audit"
+            exit 0
+          fi
+          PR_NUM="${GITHUB_EVENT_PATH:+$(jq -r '.pull_request.number // empty' < "$GITHUB_EVENT_PATH")}"
+          if [ -z "$PR_NUM" ]; then
+            echo "⚠ no PR number, skipping"
+            exit 0
+          fi
+          BODY=$(gh pr view "$PR_NUM" --json body --jq .body)
+          ISSUES=$(echo "$BODY" | grep -oE '(Closes|Fixes|Resolves) #[0-9]+' | grep -oE '[0-9]+' || true)
+          if [ -z "$ISSUES" ]; then
+            echo "⚠ no Closes/Fixes/Resolves references, skipping"
+            exit 0
+          fi
+          BAD=""
+          for ISSUE in $ISSUES; do
+            HITS=$(grep -rn "@ISSUE-$ISSUE" bdd/features/ 2>/dev/null | grep '@not_implemented' || true)
+            if [ -n "$HITS" ]; then
+              BAD="$BAD\n$HITS"
+            fi
+          done
+          if [ -z "$BAD" ]; then
+            echo "✓ no @not_implemented scenarios for closed issues"
+          else
+            printf "✗ PR closes issues but their scenarios are still @not_implemented:%s\n" "$BAD"
+            exit 1
+          fi
+
+      - name: CHANGELOG sections have matching git tags
+        run: |
+          set -euo pipefail
+          git fetch --tags 2>/dev/null || true
+          MISSING=""
+          while IFS= read -r VER; do
+            if [ "$VER" = "Unreleased" ]; then continue; fi
+            if ! git rev-parse "v$VER" >/dev/null 2>&1; then
+              MISSING="$MISSING $VER"
+            fi
+          done < <(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md | sed 's/^## \[\(.*\)\]$/\1/')
+          if [ -z "$MISSING" ]; then
+            echo "✓ all CHANGELOG versions have matching tags"
+          else
+            echo "✗ CHANGELOG sections without matching git tags:$MISSING"
+            exit 1
+          fi
--- a/.github/workflows/integrity.yml
+++ b/.github/workflows/integrity.yml
@@ -31,13 +31,17 @@ jobs:
           fetch-tags: true
 
       - name: Cargo.toml version matches a git tag
+        if: ${{ !cancelled() }}
         run: |
           set -euo pipefail
           VERSION=$(grep -m1 '^version' Cargo.toml | sed 's/.*"\(.*\)".*/\1/')
           if git rev-parse "v$VERSION" >/dev/null 2>&1; then
             echo "✓ v$VERSION tag exists"
           else
-            if [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ] && echo "${PR_TITLE:-}" | grep -qi '^chore.*release'; then
+            # Narrow allowlist: `chore: release vX.Y.Z` or `chore(release): X.Y.Z`
+            # with an actual version number after `release`. Prevents "chore: release the hounds"
+            # from bypassing the check.
+            if [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ] && echo "${PR_TITLE:-}" | grep -qiE '^chore(\([^)]*\))?:[[:space:]]+release[[:space:]]+v?[0-9]+\.[0-9]+\.[0-9]+'; then
               echo "⚠ v$VERSION has no tag (release PR in flight — OK)"
             else
               echo "✗ Cargo.toml version $VERSION has no matching git tag"
@@ -46,6 +50,7 @@ jobs:
           fi
 
       - name: README MCP tool count matches src/server.rs
+        if: ${{ !cancelled() }}
         run: |
           set -euo pipefail
           README_COUNT=$(grep -oE '[0-9]+ MCP tools?' README.md | head -1 | grep -oE '[0-9]+' || echo 0)
@@ -74,27 +79,32 @@ jobs:
             exit 1
           fi
 
-      - name: No '|| true' in workflow files
+      - name: Workflow files do not mask errors
+        if: ${{ !cancelled() }}
         run: |
           set -euo pipefail
-          HITS=$(grep -rn '|| true' .github/workflows/ | grep -v 'integrity.yml:' || true)
-          if [ -z "$HITS" ]; then
-            echo "✓ no || true masking in workflow files"
-          else
-            echo "✗ || true found in workflow files (silently masks errors):"
-            echo "$HITS"
+          # Comment-aware: the pattern matches only lines whose prefix contains no '#',
+          # so YAML and shell comments (including backticked references in explanatory
+          # text) do not trip the check. integrity.yml is excluded to avoid self-matching
+          # on its own pattern literal — this file is hand-audited.
+          if grep -rnE '^[^#]*\|\| true' .github/workflows/ --exclude=integrity.yml; then
+            echo ""
+            echo "forbidden masking pattern found in workflow files"
+            echo "Use 'continue-on-error: true' on the step, or an explicit 'if !' check."
             exit 1
           fi
+          echo "no masking pattern in workflow files"
 
       - name: cargo clippy invocations include --all-targets
+        if: ${{ !cancelled() }}
         run: |
           set -euo pipefail
           FILES_TO_CHECK=".github/workflows/ci.yml lefthook.yml CLAUDE.md"
           BAD=""
           for f in $FILES_TO_CHECK; do
-            BAD_LINES=$(grep -n 'cargo clippy' "$f" | grep -v -- '--all-targets' || true)
-            if [ -n "$BAD_LINES" ]; then
-              BAD="$BAD\n$f:\n$BAD_LINES"
+            # `if` context suppresses pipefail so no-match exits do not abort the script.
+            if BAD_LINES=$(grep -n 'cargo clippy' "$f" | grep -v -- '--all-targets'); then
+              BAD+=$'\n'"${f}:"$'\n'"${BAD_LINES}"
             fi
           done
           if [ -z "$BAD" ]; then
@@ -105,6 +115,7 @@ jobs:
           fi
 
       - name: '@not_implemented scenarios for closed issues'
+        if: ${{ !cancelled() }}
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
@@ -119,16 +130,19 @@ jobs:
             exit 0
           fi
           BODY=$(gh pr view "$PR_NUM" --json body --jq .body)
-          ISSUES=$(echo "$BODY" | grep -oE '(Closes|Fixes|Resolves) #[0-9]+' | grep -oE '[0-9]+' || true)
+          # Case-insensitive, handles `Closes/closes/close/closed`, `Fix/fixes/fixed`, `Resolves/resolved`.
+          ISSUES=""
+          if MATCHES=$(echo "$BODY" | grep -oiE '(close[sd]?|fix(e[sd])?|resolve[sd]?) #[0-9]+'); then
+            ISSUES=$(echo "$MATCHES" | grep -oE '[0-9]+')
+          fi
           if [ -z "$ISSUES" ]; then
             echo "⚠ no Closes/Fixes/Resolves references, skipping"
             exit 0
           fi
           BAD=""
           for ISSUE in $ISSUES; do
-            HITS=$(grep -rn "@ISSUE-$ISSUE" bdd/features/ 2>/dev/null | grep '@not_implemented' || true)
-            if [ -n "$HITS" ]; then
-              BAD="$BAD\n$HITS"
+            if HITS=$(grep -rn "@ISSUE-$ISSUE" bdd/features/ 2>/dev/null | grep '@not_implemented'); then
+              BAD+=$'\n'"${HITS}"
             fi
           done
           if [ -z "$BAD" ]; then
@@ -139,16 +153,18 @@ jobs:
           fi
 
       - name: CHANGELOG sections have matching git tags
+        if: ${{ !cancelled() }}
         run: |
           set -euo pipefail
           git fetch --tags 2>/dev/null || true
           MISSING=""
+          # Pattern allows pre-release and build suffixes: `0.7.0`, `0.7.0-rc.1`, `0.7.0+build.5`.
           while IFS= read -r VER; do
             if [ "$VER" = "Unreleased" ]; then continue; fi
             if ! git rev-parse "v$VER" >/dev/null 2>&1; then
               MISSING="$MISSING $VER"
             fi
-          done < <(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md | sed 's/^## \[\(.*\)\]$/\1/')
+          done < <(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+[^]]*\]' CHANGELOG.md | sed 's/^## \[\(.*\)\]$/\1/')
           if [ -z "$MISSING" ]; then
             echo "✓ all CHANGELOG versions have matching tags"
           else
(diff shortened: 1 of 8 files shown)


.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+            echo "✓ README claims $README_COUNT MCP tools, src/server.rs has $CODE_COUNT"
+          else
+            echo "✗ README claims $README_COUNT MCP tools but src/server.rs has $CODE_COUNT"
+            exit 1
+          fi
+
+      - name: No '|| true' in workflow files
+        run: |
+          set -euo pipefail
+          HITS=$(grep -rn '|| true' .github/workflows/ | grep -v 'integrity.yml:' || true)
```
mikelane: **CRITICAL BLOCKER (observed failing in CI run 24288651484):** this check is currently grepping for the literal string `|| true` anywhere in workflow files — including inside comments that merely describe the policy. The follow-up commit `68314dd` added explanatory comments to `mutants.yml` and `release.yml` that contain `\`|| true\`` in backticks, and the check is now flagging its own documentation:

```
.github/workflows/mutants.yml:73:        # continue-on-error rather than `|| true` so the failure is visible
.github/workflows/mutants.yml:75:        # bans `|| true` in workflow files because of incidents
.github/workflows/release.yml:322:            # than masking with `|| true` — the integrity workflow bans
```

As a result the PR cannot merge and subsequent checks (4 clippy, 5 @not_implemented, 6 CHANGELOG) are **all skipped** — the run shows only one failing step while the other three were silently marked `skipped` by GitHub Actions' default sequential-step behavior.

Two things want fixing on the same check:

1. Narrow the pattern so comments are excluded. E.g., `grep -rnE '^[^#]*\|\| true'`, or restrict to lines that start with `run:`/are not preceded by a `#`.
2. Make the step more informative: either match on shell tokens (`grep -rnE '\|\|[[:space:]]+true[[:space:]]*($|#)'`) so trailing context after `true` is allowed, or preferably use a real YAML parser via `yq` that only looks at `run:` string values.

Also worth noting: the current exclusion uses substring `integrity.yml:` which would spuriously exclude any other file whose `grep -rn` output happens to contain that literal, e.g. a future `integrity.yml.bak`. Prefer `grep -v '^\.github/workflows/integrity\.yml:'` anchored to the start of the line.

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+      pull-requests: read
+    env:
+      PR_TITLE: ${{ github.event.pull_request.title }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Cargo.toml version matches a git tag
```
mikelane, line 33: **HIGH: steps are sequential with no `if: always()`, so only the first failure surfaces.** The commit message for `91ff274` says "Each check is a separate step so failures are individually identifiable in the run summary," but with GitHub Actions' default behavior a failing step halts the job and all subsequent steps are marked `skipped`. The CI run for this PR shows exactly that:

```
3. Cargo.toml version matches a git tag      ✓ success
4. README MCP tool count matches src/server  ✓ success
5. No '|| true' in workflow files             ✗ failure
6. cargo clippy invocations --all-targets    ⊘ skipped
7. @not_implemented scenarios                 ⊘ skipped
8. CHANGELOG sections matching git tags      ⊘ skipped
```

Consider one of:

1. Add `if: ${{ !cancelled() }}` (or `if: always()`) to each post-first step so every check runs regardless of earlier failure. This preserves the single-job model.
2. Split into six independent jobs (each with its own `runs-on: ubuntu-latest`). This gives you a clean job matrix in the run summary and parallelism, at the cost of 6x checkouts.
3. Collapse into a single step that runs all six checks sequentially but never `exit`s until the end, accumulating failures. Cheapest to implement, loses granular step badges.

Option 1 is the minimal fix and matches the commit message's stated intent.

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+        run: |
+          set -euo pipefail
+          VERSION=$(grep -m1 '^version' Cargo.toml | sed 's/.*"\(.*\)".*/\1/')
+          if git rev-parse "v$VERSION" >/dev/null 2>&1; then
+            echo "✓ v$VERSION tag exists"
+          else
+            if [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ] && echo "${PR_TITLE:-}" | grep -qi '^chore.*release'; then
```
mikelane: **HIGH: allowlist for release PRs is overly broad.** The pattern `^chore.*release` is case-insensitive and greedy, so it matches:

- ✅ `chore: release 0.7.0` (intended)
- ✅ `chore(release): 0.7.0` (intended)
- ⚠ `chore: release the hounds` (unintended — any PR title starting with `chore` that happens to contain the word "release" anywhere bypasses the tag-sync check)
- ⚠ `chore: refactor release module` (unintended — bypasses the check)

Consider a tighter pattern:

```bash
echo "${PR_TITLE:-}" | grep -qE '^chore(\([^)]*\))?:[[:space:]]*release[[:space:]]+v?[0-9]+\.[0-9]+\.[0-9]+'
```

This requires either `chore:` or `chore(<scope>):` followed by `release` followed by a version number. Matches `chore: release 0.7.0`, `chore(release): 0.7.0`, `chore: release v1.2.3`; does not match the bypasses above.

This is the only escape-hatch on an otherwise strict check, and it's the kind of thing that will silently rot if not caught now.

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+            echo "⚠ not a PR event, skipping @not_implemented audit"
+            exit 0
+          fi
+          PR_NUM="${GITHUB_EVENT_PATH:+$(jq -r '.pull_request.number // empty' < "$GITHUB_EVENT_PATH")}"
+          if [ -z "$PR_NUM" ]; then
+            echo "⚠ no PR number, skipping"
+            exit 0
+          fi
+          BODY=$(gh pr view "$PR_NUM" --json body --jq .body)
+          ISSUES=$(echo "$BODY" | grep -oE '(Closes|Fixes|Resolves) #[0-9]+' | grep -oE '[0-9]+' || true)
```
mikelane: **MEDIUM: Check 5 is case-sensitive and only catches the `#N` short form.** The pattern `(Closes|Fixes|Resolves) #[0-9]+` will miss:

- `closes #42` (lowercase)
- `closes: #42` (with colon — some Conventional Commits styles)
- `Closes https://github.com/mikelane/git-prism/issues/42` (full URL form, which GitHub itself accepts for auto-close)
- `Closes: #42`, `Fix #42` (only `Fixes` matches, not `Fix` or `Fixed`)

For agent-authored PRs the current code happens to work because the bodies use `Closes #N` but hand-written bodies vary a lot. Consider:

```bash
# Case-insensitive plus the URL form
ISSUES=$(echo "$BODY" \
  | grep -oiE '(close[sd]?|fix(e[sd])?|resolve[sd]?)[[:space:]:]+(#|https?://[^[:space:]]+/issues/)([0-9]+)' \
  | grep -oE '[0-9]+$')
```

This is the set GitHub itself recognizes per their docs ("closing keywords": close/closes/closed/fix/fixed/fixes/resolve/resolves/resolved).

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+        run: |
+          set -euo pipefail
+          git fetch --tags 2>/dev/null || true
+          MISSING=""
+          while IFS= read -r VER; do
+            if [ "$VER" = "Unreleased" ]; then continue; fi
+            if ! git rev-parse "v$VER" >/dev/null 2>&1; then
+              MISSING="$MISSING $VER"
+            fi
+          done < <(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md | sed 's/^## \[\(.*\)\]$/\1/')
```
mikelane: **MEDIUM: Check 6 won't match pre-release CHANGELOG headings.** The regex `^## \[[0-9]+\.[0-9]+\.[0-9]+\]` matches strict X.Y.Z. When the first pre-release is cut (e.g. `## [0.7.0-rc.1]` or `## [1.0.0-alpha]`) this check will silently skip that heading rather than flagging it as missing a tag.

Also: the regex requires `]` immediately after the version, but the current CHANGELOG uses `## [0.6.0] — 2026-04-09` — the `]` is present so the anchor does match, but the sed `s/^## \[\(.*\)\]$/\1/` uses `$` end-of-line which will FAIL to strip if there's any trailing content after `]`. Double-check this by running on the current CHANGELOG:

```bash
grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md | sed 's/^## \[\(.*\)\]$/\1/'
```

This should output version numbers alone (e.g. `0.6.0`). Because `grep -oE` truncates to just the matched substring, the `$` in the sed anchor matches the END of the truncated match (not the original line), so it does work — but it's fragile.

Consider: `grep -oE '^## \[[^]]+\]' CHANGELOG.md | sed 's/^## \[\([^]]*\)\]/\1/' | grep -Ev '^(Unreleased)$'` to handle pre-release strings naturally.

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+#
+# Security: any input sourced from PR metadata (title, body, etc.) is passed
+# via env vars rather than ${{ }} expansion directly in run blocks. See
+# https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
mikelane, line 16: **MEDIUM: on-trigger only fires for PRs targeting `main`, so stacked PRs escape the check.** This is the same class of gap that PR #146 hit — when a PR is based on another PR branch (rather than main), `pull_request: branches: [main]` does not fire. The whole integrity workflow is silently skipped for stacked work.

Two options:

1. Drop the branch filter entirely so any PR runs the check:
   ```yaml
   on:
     push:
       branches: [main]
     pull_request:
   ```
2. Enumerate explicit long-lived branches (`[main, release/**, epic/**]`).

Given the retrospective explicitly called out the stacked-PR CI gap, leaving this default is a partial fix at best. Recommend option 1.

.github/workflows/integrity.yml
```diff
@@ -0,0 +1,157 @@
...
+
+      - name: cargo clippy invocations include --all-targets
+        run: |
+          set -euo pipefail
+          FILES_TO_CHECK=".github/workflows/ci.yml lefthook.yml CLAUDE.md"
+          BAD=""
+          for f in $FILES_TO_CHECK; do
+            BAD_LINES=$(grep -n 'cargo clippy' "$f" | grep -v -- '--all-targets' || true)
+            if [ -n "$BAD_LINES" ]; then
+              BAD="$BAD\n$f:\n$BAD_LINES"
```
mikelane: **LOW: the `\n` in the accumulator is literal, not a newline.** `BAD="\$BAD\n\:\n\"` stores the two-character sequence backslash-n in the variable. The later `printf "✗ ...%s\n" "\$BAD"` uses `%s` which passes the string through unchanged, so the output will look like:

```
✗ cargo clippy without --all-targets:\nci.yml:\n42:  run: cargo clippy
```

instead of actual line breaks.

Options:

1. Use `printf %b` which interprets backslash escapes in the argument.
2. Use a real newline via ANSI-C quoting: `BAD="\$BAD"\$'\n'"\$f:"\$'\n'"\$BAD_LINES"`
3. Or just `printf '%s\n' "\$BAD"` after setting `BAD$'\n...'`

Same issue in Check 5 (@not_implemented), line ~131.

scripts/validate-pr-refs.sh
```diff
@@ -0,0 +1,47 @@
...
+  # echoed number for any numeric input without actually resolving it,
+  # so checking a server-side field like `state` forces real validation.
+  if gh pr view "$N" --json state --jq .state >/dev/null 2>&1; then
+    continue
+  fi
+  if gh issue view "$N" --json state --jq .state >/dev/null 2>&1; then
+    continue
+  fi
+  FAILED="$FAILED #$N"
+done
```
mikelane, line 35: **MEDIUM: no caching means rebases and amends are slow, and a missing/unauthenticated gh degrades silently to "everything is phantom."**

On a rebase of 10 commits each referencing 3 refs, this hook hits gh 60 times (30 PR, 30 issue). For 50 commits it's 300. GitHub's rate limit for authenticated gh is 5000/hour, so that's fine for a human but will dominate wall-clock time (each gh call is ~200-500ms of network + auth + JSON parse).

A tiny on-disk cache keyed on ref number would eliminate the repeated lookups:

```bash
CACHE="\/Users/mikelane/.cache/git-prism-pr-refs"
mkdir -p "\"
for N in \; do
  if [ -f "\/\" ]; then continue; fi
  if gh pr view "\" --json state --jq .state >/dev/null 2>&1 \
    || gh issue view "\" --json state --jq .state >/dev/null 2>&1; then
    touch "\/\"
    continue
  fi
  FAILED="\ #\"
done
```

Also, there's no graceful handling when gh is missing or unauthenticated: `gh pr view` returns a non-zero exit that looks identical to "not found," so every commit message with a ref would be rejected. Consider checking once up front:

```bash
if ! command -v gh >/dev/null || ! gh auth status >/dev/null 2>&1; then
  echo "warning: gh unavailable, skipping PR-ref validation" >&2
  exit 0
fi
```

Without this, a fresh clone where the user hasn't run `gh auth login` will have every commit rejected, which is a bad UX failure mode for a local hook.

scripts/validate-pr-refs.sh
```diff
@@ -0,0 +1,47 @@
...
+set -euo pipefail
+
+COMMIT_MSG_FILE="${1:-}"
+if [ -z "$COMMIT_MSG_FILE" ] || [ ! -f "$COMMIT_MSG_FILE" ]; then
+  echo "usage: $0 <commit-msg-file>" >&2
+  exit 2
+fi
+
+# Extract all (#N) references from the commit message.
+REFS=$(grep -oE '\(#[0-9]+\)' "$COMMIT_MSG_FILE" | grep -oE '[0-9]+' | sort -u || true)
```
mikelane, line 17: **LOW: the script only extracts `(#N)` in parens, missing bare `#42` references.** The extraction uses `grep -oE '\(#[0-9]+\)'`. That catches Conventional Commits trailers but will not match body references like:

- `Refs: #42` (common in this repo per the commit messages in this very branch)
- `See #42 for context`
- `#42 is the parent issue`

Two tradeoffs:

- Stay narrow: only `(#N)` is validated. Advantage: no false positives on substring matches in pasted logs or code blocks. Disadvantage: phantom bare refs slip through. This is the current state and probably intentional.
- Go broad: match `#[0-9]+` anywhere. Advantage: catches all phantom refs. Disadvantage: false positives inside code blocks or pasted diffs.

Worth a short comment in the script header stating the intent explicitly so a future maintainer does not broaden the pattern without thinking about it. The existing comment says "(#N) references" but does not spell out "deliberately narrow to avoid pasted-code false positives."

src/server.rs
```diff
@@ -704,4 +723,23 @@ mod tests {
...
+        actual.sort();
+        let mut expected: Vec<String> = EXPECTED_TOOLS.iter().map(|s| s.to_string()).collect();
+        expected.sort();
+        assert_eq!(
+            actual, expected,
+            "MCP tool set drifted from EXPECTED_TOOLS. \
+             If you added, removed, or renamed a tool, update the EXPECTED_TOOLS constant \
+             in src/server.rs AND update README.md 'N MCP tools' count, CHANGELOG, and CLAUDE.md."
+        );
+    }
```
mikelane, line 744: **LOW: `it_registers_exactly_four_tools` is now redundant with this test.** Any drift in tool set or count now fails `it_matches_expected_tools` with a clearer error message (diff of set contents vs. just "count was N"). Consider removing the older count-only test to avoid two assertions for the same invariant.

Also worth considering: the current contract is "the SET of tool names matches." That catches the specific drift class the PR targets (get_function_context not registered). A stricter contract could also pin:

- Tool schemas (input_schema bytes, output_schema bytes) — catches schema drift
- Tool descriptions — catches description drift
- Tool order if order-dependent in any consumer

A rmcp `router.list_all()` `Tool` struct exposes `name`, `description`, `input_schema`, `annotations`. Comparing just names is a MVP; consider layering a second snapshot test (via `insta`, which is already a project dep) of the full Tool struct sequence. That would turn "four specific tools exist" into "these four tools with these exact shapes exist," which is the stronger compile-time contract the PR is reaching for.

mikelane (commented):
## Review summary

This PR implements the right four components to address the April 2026 retrospective, and the TDD audit trail for Component C (EXPECTED_TOOLS) is exemplary. But it ships with a **CI job that is currently failing** against its own head (run 24288651484 on commit 68314dd), and the failure is a design flaw in Component A Check 3 that must be fixed before merge.

## Blockers

1. **CRITICAL — Check 3 (`|| true` ban) is a false-positive generator on its own policy comments.** The follow-up commit 68314dd added explanatory comments to mutants.yml and release.yml containing backticked `|| true` strings, and the checks now trip on those comments. The integrity workflow is failing with:

   ```
   .github/workflows/mutants.yml:73:        # continue-on-error rather than `|| true` so the failure is visible
   .github/workflows/mutants.yml:75:        # bans `|| true` in workflow files because of incidents
   .github/workflows/release.yml:322:            # than masking with `|| true` — the integrity workflow bans
   ```

   The pattern needs to exclude comments (e.g. `grep -rnE '^[^#]*\|\| true'`) or better, parse YAML properly and only check `run:` string values. See inline comment on integrity.yml:80.

2. **HIGH — sequential-step observability gap.** The workflow uses default step semantics, so when Check 3 fails the remaining checks (clippy, @not_implemented, CHANGELOG) are all skipped. This contradicts the commit messages claim that "each check is a separate step so failures are individually identifiable." Add `if: ${{ !cancelled() }}` to each step or split into independent jobs. See inline comment on integrity.yml:33.

## High-priority concerns

3. **HIGH — Check 1 allowlist `^chore.*release` is too broad.** It matches `chore: release the hounds` and `chore: refactor release module` as bypasses of the tag-sync check. Tighten to require `release` followed by a version number. See inline on integrity.yml:40.

4. **MEDIUM — Check 5 regex is case-sensitive and misses URL-form closes.** `(Closes|Fixes|Resolves) #N` does not match `closes #42`, `Fix #42`, or `Closes https://github.com/.../issues/42`. See inline on integrity.yml:122.

5. **MEDIUM — Check 6 does not handle pre-release version strings.** The first `## [0.7.0-rc.1]` heading silently escapes. See inline on integrity.yml:151.

6. **MEDIUM — stacked-PR trigger gap.** `pull_request: branches: [main]` does not fire for PRs targeting non-main branches, which is the same class of gap that bit PR #146. The retrospective explicitly flagged this. See inline on integrity.yml:16.

7. **MEDIUM — validate-pr-refs.sh does not detect missing/unauthenticated gh.** A fresh clone without `gh auth login` will reject every commit message that contains a real `(#N)` because `gh pr view` returns non-zero for both "not found" and "not authenticated." Short-circuit with `command -v gh && gh auth status` up front. Also no caching for rebases. See inline on scripts/validate-pr-refs.sh:35.

## Low-severity findings

8. **LOW — literal `\\n` in BAD accumulator.** Checks 4 and 5 build `BAD="$BAD\\n..."` then `printf "...%s\\n" "$BAD"`, which emits literal backslash-n. Use `printf %b` or ANSI-C quoted `$"'\n'"`. See inline on integrity.yml:97.

9. **LOW — `(#N)` extraction is narrow by design but not documented.** `Refs: #42` in the body is not validated. Whether that is intentional or not should be in the script header. See inline on scripts/validate-pr-refs.sh:17.

10. **LOW — `it_registers_exactly_four_tools` is now redundant with `it_matches_expected_tools`.** The set comparison subsumes the count check. Consider consolidating. Stronger contract would snapshot the full `Tool` struct via `insta`. See inline on src/server.rs:744.

11. **LOW — `integrity.yml:` exclusion in Check 3 is a substring match.** A future `integrity.yml.bak` or any grep output containing the literal `integrity.yml:` would be silently excluded. Prefer `grep -v '^\\.github/workflows/integrity\\.yml:'` anchored at the start of the line.

12. **LOW — `grep -c '#\\[tool('` in Check 2 is source-code-naive.** It would count a doc comment that contains `#[tool(` as a real registration. Today it happens to not trigger, but a stricter check would filter to non-comment lines.

## What is good here

- **Excellent security hygiene on PR_TITLE:** passing the PR title via `env:` rather than `${{ }}` interpolation in a run block is the correct defense against workflow injection, and the header comment explicitly documents the reasoning. This is the right pattern.
- **TDD audit trail for Component C** (EXPECTED_TOOLS test) is textbook: RED (test before constant) -> GREEN -> TRIANGULATE (delete a tool, confirm failure message is actionable) -> GREEN. The commit message documents the full cycle.
- **The bug the agent caught in validate-pr-refs.sh is real and well-fixed.** The `gh pr view N --json number` returning the echoed number without API resolution is a gotcha that would silently make the entire phantom-ref check a no-op. Switching to `--json state` is the correct fix and the comment in the script explains why. This is the kind of detail that gets missed in code review.
- **Documentation is tied to enforcement.** Each rule in the new CLAUDE.md Release Hygiene section points at the specific workflow step or test that enforces it. This is the right shape — docs and mechanism stay in lockstep.
- **`EXPECTED_TOOLS` as a `const &[&str]`** with a docstring that explains the drift history is a clean, idiomatic Rust contract.

## Decision

**CHANGES REQUESTED.** The CI is currently failing against the head commit and that failure is exposing two independent real issues in Check 3 (false positives on comments, and the cascading-skip that hides later failures). These have to land before merge. The other findings are improvements that can either go in this PR or as a follow-up.

Note: CI status as of this review — Integrity checks FAILURE on run 24288651484 (head 68314dd2). CI, Clippy, Format success. Tests and BDD in progress but unrelated to the integrity blocker.

PR #204, opened 2026-04-11
security: recursion depth guards for python/ts/rust/cpp/c analyzers

### LINKED ISSUE #187: Tier 4: add recursion depth guards to python/ts/rust/cpp/c analyzers
## Finding

PR #181 fixed stack-overflow DoS in `csharp.rs` and `ruby.rs` analyzers by adding `MAX_RECURSION_DEPTH = 256` guards. But the PR body claimed the at-risk follow-up targets were `php/swift/kotlin/java` — the PR #181 reviewer corrected this: those four are **flat/safe**. The actual at-risk analyzers with identical unbounded-recursion patterns are:

- `src/treesitter/python.rs::extract_functions_from_node` — recurses on `class_definition` (highest traffic language, plausible exploit surface)
- `src/treesitter/typescript.rs::extract_functions_from_node` — recurses on `class_declaration` (shared by TS/TSX/JS)
- `src/treesitter/rust_lang.rs::extract_functions_from_node` — recurses on `impl_item`
- `src/treesitter/cpp.rs::collect_functions` — recurses on namespace/class/struct + `#ifdef` containers
- `src/treesitter/cpp.rs::collect_imports` — recurses on `#ifdef` containers
- `src/treesitter/c_lang.rs::collect_functions` — recurses on `#ifdef` containers
- `src/treesitter/c_lang.rs::collect_imports` — recurses on `#ifdef` containers

## Fix

For each function above, add the same `depth: usize` parameter + `MAX_RECURSION_DEPTH` guard pattern already established in `src/treesitter/csharp.rs::visit_node` and `src/treesitter/ruby.rs::extract_functions_from_node`. The constant is already shared in `src/treesitter/mod.rs`.

TDD: for each analyzer, add a test that constructs a deeply-nested source string (~5000 levels of the relevant nesting construct), runs the analyzer, and asserts no stack overflow. Reference PR #181's tests.

Alternative: convert to iterative stack-based traversal (like `extract_calls` already does) to eliminate the depth guard entirely. Worth considering.

## Discovered during

PR #181 review (2026-04-10). The original A4 gauntlet wording was imprecise; the PR reviewer did empirical verification of the actual at-risk set.

mikelane:
Closes #187.

## Summary

Tier 4 follow-up from the April 2026 B8 security deep-dive. Five tree-sitter
analyzers walked untrusted parse trees with unbounded recursion — a pathological
input (thousands of nested braces/classes/impls) could blow the Rust thread
stack and crash the MCP server for every caller, not just the one sending the
bad blob. This PR ships the same `MAX_RECURSION_DEPTH = 256` defense applied to
`csharp.rs` and `ruby.rs` in PR #181 across the five remaining analyzers.

## Five commits, one per analyzer

| Commit | Analyzer | Recursion sites guarded |
|--------|----------|------------------------|
| `f5ad3db` | `cpp.rs` | 5 in `collect_functions` (namespace, class, linkage_spec, declaration_list, preprocessor), 1 in `collect_imports` |
| `2e6cfdb` | `c_lang.rs` | 1 in `collect_functions` (preprocessor), 1 in `collect_imports` (preprocessor) |
| `ad46971` | `python.rs` | 2 in `extract_functions_from_node` (class body, `decorated_definition`) |
| `61e99a7` | `typescript.rs` | 2 in `extract_functions_from_node` (`class_declaration` body, `export_statement`) |
| `7bef7fe` | `rust_lang.rs` | 1 in `extract_functions_from_node` (`impl_item` body) |

## TDD proof

Each commit has RED→GREEN→TRIANGULATE coverage:

- **RED test** (depth 5000, 2 MB bounded-stack thread): confirms the attack
  actually SIGABRTs the unguarded analyzer. The thread's `.join()` returning
  `Err(...)` is the stack-overflow detector.
- **GREEN**: `if depth >= MAX_RECURSION_DEPTH { return; }` as the first
  statement of each recursive walker; entry points seeded at 0, recursion
  sites pass `depth + 1`.
- **TRIANGULATION test** (depth 255 or 255 sequential constructs): confirms
  legitimate code still extracts correctly at the boundary.

## Escalation clause outcomes

- **`rust_lang.rs`**: Valid Rust doesn't allow nested `impl` blocks, but
  tree-sitter's error recovery DOES produce nested `impl_item` nodes from
  `impl A { impl B { ... } }`. The RED test SIGABRTs as expected — the attack
  vector is real.
- **`typescript.rs`**: TS error recovery produces `ERROR` nodes (not nested
  `export_statement` / `class_declaration`) from stacked `export` keywords.
  No natural SIGABRT. Guard added for defense-in-depth and consistency; commit
  body documents the tree-sitter behavior. One fix covers JS + TSX + TS via
  the shared `JsDialect` walker.
- **`python.rs`**: Python's indented nesting produces small per-frame sizes;
  unguarded recursion doesn't SIGABRT the 2 MB thread at any tested depth.
  Guard added for defense-in-depth; RED test verifies structural safety (partial
  extraction completes, ≥ 256 classes extracted before guard fires).

## What didn't change

- `csharp.rs`, `ruby.rs`, `kotlin.rs` — already fixed or out of scope
- `src/treesitter/mod.rs` — `MAX_RECURSION_DEPTH = 256` already existed there;
  analyzers import it via `use super::{..., MAX_RECURSION_DEPTH}`
- `extract_calls()` — already uses an iterative `Vec<Node>` stack; no guard needed
- Silent truncation behavior — early return without error matches the existing
  csharp/ruby behavior; telemetry hook is tracked separately (see follow-up issue)

## Church purification

Post-implementation cleanup (commits `532d1c6`, `1f7e580`):
- Test functions renamed to `it_verb_noun` convention
- Constants renamed `NESTING_DEPTH` → `GENERATED_NESTING_LEVELS`, `TEST_STACK_SIZE` → `CONSTRAINED_THREAD_STACK_BYTES`
- `python.rs` assertion strengthened: `!functions.is_empty()` → `functions.len() >= MAX_RECURSION_DEPTH`
- `typescript.rs`: `JsDialect` visibility reduced to private (no external importers)
- NOTE comments added to rust_lang.rs and typescript.rs triangulation tests explaining structural constraints

---

### DIFF
--- a/src/treesitter/c_lang.rs
+++ b/src/treesitter/c_lang.rs
@@ -371,22 +371,22 @@ void guarded(int x) {
     /// Runs on a thread with a 2 MB stack: roomy enough for bounded recursion to
     /// `MAX_RECURSION_DEPTH` but far too small for unbounded recursion to 5000 frames.
     #[test]
-    fn deeply_nested_preproc_blocks_do_not_stack_overflow() {
-        const NESTING_DEPTH: usize = 5000;
-        const TEST_STACK_SIZE: usize = 2 * 1024 * 1024;
+    fn it_completes_without_overflow_on_deeply_nested_preproc_blocks() {
+        const GENERATED_NESTING_LEVELS: usize = 5000;
+        const CONSTRAINED_THREAD_STACK_BYTES: usize = 2 * 1024 * 1024;
 
         let mut source = String::new();
-        for i in 0..NESTING_DEPTH {
+        for i in 0..GENERATED_NESTING_LEVELS {
             source.push_str(&format!("#ifdef MACRO_{i}\n"));
         }
         // Function is past the cap — it should not be extracted.
         source.push_str("void deep_fn(void) {}\n");
-        for _ in 0..NESTING_DEPTH {
+        for _ in 0..GENERATED_NESTING_LEVELS {
             source.push_str("#endif\n");
         }
 
         let handle = std::thread::Builder::new()
-            .stack_size(TEST_STACK_SIZE)
+            .stack_size(CONSTRAINED_THREAD_STACK_BYTES)
             .spawn(move || {
                 let analyzer = CAnalyzer;
                 analyzer.extract_functions(source.as_bytes())
@@ -404,15 +404,15 @@ void guarded(int x) {
     /// Triangulation: 255 nested `#ifdef` blocks with a function at the innermost level.
     /// The guard fires at depth 256, so depth 255 must still allow extraction.
     #[test]
-    fn nested_preproc_blocks_at_boundary_depth_still_extract_functions() {
-        const NESTING_DEPTH: usize = 255;
+    fn it_extracts_functions_at_boundary_nesting_depth() {
+        const GENERATED_NESTING_LEVELS: usize = 255;
 
         let mut source = String::new();
-        for i in 0..NESTING_DEPTH {
+        for i in 0..GENERATED_NESTING_LEVELS {
             source.push_str(&format!("#ifdef MACRO_{i}\n"));
         }
         source.push_str("void leaf_fn(void) {}\n");
-        for _ in 0..NESTING_DEPTH {
+        for _ in 0..GENERATED_NESTING_LEVELS {
             source.push_str("#endif\n");
         }
 
--- a/src/treesitter/c_lang.rs
+++ b/src/treesitter/c_lang.rs
@@ -1,4 +1,6 @@
-use super::{CallSite, Function, LanguageAnalyzer, body_hash_for_node, sha256_hex};
+use super::{
+    CallSite, Function, LanguageAnalyzer, MAX_RECURSION_DEPTH, body_hash_for_node, sha256_hex,
+};
 use tree_sitter::Parser;
 
 pub struct CAnalyzer;
@@ -28,7 +30,15 @@ fn is_preprocessor_container(kind: &str) -> bool {
     )
 }
 
-fn collect_functions(node: &tree_sitter::Node, source: &[u8], functions: &mut Vec<Function>) {
+fn collect_functions(
+    node: &tree_sitter::Node,
+    source: &[u8],
+    functions: &mut Vec<Function>,
+    depth: usize,
+) {
+    if depth >= MAX_RECURSION_DEPTH {
+        return;
+    }
     let mut cursor = node.walk();
     for child in node.children(&mut cursor) {
         match child.kind() {
@@ -70,14 +80,22 @@ fn collect_functions(node: &tree_sitter::Node, source: &[u8], functions: &mut Ve
                 }
             }
             kind if is_preprocessor_container(kind) => {
-                collect_functions(&child, source, functions);
+                collect_functions(&child, source, functions, depth + 1);
             }
             _ => {}
         }
     }
 }
 
-fn collect_imports(node: &tree_sitter::Node, source: &[u8], imports: &mut Vec<String>) {
+fn collect_imports(
+    node: &tree_sitter::Node,
+    source: &[u8],
+    imports: &mut Vec<String>,
+    depth: usize,
+) {
+    if depth >= MAX_RECURSION_DEPTH {
+        return;
+    }
     let mut cursor = node.walk();
     for child in node.children(&mut cursor) {
         match child.kind() {
@@ -86,7 +104,7 @@ fn collect_imports(node: &tree_sitter::Node, source: &[u8], imports: &mut Vec<St
                 imports.push(text);
             }
             kind if is_preprocessor_container(kind) => {
-                collect_imports(&child, source, imports);
+                collect_imports(&child, source, imports, depth + 1);
             }
             _ => {}
         }
@@ -100,7 +118,7 @@ impl LanguageAnalyzer for CAnalyzer {
             .parse(source, None)
             .ok_or_else(|| anyhow::anyhow!("Failed to parse C source"))?;
         let mut functions = Vec::new();
-        collect_functions(&tree.root_node(), source, &mut functions);
+        collect_functions(&tree.root_node(), source, &mut functions, 0);
         Ok(functions)
     }
 
@@ -140,7 +158,7 @@ impl LanguageAnalyzer for CAnalyzer {
             .parse(source, None)
             .ok_or_else(|| anyhow::anyhow!("Failed to parse C source"))?;
         let mut imports = Vec::new();
-        collect_imports(&tree.root_node(), source, &mut imports);
+        collect_imports(&tree.root_node(), source, &mut imports, 0);
         Ok(imports)
     }
 }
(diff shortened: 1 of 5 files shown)


mikelane (commented):
## Code Review — PR #204: Recursion Depth Guards for Python/TS/Rust/C++/C Analyzers

**Solo-dev review. CI is passing. This is the authoritative review before merge.**

---

### Summary

This PR closes a genuine security vulnerability: an attacker committing a file with thousands of nested structures (namespaces in C++, `#ifdef` chains in C, `impl` blocks in Rust, classes in Python/TS) could crash git-prism via stack overflow during `get_change_manifest`. The fix is mechanically correct, consistently applied, and well-evidenced by the TDD cycle documented in the commit messages.

Overall quality is high. The issues below are ordered by severity. There are no blockers.

---

### Phase 0 — Bug Verification

**Adversarial question 2: Guard placement and depth threading.**

All five walkers place the guard as the **first statement** of the recursive function, before any allocation, cursor creation, or iteration. Confirmed at:

- `c_lang.rs:39` and `c_lang.rs:96` — both `collect_functions` and `collect_imports`
- `cpp.rs:48` and `cpp.rs:151` — both walkers
- `python.rs:30` — `extract_functions_from_node`
- `rust_lang.rs:29` — `extract_functions_from_node`
- `typescript.rs:63` — `extract_functions_from_node`

Every recursive callsite passes `depth + 1`. Every public entry point seeds `0`. No bugs found here.

**Adversarial question 3: RED proof.**

Each of the five fix commits includes a bounded-stack thread test (`CONSTRAINED_THREAD_STACK_BYTES = 2 * 1024 * 1024`) that calls `.join().expect("analyzer thread must not stack-overflow on deeply-nested input")`. The `.expect()` message serves as the SIGABRT detector — a thread abort causes `.join()` to return `Err`, which `.expect()` turns into a panic with that message. Pattern is correct.

**Adversarial question 4: Triangulation at boundary.**

All five analyzers include a depth-255 boundary test asserting correct extraction. The Rust `sequential_impls_all_extract` test uses 255 sequential (non-nested) impls rather than nested impls, and the NOTE comment at `rust_lang.rs:374–378` explains why: tree-sitter Rust error-recovers nested impls to ERROR nodes, so true depth-255 boundary testing via nesting is not achievable. This is the correct engineering decision, clearly documented.

**Adversarial question 7: JS/TSX coverage.**

`typescript.rs` registers `TypeScript`, `Tsx`, and `JavaScript` dialects via the `JsDialect` enum. All three use `extract_functions_from_node` as the single walker, and `analyzer_for_extension` in `mod.rs:88-90` routes `.ts`, `.tsx`, `.js`, `.jsx` through that shared path. The depth guard at `typescript.rs:63` covers all four extensions.

No bugs found across all five adversarial checks.

---

### Phase 1 — Quality Tools

CI passes on ubuntu + macos + windows. Format, clippy (`--all-targets`), tests, BDD acceptance, integrity checks, mutation testing incremental — all green. No action required.

---

No changes to `.pre-commit-config.yaml`, `lefthook.yml`, `pyproject.toml`, or any linter config. Scope is clean.

---

### Phase 3 — Code Standards

**MEDIUM** — `src/treesitter/rust_lang.rs:367`

The overflow test (`it_completes_without_overflow_on_deeply_nested_impls`) does not assert on function count:

```rust
result.expect("analyzer must return Ok on deeply-nested input");
// No assertion on function count — tree-sitter Rust error recovery
// produces ERROR nodes for illegal nested impls, not nested impl_item
// nodes, so the walker may or may not extract the leaf function.
```

The comment is accurate — tree-sitter Rust error-recovers nested impls to ERROR nodes, making the extraction count non-deterministic. The test is therefore correct to omit the count assertion. However, the comment says "may or may not extract the leaf function" when the commit message says the grammar DOES produce nested `impl_item` nodes under error recovery (contradicting the comment). The commit message for `7bef7fe` states: "the tree-sitter Rust grammar's error recovery DOES produce nested `impl_item` nodes from this input — contrary to the initial plan assumption." The inline comment in the test (`532d1c6`) says the opposite: "tree-sitter Rust error recovery produces ERROR nodes for illegal nested impls, not nested `impl_item` nodes."

This is a documentation contradiction between the commit message and the test comment. The test is safe either way (it just verifies no crash), but the comment needs to be aligned with the actual observed grammar behavior. Which is it: do 5000 nested impls produce ERROR nodes or nested `impl_item` nodes? The RED evidence (SIGABRT confirmed at 5000 levels) strongly implies the latter — you cannot stack-overflow on ERROR nodes that the walker ignores. If ERROR nodes were produced, the recursion would never get deep enough to overflow.

Recommendation: revise the comment to match the commit message's observed reality ("error recovery produces nested `impl_item` nodes from illegal nested impl input, causing actual stack overflow").

**LOW** — `src/treesitter/typescript.rs:519–536`

The overflow test uses stacked `export` keywords as the adversarial payload:

```rust
for _ in 0..GENERATED_NESTING_LEVELS {
    source.push_str("export ");
}
source.push_str("class C {}\n");
```

The test comment (lines 499–510) correctly explains that stacked `export` keywords error-recover to a single `export_statement` node, not nested ones, so this payload cannot trigger SIGABRT. The test then asserts `result.expect(...)` with no function count assertion — fine, given the grammar behavior. But the test name is `it_completes_without_overflow_on_deeply_nested_export_statements` when the payload doesn't actually produce deeply-nested export_statement nodes at all. The name overpromises what is actually being tested (which is more "the analyzer handles malformed stacked-export input without crashing"). This is a naming accuracy issue, not a correctness issue.

---

### Phase 5 — Test Quality

**MEDIUM** — `src/treesitter/python.rs:350–387`

The overflow test uses `GENERATED_NESTING_LEVELS = 1024` instead of `5000` as in the other four analyzers. The comment explains this is to avoid `O(n²)` string construction from indent repetition. This is a pragmatic and correct tradeoff, but it means the Python overflow test exercises 4x the depth cap, not 20x the depth cap. The test still clearly demonstrates the guard functions (1024 > 256), so this is acceptable. Worth noting that the asymmetry is intentional per the comment.

The Python overflow test asserts `functions.len() >= MAX_RECURSION_DEPTH` rather than the simpler `!functions.is_empty()` used in earlier drafts. This is the stronger assertion — it catches regressions where the guard fires too early (e.g., at depth 10). This is explicitly called out in `532d1c6` as a purification improvement. Well done.

**LOW** — `src/treesitter/typescript.rs:547–564`

The triangulation test (`it_extracts_methods_from_exported_classes`) builds 255 sequential exported classes. The NOTE comment correctly explains this is not testing depth-255 boundary (because tree-sitter error-recovers stacked exports) but rather non-regression of the export_statement arm. The assertion `functions.len() == CLASS_COUNT` is correct. One minor observation: the `method{i}(): void {}` syntax in the source string is missing a space before `{` (`{{` in the format string is the escaped brace, so the generated code is `method0(): void {}`). This is valid TypeScript and parses correctly — confirmed by the test passing. No action needed.

---

### Phase 6 — Architecture

**Adversarial question 8: Church purification scope.**

Commit `532d1c6` touches all five analyzer files but makes only cosmetic changes: test renames, constant renames, comment additions, one visibility reduction (`pub` → private on `JsDialect`), and one assertion strengthening in `python.rs`. No production code paths changed.

Commit `1f7e580` touches only `rust_lang.rs` to fix import order after `532d1c6` broke `fmt --check`. Single-file, single-concern. Clean.

**Adversarial question 9: Commit atomicity.**

Every fix commit touches exactly one file. `532d1c6` intentionally spans all five (it's a purification pass, correctly scoped to the five new analyzers only). This is appropriate: a single purification commit is cleaner than five separate cosmetic commits. No mixing of behavioral changes with cosmetic changes detected.

**Adversarial question 5: Silent truncation parity.**

The early-return behavior (`return;`) exactly matches `csharp.rs:79` and `ruby.rs:29`. No error is surfaced, no partial result is signaled — truncation is silent. This matches the existing contract. Issue #205 (confirmed open) tracks adding a truncation metric when the depth guard fires. The follow-up is correctly tracked and not missing.

**Adversarial question 6: Shared constant.**

All five files import `MAX_RECURSION_DEPTH` from `super::` — no per-file redefinitions. Confirmed via grep: all seven guarded analyzers (including the precedent `csharp.rs` and `ruby.rs`) use the same import pattern. Single source of truth maintained.

**Adversarial question 10: Documentation drift.**

README and CHANGELOG contain no claims about recursion depth limits or tree-sitter walking behavior. CLAUDE.md references `MAX_RECURSION_DEPTH = 256` in the `src/treesitter/mod.rs` description, but only in the context of the constant existing — no claims requiring update. No documentation drift found.

---

### Phase 7 — Security

This PR is a **security fix**, not a feature. The attack surface being closed: git-prism reads arbitrary git blobs from a repository and passes them to tree-sitter for parsing. A malicious committer can craft source files with pathological nesting (5000 nested C++ namespaces = ~15 KB of text) that previously caused stack exhaustion and process abort. The `2 MB bounded-stack thread` test methodology is the correct approach — it proves the crash behavior at 5000 levels without the guard, and proves safe completion with the guard.

The depth cap of 256 is generous for legitimate code (humans rarely nest more than ~20 levels) while staying well within default thread stack limits. The `mod.rs:28` constant comment explains this rationale explicitly.

---

### Praise

The TDD cycle documentation in the commit messages is exemplary. Each commit message documents RED/GREEN/TRIANGULATE with specific evidence (SIGABRT confirmed, depth-255 extraction confirmed). The escalation clause pattern — checking whether the grammar actually produces the pathological AST nodes before claiming SIGABRT, then falling back to defense-in-depth with documented rationale — shows rigorous engineering discipline. The `JsDialect` visibility reduction from `pub` to private is a nice hygiene catch. The Python assertion strengthening from `!is_empty()` to `>= MAX_RECURSION_DEPTH` is exactly right.

---

### Decision

**APPROVED** — no blockers. The one actionable recommendation before merge:

- **Resolve the comment/commit-message contradiction in `rust_lang.rs`** at lines 368–371: the comment says "ERROR nodes ... so the walker may or may not extract the leaf function" but the commit message for `7bef7fe` says the grammar DOES produce nested `impl_item` nodes (which is what caused the actual SIGABRT). The comment should be corrected to match observed reality. This is low-risk to fix inline or leave for a follow-up, but leaving a false comment in a security-sensitive test is worth the two-line fix.

All five guards are correctly placed, consistently implemented, tested with genuine RED evidence (or documented defense-in-depth escalation), and triangulated at the boundary. Issue #205 correctly tracks the observability gap. Safe to merge after the comment correction.

PR #206, opened 2026-04-12
observability: emit tracing warning when tree-sitter depth guard fires

### LINKED ISSUE #205: observability: emit truncation metric when depth guard fires in tree-sitter analyzers
## Background

PR #181 added `MAX_RECURSION_DEPTH = 256` guards to `csharp.rs` and `ruby.rs`.
PR #204 extended the same guards to `cpp.rs`, `c_lang.rs`, `python.rs`,
`typescript.rs`, and `rust_lang.rs`.

All seven analyzers return early (silent truncation) when depth hits the cap —
we prefer "missed a deeply-nested function" over "crashed the process." This is
correct behavior, but it's currently invisible to operators.

## Problem

`src/metrics.rs` has a `record_truncated` helper intended for exactly this
case, but it is never called from any depth-guard path. When the guard fires
in production (malicious or unusually-nested input), the MCP server returns a
partial result with no signal in telemetry that truncation occurred.

## What this issue should do

1. Call `record_truncated` (or an equivalent counter) when the depth guard
   fires in any of the seven analyzers.
2. The call needs access to the `Metrics` handle — currently the analyzer
   trait methods don't receive one. Options:
   - Thread `Option<&Metrics>` through the walker signatures (invasive)
   - Return a `(Vec<Function>, bool truncated)` tuple from walkers and
     record the metric in the tool handler
   - Add an `AtomicBool` to the analyzer struct (low overhead, not idiomatic)
   - Use a structured log event (`tracing::warn!`) as a lightweight alternative
3. Evaluate whether a counter-per-language or a single `truncated_by_depth`
   counter is more useful for dashboards.

## Out of scope

- Changing the silent-return behavior to return a `Result<_, Error>` — the MCP
  tool response shape must stay unchanged (tracked separately if ever desired)
- Changes to `csharp.rs` / `ruby.rs` unless the chosen approach is fully
  mechanical

## Precedent

`src/pagination.rs` threads a truncation reason through the response struct so
callers can observe paging limits. A similar pattern may be the cleanest path
here.

## References

- PR #204 (adds guards) — security audit noted this as INFO finding
- PR #181 (csharp/ruby guards) — same silent-return behavior
- `src/metrics.rs::record_truncated` — exists but unused for depth guards

mikelane:
## Summary

Closes #205

When a tree-sitter analyzer's depth guard fires (recursion hits `MAX_RECURSION_DEPTH = 256`), the analyzer silently returns a truncated result. An operator watching OTLP dashboards cannot distinguish "file has no functions" from "we stopped extracting at depth 256 because the input was adversarially deep." This PR closes that blind spot.

### Changes

**9 guard sites instrumented across 7 analyzers:**

| File | Function | `language` field |
|------|----------|-----------------|
| `src/treesitter/csharp.rs` | `visit_node` | `"csharp"` |
| `src/treesitter/ruby.rs` | `extract_functions_from_node` | `"ruby"` |
| `src/treesitter/python.rs` | `extract_functions_from_node` | `"python"` |
| `src/treesitter/typescript.rs` | `extract_functions_from_node` | `"typescript"` |
| `src/treesitter/rust_lang.rs` | `extract_functions_from_node` | `"rust"` |
| `src/treesitter/cpp.rs` | `collect_functions` | `"cpp"` |
| `src/treesitter/cpp.rs` | `collect_imports` | `"cpp"` |
| `src/treesitter/c_lang.rs` | `collect_functions` | `"c"` |
| `src/treesitter/c_lang.rs` | `collect_imports` | `"c"` |

**Uniform warning message at all 9 sites:**
```
tracing::warn!(
    depth_limit = MAX_RECURSION_DEPTH,
    language = "<lang>",
    "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
)
```
(cpp/c `collect_imports` sites use "imports may be missing" instead of "functions may be missing".)

**`tracing-test = "0.2"` added to `[dev-dependencies]`** — provides `#[traced_test]` + `logs_contain()` for the new warning-assertion tests.

**`src/telemetry.rs` fix:** `opentelemetry::global::set_meter_provider()` installs a global tracing subscriber as a side effect. This poisoned `tracing_test`'s `Once` in test builds, causing spurious log-capture failures. Gated with `#[cfg(not(test))]`.

### Why `tracing::warn!` (not alternatives)

From `HANDOFF_TASK29.md`:
- **Trait signature change** — would ripple across all 7 analyzers + callers + MCP tool layer
- **`AtomicBool` per analyzer** — adds per-struct state for a rare event; requires atomic reads in callers
- **Thread `Option<&Metrics>`** — would require adding a metrics parameter to every recursive call
- **Tuple return type** — breaks the `LanguageAnalyzer` trait contract for all callers

`tracing::warn!` needs zero trait or call-site changes and flows through the existing `tracing-opentelemetry` bridge automatically.

### New tests

Each of the 9 guard sites now has:
- `it_emits_depth_guard_warning_on_deeply_nested_*` — `#[traced_test]` asserts `logs_contain("depth guard fired")`
- `it_does_not_emit_depth_guard_warning_on_shallow_input` — triangulation: shallow input must NOT emit the warning

**601 tests passing.** clippy clean. fmt clean.

---

### DIFF
--- a/src/treesitter/c_lang.rs
+++ b/src/treesitter/c_lang.rs
@@ -39,6 +39,7 @@ fn collect_functions(
     if depth >= MAX_RECURSION_DEPTH {
         tracing::warn!(
             depth_limit = MAX_RECURSION_DEPTH,
+            language = "c",
             "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
         );
         return;
@@ -100,6 +101,7 @@ fn collect_imports(
     if depth >= MAX_RECURSION_DEPTH {
         tracing::warn!(
             depth_limit = MAX_RECURSION_DEPTH,
+            language = "c",
             "tree-sitter depth guard fired: recursive walk truncated; some imports may be missing"
         );
         return;
--- a/src/treesitter/c_lang.rs
+++ b/src/treesitter/c_lang.rs
@@ -37,6 +37,10 @@ fn collect_functions(
     depth: usize,
 ) {
     if depth >= MAX_RECURSION_DEPTH {
+        tracing::warn!(
+            depth_limit = MAX_RECURSION_DEPTH,
+            "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
+        );
         return;
     }
     let mut cursor = node.walk();
@@ -94,6 +98,10 @@ fn collect_imports(
     depth: usize,
 ) {
     if depth >= MAX_RECURSION_DEPTH {
+        tracing::warn!(
+            depth_limit = MAX_RECURSION_DEPTH,
+            "tree-sitter depth guard fired: recursive walk truncated; some imports may be missing"
+        );
         return;
     }
     let mut cursor = node.walk();
@@ -166,6 +174,7 @@ impl LanguageAnalyzer for CAnalyzer {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use tracing_test::traced_test;
 
     #[test]
     fn extracts_multiple_functions() {
@@ -449,5 +458,65 @@ void guarded(int x) {
         let analyzer = CAnalyzer;
         let calls = analyzer.extract_calls(source).unwrap();
         assert!(calls.is_empty());
+    }
+
+    /// Depth-guard warning: deeply-nested preprocessor blocks must emit the warning when truncated.
+    #[test]
+    #[traced_test]
+    fn it_emits_depth_guard_warning_on_deeply_nested_preproc_blocks() {
+        const NESTING_DEPTH: usize = 300;
+
+        let mut source = String::new();
+        for i in 0..NESTING_DEPTH {
+            source.push_str(&format!("#ifdef MACRO_{i}\n"));
+        }
+        source.push_str("void deep_fn(void) {}\n");
+        for _ in 0..NESTING_DEPTH {
+            source.push_str("#endif\n");
+        }
+
+        let analyzer = CAnalyzer;
+        let _ = analyzer.extract_functions(source.as_bytes());
+        assert!(logs_contain("depth guard fired"));
+    }
+
+    /// Triangulation: shallow input must NOT emit the depth-guard warning.
+    #[test]
+    #[traced_test]
+    fn it_does_not_emit_depth_guard_warning_on_shallow_functions() {
+        let source = b"void foo(void) {}\nvoid bar(void) {}\n";
+        let analyzer = CAnalyzer;
+        let _ = analyzer.extract_functions(source);
+        assert!(!logs_contain("depth guard fired"));
+    }
+
+    /// Depth-guard warning: deeply-nested preprocessor blocks in collect_imports must emit warning.
+    #[test]
+    #[traced_test]
+    fn it_emits_depth_guard_warning_on_deeply_nested_preproc_in_imports() {
+        const NESTING_DEPTH: usize = 300;
+
+        let mut source = String::new();
+        for i in 0..NESTING_DEPTH {
+            source.push_str(&format!("#ifdef MACRO_{i}\n"));
+        }
+        source.push_str("#include <deep.h>\n");
+        for _ in 0..NESTING_DEPTH {
+            source.push_str("#endif\n");
+        }
+
+        let analyzer = CAnalyzer;
+        let _ = analyzer.extract_imports(source.as_bytes());
+        assert!(logs_contain("depth guard fired"));
+    }
+
+    /// Triangulation: shallow import input must NOT emit the depth-guard warning.
+    #[test]
+    #[traced_test]
+    fn it_does_not_emit_depth_guard_warning_on_shallow_imports() {
+        let source = b"#include <stdio.h>\n#include <stdlib.h>\n";
+        let analyzer = CAnalyzer;
+        let _ = analyzer.extract_imports(source);
+        assert!(!logs_contain("depth guard fired"));
     }
 }
--- a/src/treesitter/c_lang.rs
+++ b/src/treesitter/c_lang.rs
@@ -40,6 +40,7 @@ fn collect_functions(
         tracing::warn!(
             depth_limit = MAX_RECURSION_DEPTH,
             language = "c",
+            operation = "functions",
             "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
         );
         return;
@@ -102,6 +103,7 @@ fn collect_imports(
         tracing::warn!(
             depth_limit = MAX_RECURSION_DEPTH,
             language = "c",
+            operation = "imports",
             "tree-sitter depth guard fired: recursive walk truncated; some imports may be missing"
         );
         return;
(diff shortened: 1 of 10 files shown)


src/treesitter/typescript.rs
```diff
@@ -61,6 +61,11 @@ fn extract_functions_from_node(
...
 ) {
     if depth >= MAX_RECURSION_DEPTH {
+        tracing::warn!(
+            depth_limit = MAX_RECURSION_DEPTH,
+            language = "typescript",
+            "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
+        );
         return;
     }
     let mut cursor = node.walk();
```
mikelane, line 76: **CRITICAL [OBSERVABILITY]** — `language = "typescript"` is hardcoded in `extract_functions_from_node`, which is shared by all three `JsDialect` variants (TypeScript, TSX, JavaScript). When the depth guard fires on a `.tsx` or `.js` file, the emitted OTLP log record will say `language = "typescript"` regardless of the actual dialect. An operator querying `language = "javascript"` will silently miss these events.

`extract_functions_from_node` needs a `language: &str` parameter threaded in from each `LanguageAnalyzer` impl at the call site in `extract_functions`. The `JsDialect` enum already carries the information — it just is not plumbed through.

Suggested fix — derive the dialect label in `extract_functions` and thread it down:
```rust
let lang = match self.dialect {
    JsDialect::TypeScript => "typescript",
    JsDialect::Tsx        => "tsx",
    JsDialect::JavaScript => "javascript",
};
extract_functions_from_node(source, &root, None, &mut functions, 0, lang);
```
This is a correctness defect for the stated goal of this PR (queryable structured observability attributes).

src/treesitter/rust_lang.rs
```diff
@@ -327,6 +333,41 @@ use anyhow::Result;
...
+        let _ = analyzer.extract_functions(source.as_bytes());
+        assert!(logs_contain("depth guard fired"));
+    }
+
+    /// Triangulation: shallow input must NOT emit the depth-guard warning.
+    #[test]
+    #[traced_test]
+    fn it_does_not_emit_depth_guard_warning_on_shallow_input() {
+        let source = b"fn foo() {}\nfn bar() {}\n";
+        let analyzer = RustAnalyzer;
```
mikelane, line 369: **HIGH [TEST]** — `logs_contain("depth guard fired")` only checks whether the warn fired at all. It does not verify the `language` field. A mutation changing `language = "rust"` to `language = "bogus"` passes this test. Consider asserting the structured field value too:
```rust
assert!(logs_contain("language=rust"));
```
or combining both into a single assertion on the full message substring:
```rust
assert!(logs_contain("depth guard fired") && logs_contain("language=rust"));
```
This same weakness applies to all 9 positive `#[traced_test]` sites across c_lang, cpp, csharp, python, ruby, rust_lang, and typescript.

src/treesitter/c_lang.rs
```diff
@@ -37,6 +37,11 @@ fn collect_functions(
     depth: usize,
 ) {
     if depth >= MAX_RECURSION_DEPTH {
+        tracing::warn!(
+            depth_limit = MAX_RECURSION_DEPTH,
+            language = "c",
+            "tree-sitter depth guard fired: recursive walk truncated; some functions may be missing"
+        );
```
mikelane, line 45: **MEDIUM [OBSERVABILITY]** — C has two guard sites: one in `collect_functions` ("some functions may be missing") and one in `collect_imports` ("some imports may be missing"). The difference is only in the human-readable message string, not in any queryable structured attribute. An OTLP dashboard cannot filter on `operation = "functions"` vs `operation = "imports"` — it would have to parse the message text.

Consider adding an `operation` field:
```rust
tracing::warn!(
    depth_limit = MAX_RECURSION_DEPTH,
    language = "c",
    operation = "functions",
    "tree-sitter depth guard fired: ..."
);
```
This applies equally to the two C++ guard sites.

src/treesitter/ruby.rs
```diff
@@ -412,4 +449,37 @@ end
             "expected partial extraction to include outer classes"
         );
```
mikelane, line 455: **LOW [NAMING]** — `TEST_STACK_SIZE` here (and in `csharp.rs`) does not match `CONSTRAINED_THREAD_STACK_BYTES` used in four other files introduced by this same PR (rust_lang, python, cpp, c_lang, typescript). Two names for the same concept in code written in the same PR is a consistency gap.

`CONSTRAINED_THREAD_STACK_BYTES` is more precise — it communicates both the constraint intent and the unit. Prefer it uniformly.

mikelane (commented):
## Review: observability: emit tracing warning when tree-sitter depth guard fires

### Summary

The intent of this PR is correct and the implementation is structurally sound. Adding `tracing::warn!` at every depth-guard site turns a silent truncation into a queryable OTLP event — exactly right for operators debugging why function extraction returned partial results. The `#[cfg(not(test))]` gate on `set_meter_provider` and the subscriber initialization is a real correctness fix, not a workaround. The test coverage pattern (positive `#[traced_test]`, negative triangulation, constrained-thread overflow guard) is consistent and well-documented.

One correctness defect in the observability data must be fixed before merge. Two improvements are strong recommendations.

---

### Blockers (must fix before merge)

**CRITICAL — `language = "typescript"` hardcoded for all three JS dialects** (`src/treesitter/typescript.rs` lines 64–69, inline comment posted)

`extract_functions_from_node` is a free function shared by all three `JsDialect` variants. The `language` field is hardcoded to `"typescript"` regardless of whether the file being analyzed is `.tsx` or `.js`. When the guard fires on a JavaScript file, the exported OTLP record says `language = "typescript"`. An operator querying `language = "javascript"` will silently receive no results.

Fix: add `language: &str` as a parameter to `extract_functions_from_node`, derive it from `self.dialect` in `extract_functions`, and thread it through. The `JsDialect` enum already holds the information.

---

### Recommendations (strong — address before merge)

**HIGH — `logs_contain("depth guard fired")` does not verify the `language` field** (inline comment posted on `rust_lang.rs`)

All 9 positive `#[traced_test]` assertions confirm the warn fired but do not assert the structured `language` field value. A mutation replacing `language = "rust"` with `language = "bogus"` passes all 601 tests. After fixing the `language = "typescript"` blocker above, strengthen all 9 sites:
```rust
assert!(logs_contain("depth guard fired") && logs_contain("language=rust"));
```

**MEDIUM — No `operation` field on C/C++ dual guard sites** (inline comment posted on `c_lang.rs`)

C and C++ each have two guard sites (`collect_functions` and `collect_imports`). The difference is expressed only in the human-readable message string. An OTLP query cannot distinguish a truncated function scan from a truncated import scan without parsing message text. Adding `operation = "functions"` / `operation = "imports"` at the four affected sites costs nothing and makes the events properly filterable.

---

### Notes (low priority, author's discretion)

- `src/treesitter/typescript.rs:136` — `value.unwrap()` after `is_arrow` is logically safe but non-local. Prefer `if let Some(arrow_node) = value.filter(|v| v.kind() == "arrow_function")`. Pre-existing code touched by this PR.
- `TEST_STACK_SIZE` in `ruby.rs:422` and `csharp.rs:451` vs `CONSTRAINED_THREAD_STACK_BYTES` in the five other files in this PR. `CONSTRAINED_THREAD_STACK_BYTES` is the more precise name; inline comment posted on `ruby.rs`.
- `NESTING_DEPTH` vs `GENERATED_NESTING_LEVELS` — two names for the same concept introduced in this PR. Prefer `GENERATED_NESTING_LEVELS` throughout.
- `src/telemetry.rs:166` — redundant inner `unsafe {}` block inside `unsafe fn clear_telemetry_env()`. The function body is already an unsafe context.
- `ruby.rs:447` — overflow test asserts `!functions.is_empty()` while Python's equivalent asserts `functions.len() >= MAX_RECURSION_DEPTH`. Consider strengthening.

---

### Praise

The `#[cfg(not(test))]` gate on the subscriber initialization and `set_meter_provider` is the right fix. The comment at `telemetry.rs:127–131` accurately explains the three-way interaction between `#[traced_test]`'s `Once` guard, `set_meter_provider`'s OTel diagnostics, and the global subscriber — making a subtle interaction explicit rather than relying on implicit CI behavior.

The three-tier test structure (traced positive test at NESTING_DEPTH=300 to avoid thread-spawn limitations, constrained-thread overflow test at 5000, sequential triangulation at 255) is a well-reasoned design that covers the guard boundary, the overflow safety property, and the non-regression property independently.

---

### Decision: CHANGES REQUESTED

The `language = "typescript"` hardcoding ships incorrect OTLP data for JS and TSX files, directly contradicting the stated goal. Fix that and strengthen the `logs_contain` assertions before merging.

PR #210, opened 2026-04-12
fix(telemetry): switch OTLP exporter from gRPC/tonic to HTTP/protobuf

### LINKED ISSUE #209: fix(telemetry): switch OTLP exporter from gRPC/tonic to HTTP/protobuf
## Problem

The gRPC tonic exporter (`with_tonic()`) silently fails to establish a TCP connection to the OTLP collector. No data reaches SigNoz despite:
- Port 4317 being reachable (nc confirms TCP)
- The env var being correctly set on the process
- The collector accepting gRPC connections

Root cause: tonic uses H2C (HTTP/2 cleartext) which requires an upgrade handshake. OrbStack's port-forwarding proxy does not properly forward H2C upgrade headers, so every gRPC connection attempt silently fails.

Confirmed: HTTP OTLP on port 4318 returns 200 and data flows correctly.

## Fix

1. Add `http-proto` feature to `opentelemetry-otlp` in `Cargo.toml`
2. Change `with_tonic()` to `with_http()` in `src/telemetry.rs` for both trace and metrics exporters
3. Remove `grpc-tonic` feature (no longer needed)
4. Update CLAUDE.md and CLAUDE.local.md to document that the correct endpoint is `http://localhost:4318`

## Config change needed

After fix, update `~/.claude.json` MCP server env:
```json
"GIT_PRISM_OTLP_ENDPOINT": "http://localhost:4318"
```

## Acceptance

- `cargo test` passes
- Running the MCP server with `GIT_PRISM_OTLP_ENDPOINT=http://localhost:4318` and making tool calls results in metrics appearing in SigNoz within 60s

mikelane:
## Summary

- Switches `opentelemetry-otlp` from `grpc-tonic` to `http-proto` feature — removes the entire tonic/h2/tower/prost stack
- Changes both `SpanExporter` and `MetricExporter` builders from `.with_tonic()` to `.with_http()` in `src/telemetry.rs`
- Fixes shutdown error handling: `Drop` now logs flush failures instead of silently swallowing them with `let _ = ...`
- Adds startup confirmation message so operators can verify telemetry is active
- Updates test dummy endpoints from port 4317 (gRPC) to 4318 (HTTP) for documentation accuracy

## Root cause

The gRPC/tonic exporter uses H2C (HTTP/2 cleartext upgrade). OrbStack's port-forwarding proxy does not forward the H2C upgrade handshake, so every gRPC export attempt silently fails — no TCP connection was ever established despite the OTLP collector running and port 4317 being reachable via TCP.

HTTP OTLP on port 4318 is confirmed working (200 response).

## Config change required

Update `~/.claude.json` git-prism MCP server env after merge:
```json
"GIT_PRISM_OTLP_ENDPOINT": "http://localhost:4318"
```

## Test plan

- [ ] All 606 tests pass (verified by pre-push hook)
- [ ] Clippy clean, fmt clean
- [ ] After merge + binary rebuild: set `GIT_PRISM_OTLP_ENDPOINT=http://localhost:4318` and confirm metrics appear in SigNoz within 60s

Closes #209

### DIFF
--- a/bdd/steps/telemetry_steps.py
+++ b/bdd/steps/telemetry_steps.py
@@ -250,14 +250,6 @@ def _spawn_server(
             env.pop(key, None)
     if endpoint is not None:
         env["GIT_PRISM_OTLP_ENDPOINT"] = endpoint
-        # opentelemetry-otlp 0.28's HTTP exporter only appends `/v1/traces`
-        # and `/v1/metrics` when it reads the endpoint from
-        # `OTEL_EXPORTER_OTLP_ENDPOINT`. When the endpoint is provided via
-        # the `with_endpoint()` builder call (as production code does),
-        # the SDK posts to the endpoint as-is. Setting this env var lets
-        # the mock collector receive requests on the canonical signal
-        # paths, matching how a real OTLP backend would behave.
-        env["OTEL_EXPORTER_OTLP_ENDPOINT"] = endpoint
     proc = subprocess.Popen(  # noqa: S603 -- test harness, inputs are fixed strings
         [context.binary_path, "serve"],
         stdin=subprocess.PIPE,
--- a/bdd/steps/telemetry_steps.py
+++ b/bdd/steps/telemetry_steps.py
@@ -1,13 +1,10 @@
 """Step definitions for the OpenTelemetry (metrics + traces) feature.
 
-The production telemetry module does not exist yet — these scenarios are
-tagged @not_implemented and must FAIL with assertion errors until the
-`src/telemetry.rs` module and instrumentation sites land.
-
-This file also defines a minimal in-memory OTLP gRPC collector. It
-implements the two OTLP service servicers (`TraceServiceServicer` and
-`MetricsServiceServicer`) from `opentelemetry-proto` and records every
-export request it receives. Steps assert against that recorded state.
+This file defines a minimal in-memory OTLP HTTP/protobuf collector. It
+accepts POSTs to `/v1/traces` and `/v1/metrics`, parses the binary
+protobuf bodies into `ExportTraceServiceRequest` /
+`ExportMetricsServiceRequest` messages, and records every received
+request so step definitions can assert against the resulting state.
 """
 
 from __future__ import annotations
@@ -15,22 +12,16 @@ from __future__ import annotations
 import json
 import os
 import subprocess
+import threading
 import time
-from concurrent import futures
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from threading import Lock
 from typing import Any
 
-import grpc
 from behave import given, then, when
 from behave.runner import Context
-from opentelemetry.proto.collector.metrics.v1 import (
-    metrics_service_pb2,
-    metrics_service_pb2_grpc,
-)
-from opentelemetry.proto.collector.trace.v1 import (
-    trace_service_pb2,
-    trace_service_pb2_grpc,
-)
+from opentelemetry.proto.collector.metrics.v1 import metrics_service_pb2
+from opentelemetry.proto.collector.trace.v1 import trace_service_pb2
 
 VALID_REF_PATTERNS = {
     "worktree",
@@ -43,44 +34,42 @@ VALID_REF_PATTERNS = {
 
 
 class MockOtlpCollector:
-    """In-memory OTLP gRPC collector for assertion use in BDD tests.
+    """In-memory OTLP HTTP/protobuf collector for assertion use in BDD tests.
 
-    Implements the two OTLP service servicers that a real OTLP backend
-    exposes, storing every received export request in memory so step
-    definitions can assert on the state the server emitted.
+    Runs a tiny ThreadingHTTPServer that accepts POSTs to `/v1/traces`
+    and `/v1/metrics`, parses the binary protobuf bodies into the
+    standard `Export*ServiceRequest` messages, and stores every received
+    request so step definitions can assert on the state the server emitted.
     """
 
     def __init__(self) -> None:
         self._lock = Lock()
         self._trace_requests: list[Any] = []
         self._metric_requests: list[Any] = []
-        self.server: grpc.Server | None = None
+        self._server: ThreadingHTTPServer | None = None
+        self._thread: threading.Thread | None = None
         self.port: int | None = None
 
     def start(self) -> None:
-        """Bind a grpc.Server to a free port and register both servicers."""
-        self.server = grpc.server(futures.ThreadPoolExecutor(max_workers=4))
-        trace_service_pb2_grpc.add_TraceServiceServicer_to_server(
-            _TraceServicer(self), self.server,
+        """Bind a ThreadingHTTPServer to a free port and serve in the background."""
+        collector = self
+        handler_cls = _make_handler(collector)
+        self._server = ThreadingHTTPServer(("localhost", 0), handler_cls)
+        self.port = self._server.server_address[1]
+        self._thread = threading.Thread(
+            target=self._server.serve_forever,
+            daemon=True,
         )
-        metrics_service_pb2_grpc.add_MetricsServiceServicer_to_server(
-            _MetricsServicer(self), self.server,
-        )
-        self.port = self.server.add_insecure_port("localhost:0")
-        if self.port == 0:
-            raise RuntimeError("gRPC mock collector failed to bind")
-        self.server.start()
-        # Verify the server is actually accepting connections before returning
-        channel = grpc.insecure_channel(f"localhost:{self.port}")
-        try:
-            grpc.channel_ready_future(channel).result(timeout=5)
-        finally:
-            channel.close()
+        self._thread.start()
 
     def stop(self) -> None:
-        if self.server is not None:
-            self.server.stop(grace=1).wait()
-            self.server = None
+        if self._server is not None:
+            self._server.shutdown()
+            self._server.server_close()
+            self._server = None
+        if self._thread is not None:
+            self._thread.join(timeout=5)
+            self._thread = None
 
     def record_trace(self, request: Any) -> None:
         with self._lock:
@@ -122,22 +111,45 @@ class MockOtlpCollector:
         return out
 
 
-class _TraceServicer(trace_service_pb2_grpc.TraceServiceServicer):
-    def __init__(self, collector: MockOtlpCollector) -> None:
-        self._collector = collector
+def _make_handler(collector: MockOtlpCollector) -> type[BaseHTTPRequestHandler]:
+    """Build a request handler class bound to a specific collector instance."""
 
-    def Export(self, request, context):  # noqa: N802 -- gRPC protocol requires this name
-        self._collector.record_trace(request)
-        return trace_service_pb2.ExportTraceServiceResponse()
+    class _OtlpHttpHandler(BaseHTTPRequestHandler):
+        def do_POST(self) -> None:  # noqa: N802 -- BaseHTTPRequestHandler protocol
+            content_length = int(self.headers.get("Content-Length", "0") or "0")
+            body = self.rfile.read(content_length) if content_length > 0 else b""
 
+            if self.path == "/v1/traces":
+                request = trace_service_pb2.ExportTraceServiceRequest()
+                request.ParseFromString(body)
+                collector.record_trace(request)
+                response = trace_service_pb2.ExportTraceServiceResponse()
+                self._send_protobuf(response.SerializeToString())
+                return
 
-class _MetricsServicer(metrics_service_pb2_grpc.MetricsServiceServicer):
-    def __init__(self, collector: MockOtlpCollector) -> None:
-        self._collector = collector
+            if self.path == "/v1/metrics":
+                request = metrics_service_pb2.ExportMetricsServiceRequest()
+                request.ParseFromString(body)
+                collector.record_metrics(request)
+                response = metrics_service_pb2.ExportMetricsServiceResponse()
+                self._send_protobuf(response.SerializeToString())
+                return
 
-    def Export(self, request, context):  # noqa: N802 -- gRPC protocol requires this name
-        self._collector.record_metrics(request)
-        return metrics_service_pb2.ExportMetricsServiceResponse()
+            self.send_response(404)
+            self.end_headers()
+
+        def _send_protobuf(self, body: bytes) -> None:
+            self.send_response(200)
+            self.send_header("Content-Type", "application/x-protobuf")
+            self.send_header("Content-Length", str(len(body)))
+            self.end_headers()
+            self.wfile.write(body)
+
+        def log_message(self, format: str, *args: Any) -> None:  # noqa: A002 -- BaseHTTPRequestHandler protocol
+            # Silence default stderr access logging during tests.
+            return
+
+    return _OtlpHttpHandler
 
 
 # ---------- helpers used by the steps ----------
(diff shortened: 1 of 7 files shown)


src/telemetry.rs
```diff
@@ -141,6 +165,8 @@ pub fn init() -> TelemetryGuard {
         }
     }
 
+    eprintln!("git-prism: telemetry initialized (HTTP/protobuf, endpoint={base})");
```
mikelane, line 220: **BLOCKER — OBSERVABILITY HIGH: Darkness dressed as light**

This success message is unconditional — it fires regardless of whether `Registry::default().with(otel_layer).try_init()` at line 163 actually succeeded.

The failure path sequence currently looks like this:

1. Line 163: `try_init()` fails (the most common real-world cause is "a subscriber is already registered" — e.g., in tests, in re-entrant init, in any future binary that composes `git-prism` as a library, or when rmcp itself installs a subscriber).
2. Line 164: `eprintln!` writes `"git-prism: failed to initialize tracing subscriber: ..."` to stderr.
3. **Line 168 (this line)**: `eprintln!` writes `"git-prism: telemetry initialized (HTTP/protobuf, endpoint=...)"` to stderr *immediately after the failure message*.
4. Lines 170-173: a `TelemetryGuard` is returned with `tracer_provider: Some(_)` and `meter_provider: Some(_)`.

The operator sees contradictory stderr output and a guard that looks fully active, but the `tracing_opentelemetry::layer` was never attached to any subscriber — so every `tracing::info_span!` / `tracing::event!` call in the MCP tool handlers falls on the floor. Meanwhile `set_meter_provider` (line 156) may have succeeded independently, so metrics *still flow*. The resulting partial-init state is actively hostile to debugging: the operator sees metrics arriving at SigNoz, concludes telemetry works, then spends hours wondering why traces are empty.

Compare this to how the two exporter-build failures at lines 95-126 are handled — those print a single clear error and return a no-op guard. Subscriber init failure deserves the same rigor.

Suggested fix (one of):

1. Move the success `eprintln!` inside the `Ok(_)` arm of `try_init()`, and have the `Err(_)` arm print `"git-prism: telemetry disabled: subscriber init failed"` and return a no-op guard.
2. Return `Result<TelemetryGuard, TelemetryInitError>` from `init()` and let `server.rs:590` decide how to handle a failed init explicitly.
3. At a minimum, differentiate the success message text so it says `"providers built"` rather than `"telemetry initialized"` when the subscriber was not attached.

Either way, never print "initialized" after a failure.

src/telemetry.rs
```diff
@@ -250,13 +280,27 @@ mod tests {
...
+        assert_eq!(traces, "http://localhost:4318/v1/traces");
+        assert_eq!(metrics, "http://localhost:4318/v1/metrics");
+    }
+
+    #[test]
+    fn it_appends_signal_paths_to_a_bare_base() {
+        let (traces, metrics) = signal_endpoints("http://localhost:4318");
+        assert_eq!(traces, "http://localhost:4318/v1/traces");
+        assert_eq!(metrics, "http://localhost:4318/v1/metrics");
+    }
```
mikelane, line 347: **BLOCKER — TEST CRITICAL: Missing regression test for subscriber-init failure**

These two new tests exercise `signal_endpoints` well — the URL construction path is covered.

What is NOT covered: the failure path at lines 163-168 where `try_init()` fails. Every existing telemetry test (`test_init_with_endpoint_creates_providers`, `test_guard_drop_does_not_panic`, `test_init_with_custom_service_name_succeeds`) asserts only on `guard.is_active()`, which in turn only checks `tracer_provider.is_some()`. A mutation that skipped the subscriber init entirely — or a regression where `try_init()` silently returns `Err` — would pass every existing test, because the guard still contains `Some(tracer_provider)`.

This is the test-side regression vector for the blocker on lines 163-168. When that blocker is fixed, the fix needs to come with a test that would have caught the regression.

Design suggestions:

1. **Return `Result` from `init()`**. Then a test can call `init()` twice on the same thread (or install a subscriber beforehand) and assert the second call returns a specific error variant. Clean, mechanical, no flakiness.
2. **Split "build providers" from "attach subscriber"** into two functions. The first is hermetic and unit-testable. The second is a thin production-only wrapper. Currently `#[cfg(test)]` at line 155 blocks the subscriber init path entirely in tests precisely because it cannot be made hermetic — that gating is a signal that the abstraction is wrong.
3. **At minimum**, add a test that captures stderr during `init()` and asserts that "telemetry initialized" never appears after "failed to initialize tracing subscriber" in the output stream. A string-level smoke test is ugly, but it would catch the regression.

Tangential: `is_active()` at line 46 is `#[cfg(test)]` — if the fix exposes this publicly so `server.rs` can distinguish "intentionally disabled" from "requested but failed", that also gives tests a richer oracle to work with.

src/telemetry.rs
```diff
@@ -62,15 +83,18 @@ pub fn init() -> TelemetryGuard {
         }
     };
 
+    let base = endpoint.trim_end_matches('/');
+    let (traces_endpoint, metrics_endpoint) = signal_endpoints(base);
```
mikelane, line 135: **WARNING — RUST MEDIUM: Redundant `trim_end_matches`**

The trim happens twice: once here on line 86 where `base` is assigned, and again inside `signal_endpoints` at line 29 (`let trimmed = base.trim_end_matches('/');`). This is a minor code smell — not incorrect, just a small DRY violation that raises the question "which layer is responsible for normalization?"

Two options:

1. **Remove the outer trim on line 86.** Pass `endpoint.as_str()` directly to `signal_endpoints`, and keep the log message on line 168 printing `endpoint` (the un-trimmed value). Cost: the `"telemetry initialized (endpoint=...)"` log may contain a trailing slash. Benefit: the helper is the single source of truth.
2. **Remove the inner trim inside `signal_endpoints`** and document in the helper that callers must pre-trim. Cost: the helper becomes partial — a naive caller passing `"http://host:4318/"` gets `"http://host:4318//v1/traces"`. Benefit: one less function call per init.

Option 1 is the Rust-idiomatic choice (helpers enforce their own preconditions; callers should not need to know the helper's internals), and it also happens to be the cheaper fix.

While you are in there: the helper doc comment (lines 19-27) says the helper "constructs the full signal URLs explicitly so a user-supplied `GIT_PRISM_OTLP_ENDPOINT=http://collector:4318` reaches the canonical signal paths" — but it does not spell out the pre-condition contract. If option 2 is picked, that contract needs to be stated in the doc comment.

bdd/steps/telemetry_steps.py
```diff
@@ -43,44 +34,42 @@
...
-        if self.server is not None:
-            self.server.stop(grace=1).wait()
-            self.server = None
+        if self._server is not None:
+            self._server.shutdown()
+            self._server.server_close()
+            self._server = None
+        if self._thread is not None:
+            self._thread.join(timeout=5)
+            self._thread = None
```
mikelane, line 72: **WARNING — PYTHON HIGH: Handler-thread lifetime hazard**

`ThreadingHTTPServer.shutdown()` on line 67 stops the accept loop, but it does not join the worker threads that are already handling in-flight requests. If a POST is mid-`ParseFromString` when `stop()` is called, the worker thread keeps running after `stop()` returns, and the daemon-ness of `self._thread` (the accept loop) on line 62 does not help — the worker threads are spawned by `ThreadingHTTPServer` itself with `daemon=False` by default.

Two consequences in the BDD suite:

1. **Cross-scenario contamination.** If scenario A's collector is stopped while a worker is still parsing a late export, that worker can survive into scenario B and call `collector.record_trace(...)` on the (now `None`-ified) old collector reference. Depending on timing, this surfaces as `AttributeError` or as silently-lost exports that "should have arrived in scenario B."
2. **Zombie threads on test-runner exit.** `behave` will not exit cleanly if non-daemon threads are alive. CI masks this as a hang until the runner hard-kills the process.

Fix: set `daemon_threads = True` on the server in `start()`:

```python
self._server = ThreadingHTTPServer(("localhost", 0), handler_cls)
self._server.daemon_threads = True  # <-- add this
```

`ThreadingHTTPServer` inherits from `socketserver.ThreadingMixIn` which exposes `daemon_threads` exactly for this case. With it set to `True`, in-flight worker threads are daemonized, so they cannot outlive the test process and cannot block `behave` exit.

Related: `do_POST` (line 118) has no try/except around `request.ParseFromString(body)`. A malformed protobuf body — or a protocol-version skew where the wire format no longer matches `ExportTraceServiceRequest` — raises `google.protobuf.message.DecodeError` out of `do_POST`, which `BaseHTTPRequestHandler` catches and logs via `log_message` (but that is overridden to be silent at line 148). Net result: the connection drops, the client sees an HTTP error, the test fails downstream with "no metric found" — and the real parse error is invisible. Wrap `ParseFromString` in a try/except and return 400 with the error text so a future protobuf schema drift fails loudly instead of silently.

mikelane (commented):
# Review: fix(telemetry) — HTTP/protobuf migration

**Verdict: REQUEST CHANGES** (posted as `COMMENT` because GitHub does not allow self-approval on owner-authored PRs — the author should treat this review as blocking.)

CI is green across all 16 checks and the transport migration itself is correct. However, the observability story on the Rust side has a high-severity partial-init bug, the test suite cannot catch a regression of it, and the BDD harness has a subprocess-pipe deadlock risk that is a textbook flaky-CI time bomb. There are also two stale-documentation blockers that reference the old gRPC transport.

All of this is approximately two hours of focused work.

---

## Blockers (must fix before merge)

### B1. Darkness dressed as light — subscriber init failure is silently masked

**Location:** `src/telemetry.rs:163-168`

See inline comment on line 168. Summary: when `Registry::default().with(otel_layer).try_init()` fails, the function prints the error to stderr, then unconditionally prints `"telemetry initialized (HTTP/protobuf, endpoint=...)"` on the next line, then returns a `TelemetryGuard` whose provider fields are both `Some(_)`. The tracing layer was never attached, so every `tracing::info_span!` call falls on the floor, while `set_meter_provider` may have succeeded — producing a partial-init state where metrics flow and traces do not, with no diagnostic hint about why.

The fix should mirror how the two exporter-build failures at lines 95-126 are handled: bail to a no-op guard with one clear error message.

### B2. Test suite has no oracle for subscriber attachment

**Location:** `src/telemetry.rs:283-295` (new `signal_endpoints` tests — good as far as they go, missing the regression coverage for B1)

See inline comment on the new test block. Summary: every existing test asserts only on `guard.is_active()`, which in turn only checks `tracer_provider.is_some()`. A mutation that broke subscriber attachment but still built providers would pass every test. This is the regression vector for B1 — the fix needs to come with a test that would have caught it.

### B3. Pipe-deadlock risk in `_send_tool_call`

**Location:** `bdd/steps/telemetry_steps.py:214-229` (not in the PR diff — the function is unchanged, but it is called from newly-modified code paths and the regression risk is activated by the transport swap)

The sequence `proc.stdout.readline(); proc.stdout.readline(); proc.stdin.close(); proc.wait(timeout=10)` is the canonical subprocess pipe deadlock:

1. Two `readline()` calls drain at most two lines of stdout.
2. `stdin.close()` triggers server shutdown.
3. `wait(timeout=10)` blocks without draining stdout/stderr.

The child's stderr now carries additional diagnostics from this PR — `"git-prism: telemetry initialized"` on line 168 of `src/telemetry.rs`, plus any telemetry-flush errors from the `Drop` impl on lines 59/64. The pipe buffer is 8 KiB on macOS and 64 KiB on Linux. Once full, the child blocks on write, the parent hits `TimeoutExpired`, and `proc.terminate()` races the `TelemetryGuard` `Drop` impl. On a slow CI runner the SIGTERM wins, the exporter never flushes, and the mock collector ends up with fewer metrics/spans than expected. The assertion failure downstream reads "no metric found" — a failure message that gives zero hint about the real cause.

**Fix:**

```python
proc.stdin.write(json.dumps(request) + "\n")
proc.stdin.flush()
try:
    stdout_data, stderr_data = proc.communicate(timeout=10)
except subprocess.TimeoutExpired:
    proc.kill()
    stdout_data, stderr_data = proc.communicate(timeout=5)
```

`communicate()` drains both pipes concurrently, so the child never blocks on a write. Then parse `stdout_data` as newline-delimited JSON-RPC and assert the initialize and tool-call responses are not error objects — the current code on lines 217-218 never inspects the payload, so a JSON-RPC error response would also surface downstream as "no metric found."

The `time.sleep(0.5)` on line 229 can go away at the same time: after `communicate()` returns, the child has exited, so telemetry has either flushed or it has not. The sleep accomplishes nothing except adding 0.5s * N scenarios of wall time.

### B4. Orphaned `grpcio` dependency

**Location:** `bdd/requirements.txt:2` (not in PR diff — but it should have been)

After this migration there are zero `import grpc` references anywhere under `bdd/`. `grpcio>=1.80.0` is a 30+ MB C-extension wheel that is no longer used. It is cost without benefit — slower CI bootstrap, more CVE surface area, and a confusing signal to future readers.

The CVE comment on line 4 (`# CVE-2024-7254 floor; transitive via opentelemetry-proto and grpcio`) also references `grpcio` — which is now misleading, since `grpcio` is no longer transitively pulling in `protobuf` if it is not installed at all.

**Fix:**

```diff
 behave>=1.2.6
-grpcio>=1.80.0
 opentelemetry-proto>=1.29.0
-# CVE-2024-7254 floor; transitive via opentelemetry-proto and grpcio
+# CVE-2024-7254 floor; transitive via opentelemetry-proto
 protobuf>=5.29.2
```

Worth confirming with `rg "^(import|from) grpc" bdd/` before merging that nothing else in the BDD harness is still importing `grpc` through a path the surface scan missed.

### B5. Feature description documents the wrong transport

**Location:** `bdd/features/telemetry.feature:5` (not in PR diff)

Current:

```gherkin
The git-prism MCP server emits OpenTelemetry metrics and traces over OTLP
gRPC when configured via environment variables. Telemetry is off by default.
```

After this PR the server emits over HTTP/protobuf. A BDD spec that lies about the system under test is worse than no spec — it fails the contract the BDD suite exists to enforce. Change `OTLP gRPC` to `OTLP HTTP/protobuf`. One-line edit.

---

## Warnings (should fix or justify)

| # | Location | Severity | Issue |
|---|----------|----------|-------|
| W1 | `src/telemetry.rs:74` | OBS HIGH | `init()` is infallible — `server.rs:590` cannot distinguish "intentionally disabled" from "requested but failed." Expose `is_active()` (currently `#[cfg(test)]` on line 46), or return `Result<TelemetryGuard, TelemetryInitError>`. Pairs with B1. |
| W2 | `src/telemetry.rs:59, 63, 103, 120, 168` | OBS HIGH | All telemetry diagnostics use `eprintln!` instead of `tracing::warn!`/`tracing::error!`. For an MCP stdio server, stderr is fragile — parent processes may discard it, so telemetry failures become invisible. Exception: line 164 itself cannot use `tracing` because the subscriber is the thing that failed. Everything else can, since `tracing` is already a dependency. |
| W3 | `src/telemetry.rs:258-281` | TEST CRIT | `test_guard_drop_does_not_panic` — "does not panic" is the weakest possible assertion because panics propagate naturally. The test has no `assert!` after the drops. A silent file-handle leak or mid-flush hang would pass. Rename, or add an assertion that inspects flush state. |
| W4 | `src/telemetry.rs:298-314` | TEST CRIT | `test_init_with_custom_service_name_succeeds` — name claims "custom service name" behavior, body only asserts `guard.is_active()`. A mutation that ignored `GIT_PRISM_SERVICE_NAME` entirely would pass. Either rename the test, or read back the resource attributes and assert the custom name is present. |
| W5 | `bdd/steps/telemetry_steps.py:65-72` | PY HIGH | `MockOtlpCollector.stop()` — see inline comment. `ThreadingHTTPServer.shutdown()` does not join worker threads; set `daemon_threads = True` on the server. |
| W6 | `bdd/steps/telemetry_steps.py:118-139` | PY HIGH | `do_POST` has no try/except around `ParseFromString`. A malformed protobuf or version skew propagates out of the handler and is silenced by the overridden `log_message` — downstream failure reads as "no metric found." Wrap and return 400 with the error text. |
| W7 | `bdd/steps/telemetry_steps.py:229, 417` | PY HIGH | Load-bearing `time.sleep(0.5)` / `time.sleep(1)`. Classic CI flake source. Replace with a bounded poll: `wait_for(lambda: len(collector.get_metric_requests()) >= expected, timeout=5)`. |
| W8 | `src/telemetry.rs:86-87` | RUST MED | Redundant `trim_end_matches` — see inline comment. Helper is the right home for normalization. |
| W9 | `src/telemetry.rs:284-295` | TEST WARN | New `signal_endpoints` tests only cover bare base and single trailing slash. Missing: path-prefixed base (`http://collector:4318/otlp` — does it become `/otlp/v1/traces`?), multiple trailing slashes, empty string, HTTPS. The doc comment's contract ("reaches the canonical signal paths that real OTLP backends expect") is broader than the tests verify. |

---

## Praise

- **Drop impl error logging is a real improvement.** Was `let _ = ...`, now logs on failure. That is exactly the right direction for telemetry observability of telemetry itself.
- **`signal_endpoints()` helper doc comment is excellent.** It explains the `opentelemetry-otlp` 0.28 gotcha that motivated the helper — future-me will thank present-you for writing that down instead of leaving a naked `format!` behind.
- **Env-var scrubbing in `_spawn_server`** (`bdd/steps/telemetry_steps.py:248-250`) is a nice defense against the local-dev-environment-leaks-into-tests failure mode, and the comment explains exactly why (`OTEL_EXPORTER_OTLP_*` has higher SDK precedence than `with_endpoint()`). Precisely the kind of foot-gun that BDD authors forget about until it bites in CI.
- **Scope discipline.** This is a targeted transport swap — no drive-by refactors, no speculative abstractions. Clean revert boundary.
- **Zero secret findings** on a scan of the diff. No hardcoded endpoints beyond `http://localhost:4318` in tests, no auth tokens, no accidental env-var dumps.
- **Test count grew** from 606 to 608 with the new `signal_endpoints` tests — which is the right direction even though those tests do not yet cover the regression in B1/B2.

---

## Decision

**REQUEST CHANGES.** B1 + B2 are tightly coupled observability regressions: in a common real-world condition (subscriber already registered), traces silently disappear and the diagnostic output actively lies about success. B3 is a flaky-CI time bomb. B4 and B5 are smaller but correctness-and-honesty issues that fit cleanly into the same fix pass.

None of this is hard to fix — the whole list is about two hours of focused work. Do B1 and B2 together (they share the same test), do B3, and sweep B4+B5 on the way out.

Issue #212, opened 2026-04-13
fix(tools): response-size guardrails and pagination for get_change_manifest and get_function_context


mikelane:
## Priority: HIGHEST

This is structural — neither of the two "cheap" tools currently have a response-size bound, and on real PRs they produce outputs that exceed MCP context limits. This is antithetical to the entire purpose of git-prism.

## Observed failures

**Case 1 — `get_change_manifest` with `include_function_analysis: true`: ~10.7k tokens**

On a real PR (saymore `origin/main..origin/issue-96-pass1-extraction`), the manifest alone returned 10,700 tokens. This is supposed to be the *cheapest first-resort tool*. Per-function signature strings balloon the response linearly with the number of changed functions. The tool paginates by *file count* (`page_size`), but function analysis inside each file entry is unbounded.

**Case 2 — `get_function_context`: 111,777 characters → rejected by MCP, dumped to disk**

Same session, `get_function_context` on the same range produced 111,777 characters — exceeding the MCP token limit entirely. Claude Code's fallback wrote the payload to a file on disk for the agent to read back. That is the literal worst case: an agent-optimized tool whose response does not fit in the agent's context, forcing a raw-file-read fallback. Worse than a `git diff`.

## Root cause

1. `get_change_manifest` paginates by file count only. `include_function_analysis: true` embeds unbounded function detail *per file*, so a 20-file PR with heavy function changes can easily blow past any token budget even at small `page_size`.
2. `get_function_context` has **no pagination at all**. It returns every changed function with every caller, callee, and test reference in one shot.
3. Neither tool emits the existing `git_prism.response.truncated` metric when the structural bound is hit, because no bound exists.

### `get_change_manifest`

1. **Flip `include_function_analysis` default to `false`.** Opt-in for detail. This is a breaking API change but it aligns the default with the "cheap first-resort" contract the tool description claims.
2. **Add a token/byte budget** for the response (default e.g. 8000 tokens). When adding a file entry would exceed the budget, truncate per-entry function detail to signatures only (no body/imports), and set `function_analysis_truncated: true` in metadata with a list of files that were trimmed.
3. **Emit `git_prism.response.truncated{reason="token_budget", tool="get_change_manifest"}`** whenever the budget is hit.

### `get_function_context`

1. **Add function-level pagination.** Current pagination is file-based; function context is unbounded inside. Cursor should be a `function_offset`, not a file offset.
[...]

PR #219, opened 2026-04-14
feat(metrics)!: add token_estimate helper and bounded truncation reasons

mikelane:
## Summary

PR 2 of 4 in the stack for #212 (response-size guardrails and pagination for `get_change_manifest` and `get_function_context`). This is the **shared infrastructure layer** — nothing in this PR enforces a budget yet, but PRs 3 and 4 now have the helpers they need to do so:

- New `src/tools/size.rs` module with `estimate_tokens(char_count)` and `estimate_response_tokens<T: Serialize>(&T)` helpers
- New `classify_truncation_reason` flat-match classifier in `src/privacy.rs` (mirrors `classify_error_kind`, per issue #212 Design Decision B)
- `Metrics::record_truncated` normalizes the `reason` label at the metric boundary, so `git_prism.response.truncated` cardinality is bounded by construction regardless of what the caller passes
- `token_estimate: usize` field added to `ManifestResponse::metadata` and `FunctionContextResponse::metadata`, populated via the two-pass `estimate_response_tokens` pattern in `build_manifest` / `build_worktree_manifest` / `build_function_context`
- `src/tools/snapshots.rs` refactored to use `size::estimate_tokens(total_chars)` instead of the inline `total_chars / 4`

## BDD RED → GREEN

Two scenarios from `bdd/features/response_size_guardrails.feature` transition out of `@not_implemented`:

- `Change manifest reports its payload size for budgeting follow-up calls`
- `Function context reports its payload size for budgeting follow-up calls`

Both now assert `metadata.token_estimate` is a positive integer on a non-trivial fixture. The remaining 11 scenarios in the feature stay tagged `@not_implemented` — they will be unblocked by PRs 3 and 4.

Verified locally:

```console
$ behave --tags="@ISSUE-212 and not @not_implemented" bdd/features/response_size_guardrails.feature
1 feature passed, 0 failed, 0 skipped
2 scenarios passed, 0 failed, 11 skipped
6 steps passed, 0 failed, 48 skipped
```

## Why `classify_truncation_reason` is a flat exact-match (not substring)

Unlike `classify_error_kind` — which substring-matches error messages coming from third-party libraries like `gix` and `tree-sitter` — `classify_truncation_reason` is called only from within the codebase, always with a known-good string literal. An exhaustive flat match means mutation testing can kill every arm with one test each, and the `_ => "unknown"` safety net catches future typos without unbounding the cardinality.

Known labels: `paginated`, `max_files`, `max_file_size`, `token_budget`. Only `paginated` and `max_file_size` are emitted by existing call sites (`src/server.rs:206,475`). `token_budget` is reserved for PRs 3 and 4; `max_files` is reserved for a future change to the file-count truncation path.

## Why `token_estimate` uses a two-pass pattern

`estimate_response_tokens` serializes the struct to JSON and divides byte length by 4, but the struct contains the estimate itself — a chicken-and-egg. The builders work around this by constructing the response with `token_estimate: 0`, computing the estimate, then writing it back into the metadata before returning. This undercounts by 1 token at worst (the byte delta between `"token_estimate":0` and `"token_estimate":<N>` crosses one 4-byte boundary on realistic-sized responses), which is acceptable for a budgeting hint. Documented at the call sites.

Same pattern appears three times (`build_manifest`, `build_worktree_manifest`, `build_function_context`). Per CLAUDE.md's "Three similar lines is better than a premature abstraction," the duplication is intentional — PRs 3 and 4 will add budget-enforcement logic that reshapes these call sites, and extracting a helper now would get un-extracted then.

## Gauntlet

Ran the full pre-review gauntlet (bug hunt → quality → security → 4 church purists: rust, test, observability, arch). Results:

- **Bug hunt**: CLEAN
- **Quality**: 1 medium (three-site duplication) deferred as above per CLAUDE.md rule; 3 low notes
- **Security**: 0 CRITICAL/HIGH/MEDIUM/LOW. 1 INFO — pre-existing `ManifestMetadata.repo_path` raw-path exposure is now materialized in a transient `serde_json::to_string` buffer during estimation. Not introduced by this PR, worth a follow-up
- **rust-purist**: CLEAN
- **test-purist**: strengthened `it_normalizes_unknown_truncation_reason_without_panicking` to also exercise `classify_truncation_reason` directly; added symmetric `it_reports_a_positive_token_estimate_for_a_non_trivial_context_response` in `context.rs`. Pushed as `56d714e`
- **observability-purist**: flagged a ≤1-token divergence between the `token_estimate` field and `git_prism.response.tokens_estimated` metric (server.rs recomputes on the final serialization, which contains the non-zero estimate string). Deferred to a follow-up issue — out of PR 2's scope, and Design Decision B says "zero signature changes at existing call sites"
- **arch-purist**: proposed moving `classify_truncation_reason` from `privacy.rs` to `metrics.rs` on cohesion grounds. Reverted — issue #212 Design Decision (B) explicitly chose `privacy.rs` and "mirror classify_error_kind"; a gauntlet agent does not override the ticket's design decisions

## Commits

1. `98a88d6` — `test(bdd): remove @not_implemented from token_estimate surfacing scenarios` (RED)
2. `c20a3f6` — `feat(tools): add size::estimate_tokens helper and estimate_response_tokens`
3. `b29cbe3` — `refactor(snapshots): use size::estimate_tokens for token estimate`
4. `300adef` — `feat(tools)!: surface token_estimate on manifest and function context metadata`
5. `f632634` — `feat(metrics,privacy): bound truncation reason label cardinality`
6. `56d714e` — `test(tools,metrics): lock in token_estimate wiring and truncation classifier call path`

## Verification

- [x] `cargo test --bin git-prism` — 630 passed, 0 failed
- [x] `behave --tags="@ISSUE-212 and not @not_implemented" bdd/features/response_size_guardrails.feature` — 2 scenarios pass, 11 skipped
- [x] `behave --tags="~@not_implemented"` — no regressions elsewhere in the BDD suite
- [x] Lefthook pre-push hook (fmt + clippy + test) green on every push

## Deferred follow-ups

These surfaced during the gauntlet but are out of scope for PR 2. Not blockers.

- **Field/metric divergence.** `server.rs::record_tokens_estimated` recomputes tokens from the final serialization; `ManifestMetadata.token_estimate` is computed before the write-back. They differ by ≤1 token. Consider making `server.rs` read from the field instead when PR 3 reshapes the budget path. (observability agent, CRITICAL per its framing)
- **`ManifestMetadata.repo_path` privacy.** The raw absolute path is in the response body and now also in a transient serialization buffer. Pre-existing on main. Consider hashing via `privacy::hash_repo_path` or dropping the field entirely. (security agent, INFO)
- **`SIZE_BUCKETS` shared between bytes and tokens histograms.** `git_prism.response.tokens_estimated` uses the same histogram buckets as `response_bytes`, which means the top tail is dead weight for tokens. Consider a separate `TOKEN_BUCKETS` at ~25% scale. (observability agent, MEDIUM)
- **Structured log alongside `record_truncated`.** `record_truncated` fires a metric but no `tracing::info!`. On-call can see *that* truncation happened but not *which* requests. Consider adding when PR 3 wires the `token_budget` label. (observability agent, LOW)
- **`response_truncated` span attribute as bool.** Will need to become a reason string in PR 3 when multiple truncation reasons coexist. (observability agent, LOW)

Part of #212.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -343,13 +343,39 @@ mod tests {
 
     #[test]
     fn it_normalizes_unknown_truncation_reason_without_panicking() {
-        // Indirect behavior check: an unrecognized reason string must pass
-        // through record_truncated cleanly, because classify_truncation_reason
-        // folds it to "unknown" at the metric boundary. The real per-arm
-        // assertion lives in src/privacy.rs::tests; this test locks in the
-        // wiring so a future refactor that drops the normalization call would
-        // fail to compile or panic here.
+        // The global meter is a no-op in unit tests, so we cannot read back
+        // the attribute value record_truncated emits. That makes the final
+        // three calls below a smoke test for the call path, NOT a substitute
+        // for the per-arm assertions on classify_truncation_reason that live
+        // in src/privacy.rs::tests.
+        //
+        // To keep the "record_truncated goes through the classifier" wiring
+        // honest without observing the meter, the assertions below also
+        // exercise the classifier directly on the same inputs. A mutation
+        // that removed the classify_truncation_reason call from
+        // record_truncated would leave the emitted label unbounded — and
+        // while THIS test cannot observe that directly, the type signature
+        // of `KeyValue::new("reason", normalized)` inside record_truncated
+        // requires a `&'static str`, so any mutant that tried to substitute
+        // the raw `reason: &str` parameter would fail to compile.
         let metrics = Metrics::new();
+
+        // Known labels must pass through unchanged.
+        assert_eq!(
+            crate::privacy::classify_truncation_reason("paginated"),
+            "paginated",
+        );
+        assert_eq!(
+            crate::privacy::classify_truncation_reason("token_budget"),
+            "token_budget",
+        );
+        // Unrecognized labels must fold to the "unknown" safety-net arm so
+        // metric cardinality stays bounded.
+        assert_eq!(
+            crate::privacy::classify_truncation_reason("wildly_unrecognized_reason_42"),
+            "unknown",
+        );
+
         metrics.record_truncated("test_tool", "wildly_unrecognized_reason_42");
         metrics.record_truncated("test_tool", "paginated");
         metrics.record_truncated("test_tool", "token_budget");
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -221,11 +221,16 @@ impl Metrics {
     }
 
     pub fn record_truncated(&self, tool: &str, reason: &str) {
+        // Normalize at the metric boundary so attribute cardinality on the
+        // `reason` label is bounded by construction regardless of what the
+        // caller passes. The classifier is a flat exact-match — see
+        // `crate::privacy::classify_truncation_reason` for the full rationale.
+        let normalized = crate::privacy::classify_truncation_reason(reason);
         self.response_truncated.add(
             1,
             &[
                 KeyValue::new("tool", tool.to_string()),
-                KeyValue::new("reason", reason.to_string()),
+                KeyValue::new("reason", normalized),
             ],
         );
     }
@@ -334,5 +339,19 @@ mod tests {
         ] {
             metrics.record_error("test_tool", kind);
         }
+    }
+
+    #[test]
+    fn it_normalizes_unknown_truncation_reason_without_panicking() {
+        // Indirect behavior check: an unrecognized reason string must pass
+        // through record_truncated cleanly, because classify_truncation_reason
+        // folds it to "unknown" at the metric boundary. The real per-arm
+        // assertion lives in src/privacy.rs::tests; this test locks in the
+        // wiring so a future refactor that drops the normalization call would
+        // fail to compile or panic here.
+        let metrics = Metrics::new();
+        metrics.record_truncated("test_tool", "wildly_unrecognized_reason_42");
+        metrics.record_truncated("test_tool", "paginated");
+        metrics.record_truncated("test_tool", "token_budget");
     }
 }
--- a/src/privacy.rs
+++ b/src/privacy.rs
@@ -94,6 +94,24 @@ pub fn classify_ref_mode(base_ref: &str, head_ref: Option<&str>) -> &'static str
     }
 }
 
+/// Maps raw truncation-reason strings to a bounded label set for metrics.
+///
+/// Unlike [`classify_error_kind`], which does substring matching on error
+/// messages coming from third-party libraries, this function uses a flat
+/// exact match because the caller is always the codebase itself passing a
+/// known-good literal. An exhaustive match means mutation testing can kill
+/// every arm cleanly, and `unknown` exists as a safety net for future
+/// refactors that might introduce a typo or a forgotten label.
+pub fn classify_truncation_reason(reason: &str) -> &'static str {
+    match reason {
+        "paginated" => "paginated",
+        "max_files" => "max_files",
+        "max_file_size" => "max_file_size",
+        "token_budget" => "token_budget",
+        _ => "unknown",
+    }
+}
+
 /// Maps error description strings to a bounded label set for metrics.
 pub fn classify_error_kind(err: &str) -> &'static str {
     let lower = err.to_lowercase();
@@ -337,5 +355,38 @@ mod tests {
         assert_eq!(classify_ref_mode("main", Some("HEAD")), "branch");
         assert_eq!(classify_ref_mode("HEAD~3", Some("HEAD")), "single_commit");
         assert_eq!(classify_ref_mode(&"a".repeat(40), Some("HEAD")), "sha");
+    }
+
+    // --- classify_truncation_reason ---
+    //
+    // One test per arm so mutation testing can point at exactly one killed
+    // mutant per test. The final "unknown" test covers both the fallthrough
+    // arm and the safety-net behavior for unrecognized labels.
+
+    #[test]
+    fn it_classifies_paginated_as_paginated() {
+        assert_eq!(classify_truncation_reason("paginated"), "paginated");
+    }
+
+    #[test]
+    fn it_classifies_max_files_as_max_files() {
+        assert_eq!(classify_truncation_reason("max_files"), "max_files");
+    }
+
+    #[test]
+    fn it_classifies_max_file_size_as_max_file_size() {
+        assert_eq!(classify_truncation_reason("max_file_size"), "max_file_size");
+    }
+
+    #[test]
+    fn it_classifies_token_budget_as_token_budget() {
+        assert_eq!(classify_truncation_reason("token_budget"), "token_budget");
+    }
+
+    #[test]
+    fn it_classifies_unknown_reason_as_unknown() {
+        assert_eq!(classify_truncation_reason("TOKEN_BUDGET"), "unknown");
+        assert_eq!(classify_truncation_reason("something_else"), "unknown");
+        assert_eq!(classify_truncation_reason(""), "unknown");
     }
 }
--- a/src/tools/manifest.rs
+++ b/src/tools/manifest.rs
@@ -8,6 +8,7 @@ use crate::git::diff::ChangeType;
 use crate::git::generated::GeneratedFileDetector;
 use crate::git::reader::RepoReader;
 use crate::pagination::{PaginationCursor, PaginationInfo, encode_cursor};
+use crate::tools::size;
 use crate::tools::types::{
     FunctionChange, FunctionChangeType, ImportChange, ManifestFileEntry, ManifestMetadata,
     ManifestOptions, ManifestResponse, ManifestSummary, ToolError, detect_language,
@@ -368,7 +369,7 @@ pub fn build_manifest(
         languages_affected: all_languages_affected,
     };
 
-    Ok(ManifestResponse {
+    let mut response = ManifestResponse {
         metadata: ManifestMetadata {
             repo_path: repo_path.display().to_string(),
             base_ref: base_ref.to_string(),
@@ -377,6 +378,10 @@ pub fn build_manifest(
             head_sha: head_commit.sha,
             generated_at: Utc::now(),
             version: env!("CARGO_PKG_VERSION").to_string(),
+            // Placeholder; overwritten below via a two-pass estimate so the
+            // final value reflects the fully-populated response. See the
+            // ManifestMetadata::token_estimate doc comment for the caveat.
+            token_estimate: 0,
         },
         summary,
         files: manifest_files,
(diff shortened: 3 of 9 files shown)


src/tools/size.rs
```diff
@@ -0,0 +1,115 @@
...
+/// count, using the standard "~4 characters per token" heuristic.
+///
+/// `char_count` is the length of the UTF-8 JSON payload in bytes or characters —
+/// for ASCII-heavy JSON the two are equivalent, and the helper deliberately
+/// accepts whichever the caller has cheapest to compute. Uses integer division
+/// (floor), which matches the existing inline implementation in
+/// [`crate::tools::snapshots`] so the refactor there is a behavior-preserving
+/// swap.
+#[must_use]
+pub fn estimate_tokens(char_count: usize) -> usize {
```
mikelane, line 22: Doc nit: this comment correctly notes that ASCII-heavy JSON makes bytes vs character count equivalent, but the parameter is still named `char_count`. Both call sites (`snapshots.rs` total_chars accumulator, and `estimate_response_tokens` via `String::len()`) actually pass byte counts. Would renaming to `serialized_len` (or even `byte_len`) eliminate the half-truth without losing the doc explanation? The doc could then describe the heuristic and drop the "or characters" hedge.

Not a blocker — the helper is internal and the behavior is right. Flag for the next time anyone touches this signature.

src/tools/size.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// Returns `0` if serialization fails. For types that derive `Serialize`
+/// cleanly this is effectively unreachable, but we prefer a degraded budgeting
+/// hint over a panic at the metric boundary. Callers that care about real
+/// serialization errors should serialize themselves — this helper is meant
+/// to be called right before returning a response struct that will be
+/// serialized anyway, where a failure here would also break the actual
+/// response.
+#[must_use]
+pub fn estimate_response_tokens<T: Serialize>(value: &T) -> usize {
```
mikelane, line 37: Naming nit: `estimate_response_tokens` is generic over `T: Serialize`, so the "response" in the name is aspirational — the type system would happily accept a `HashMap`, a `Vec<u8>`, or any other serializable value. `estimate_serialized_tokens` would more accurately describe what the function actually does, and the doc comment could keep the "intended for response structs right before return" guidance.

Defer-able. Worth flagging in case PR 3 imports this from a non-response context (e.g., for a sub-tree budget) and the name starts looking actively misleading.

src/privacy.rs
```diff
@@ -94,6 +94,24 @@ pub fn classify_ref_mode(base_ref: &str, head_ref: Option<&str>) -> &'static str
...
+/// Unlike [`classify_error_kind`], which does substring matching on error
+/// messages coming from third-party libraries, this function uses a flat
+/// exact match because the caller is always the codebase itself passing a
+/// known-good literal. An exhaustive match means mutation testing can kill
+/// every arm cleanly, and `unknown` exists as a safety net for future
+/// refactors that might introduce a typo or a forgotten label.
+pub fn classify_truncation_reason(reason: &str) -> &'static str {
+    match reason {
+        "paginated" => "paginated",
+        "max_files" => "max_files",
```
mikelane, line 108: The `max_files` arm has no production emitter anywhere in the repo today — it is reserved for a later PR in the issue #212 stack. That is fine and the doc comment above explains the design correctly.

One small follow-up worth tracking: when PR 3 (manifest budget) lands, please make sure something actually calls `record_truncated("get_change_manifest", "max_files")` from the budget path, otherwise this arm becomes dead code that mutation testing might silently cover via the `unknown` fallback. A grep test `scripts/check-truncation-arms-have-emitters.sh` would be a heavyweight solution; a TODO comment referencing #212 PR 3 here would be a lighter one.

src/tools/types.rs
```diff
@@ -66,6 +66,16 @@ pub struct ManifestMetadata {
...
+    /// Estimated token count of the serialized response, via the ~4-chars-
+    /// per-token heuristic in [`crate::tools::size::estimate_response_tokens`].
+    /// Agents use this as a cheap pre-flight hint before requesting a follow-
+    /// up call (e.g., `get_function_context` on the same range). Populated in
+    /// a two-pass build: the response is first constructed with `0`, then the
+    /// estimate is computed on that struct and written back. The final value
+    /// is therefore a lower bound that undercounts by the single-digit
+    /// character delta between `"token_estimate":0` and the real value, which
+    /// is acceptable for a budgeting hint.
+    pub token_estimate: usize,
```
mikelane, line 78: The doc comment is excellent — explicitly calling out the lower-bound caveat from the two-pass estimate is exactly the kind of thing future maintainers will thank you for. One follow-up worth raising as a comment for PR 3's reviewer (not this one):

`ManifestMetadata.repo_path` is a raw absolute filesystem path that is already on the wire (pre-existing, not introduced here), but with this PR it is now materialized into an intermediate `serde_json::to_string` buffer once for `estimate_response_tokens` and again by rmcp for transport. Heap strings are not zeroized on drop, so an attacker with process-memory read has a slightly larger window than before. Pre-existing field, marginal new exposure.

Worth thinking about for PR 3: could the budget projection run over the `summary` + `files` + `pagination` subtree instead of the whole response? That would also let `token_estimate` itself drop `metadata` from the calculation, which solves the "placeholder skews the estimate by ≤1 token" problem at the same time. Not a PR 2 ask — just a design heads-up.

src/tools/manifest.rs
```diff
@@ -387,7 +392,9 @@ pub fn build_manifest(
             page_size,
             next_cursor,
         },
-    })
+    };
+    response.metadata.token_estimate = size::estimate_response_tokens(&response);
```
mikelane, line 396: The two-pass placeholder pattern is the right call here — the alternative (compute on a partial struct then mutate after construction) reads worse, and CLAUDE.md is explicit that three similar lines beats a premature abstraction. The placeholder comment makes the intent obvious to the next reader.

One observability note for follow-up (not a PR 2 blocker): `server.rs` currently records the `git_prism.response.tokens_estimated` metric by re-serializing the response after `build_manifest` returns. That second serialization happens AFTER `token_estimate` has been written back, so the value the metric sees disagrees with `metadata.token_estimate` by ≤1 token (the literal width difference between `"token_estimate":0` and `"token_estimate":N`). When PR 3 starts enforcing budgets it will probably want to read the field directly from `response.metadata.token_estimate` rather than recomputing — issue #212 Design Decision (B) blocks that change here, but it is worth a follow-up note.

src/metrics.rs
```diff
@@ -221,11 +221,16 @@ impl Metrics {
...
+        // caller passes. The classifier is a flat exact-match — see
+        // `crate::privacy::classify_truncation_reason` for the full rationale.
+        let normalized = crate::privacy::classify_truncation_reason(reason);
         self.response_truncated.add(
             1,
             &[
                 KeyValue::new("tool", tool.to_string()),
-                KeyValue::new("reason", reason.to_string()),
+                KeyValue::new("reason", normalized),
             ],
```
mikelane, line 234: Nice — normalizing at the metric boundary instead of at every call site is exactly the right shape. The flat-match classifier means future call sites cannot blow up cardinality even if someone forgets the design rule, and the doc comment on `classify_truncation_reason` makes the intent obvious.

Worth thinking about for PR 3: the `response_truncated` counter is currently `(tool, reason)`-only. As soon as PR 3 has multiple reasons firing on the same tool, the existing `response_truncated` span attribute (a bool elsewhere in the codebase) will need to become a string. Pre-existing trajectory, just a heads-up.

src/tools/snapshots.rs
```diff
@@ -45,7 +46,7 @@ pub fn build_snapshots(
             generated_at: Utc::now(),
         },
         files,
-        token_estimate: total_chars / 4,
+        token_estimate: size::estimate_tokens(total_chars),
```
mikelane, line 49: Clean refactor — `size::estimate_tokens(total_chars)` reads better than the inline `/4` and the snapshot test for the new helper guarantees the behavior is preserved (rounds-down semantics on the same input). No notes.

mikelane (commented):
## Verdict: APPROVE (comment-only, since solo-author cannot self-approve)

PR 2 of the issue #212 stack is well-scoped, well-tested, and respects the design decisions from the parent issue. Every file touched maps to the PR 2 row in the stack plan; no drive-by additions, no scope leakage, no premature abstraction of the two-pass pattern (which CLAUDE.md explicitly says is the right call at three call sites).

### What is excellent

- **`classify_truncation_reason` lives in `src/privacy.rs` next to `classify_error_kind`.** Mirrors the existing pattern, normalizes at the metric boundary, and the flat-match shape means mutation testing kills every arm cleanly.
- **`ManifestMetadata.token_estimate` doc comment is honest about the lower-bound caveat.** Future maintainers will not be surprised by the ≤1 token undercount from the placeholder serialization.
- **The metrics test on `record_truncated`** is candid that it cannot read back attribute values from the no-op meter and uses the classifier directly to lock in the wiring contract. That kind of self-aware test commentary is rare and welcome.
- **`snapshots.rs` refactor is a clean behavior-preserving swap** — the new helper covers the rounds-down semantics with parametrized cases.

### Inline comments posted (all defer-able)

1. **`size.rs:22`** — `estimate_tokens(char_count)` parameter name vs. the byte-count reality at the call sites. Defer-able rename.
2. **`size.rs:37`** — `estimate_response_tokens` is generic over `T: Serialize`; consider `estimate_serialized_tokens` so PR 3 sub-tree budgeting reads naturally.
3. **`privacy.rs:108`** — `max_files` arm has no production emitter today; flag for PR 3 to actually wire it so the arm does not silently rot under the `unknown` fallback.
4. **`types.rs:69-78`** — Doc comment praise plus a heads-up about `repo_path` exposure widening through the new transient serialization, and a suggestion to project the budget over `summary`+`files`+`pagination` only in PR 3.
5. **`manifest.rs:372-396`** — Two-pass pattern is correct here; flag the `server.rs` recompute drift for PR 3 to fix when it starts reading `response.metadata.token_estimate` directly.
6. **`metrics.rs:223-234`** — Normalization shape is great; heads-up that the `response_truncated` span attribute will need to become a string when PR 3 has multiple reasons.
7. **`snapshots.rs:49`** — Praise.

### Not commented but worth tracking out-of-band

- `manifest.rs` is 3410 lines (6.8x the 500-line soft limit). Pre-existing, this PR adds 40 lines, not a blocker. Worth a dedicated split issue.
- `context.rs` just crossed 500 lines this PR (548). Pre-existing trajectory, monitor.
- `types.rs` now 779 lines. God-module trajectory, monitor.
- `SIZE_BUCKETS` shared between `response_bytes` and `response_tokens_estimated` histograms — tail buckets on the tokens histogram are mostly dead. Pre-existing, defer.

### Decision

**APPROVED** (posting as `COMMENT` event because GitHub blocks self-approval on owner-authored PRs). All blockers are pre-existing or out of PR 2 scope; nothing in the diff is wrong.

PR #220, opened 2026-04-16
feat(manifest)!: token budget and breaking include_function_analysis flip

mikelane:
## Summary

PR 3 of issue #212 — fixes unbounded response sizes from \`get_change_manifest\` via two breaking changes and a tiered truncation algorithm.

**Breaking changes:**
- \`include_function_analysis\` default flipped from \`true\` to \`false\` (opt-in).
- New \`max_response_tokens\` parameter (default \`8192\`) bounds the response via progressive per-file truncation. Trimmed files list in \`metadata.function_analysis_truncated\`. Pass \`0\` to disable.
- \`record_truncated\` metric now carries a bounded \`reason\` label (new \`token_budget\` event emitted on budget hits).

**Algorithm:** three-tier per-file downgrade, greedy tier1-first.
1. Phase 1 — all files fit at full: no trim.
2. Phase 2 — all files fit at tier 1 (functions kept, imports stripped): apply uniformly.
3. Phase 3 — greedy tier1-first: early files get tier 1, later files get tier 2 (bare), files beyond the budget are dropped. Files are returned in page order (git diff order) rather than skipping smaller later files — callers needing dropped files should re-request with a larger budget or \`max_response_tokens: 0\`.

A 5% safety margin (minimum 16 tokens) is reserved for the \`function_analysis_truncated\` list overhead that gets populated after enforcement.

**CLI combined mode:** \`collect_all_manifest_pages\` disables per-page enforcement during collection and applies \`enforce_token_budget\` once on the combined result. When budget enforcement drops files, \`pagination.page_size\` is updated to reflect the actual returned count — callers can detect file dropping via \`response.files.len() < pagination.total_items\` and re-request without enforcement (\`--max-response-tokens 0\`) to get all files.

**Internal caller:** \`src/tools/context.rs\` passes \`max_response_tokens: None\` so \`get_function_context\` (PR 4) gets the full unbudgeted manifest.

## Test plan

- [x] 635 unit tests pass (includes 5 new for \`enforce_token_budget\`)
- [x] 97 BDD scenarios pass (includes 5 new for PR 3 of issue #212)
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean
- [x] pre-push hooks pass
- [x] pr-reviewer gauntlet: all findings addressed (2 blockers, 3 warnings fixed in follow-up commit 7e34d0e)

## BDD scenarios

The 5 PR 3 scenarios from \`bdd/features/response_size_guardrails.feature\`:
- Default manifest call omits function analysis
- Opt-in manifest call includes function analysis
- Over-budget manifest trims function detail to signatures only (2048 budget)
- Over-budget manifest emits the token_budget truncation metric
- Change manifest stays within the 8192 token budget on an extreme change (200 files / 1000 functions)

PR 4 will own the remaining \`get_function_context\` scenarios (pagination, name filter, caller/callee truncation).

## Documentation

- \`CHANGELOG.md\` — \`[Unreleased]\` section added with the three breaking changes.
- \`CLAUDE.md\` — Tool Discipline section updated for opt-in function analysis and the budget.
- \`src/server.rs\` — \`#[tool]\` doc comment for \`get_change_manifest\` explicitly documents the flipped default, the budget, and \`function_analysis_truncated\`.

Refs #212

### DIFF
--- a/bdd/steps/size_assertion_steps.py
+++ b/bdd/steps/size_assertion_steps.py
@@ -419,9 +419,22 @@ def step_telemetry_metric_truncation(
     context: Context, reason: str, tool: str,
 ) -> None:
     """Verify that budget enforcement ran by checking response metadata.
-    Full OTLP metric verification requires a collector harness (out of scope
-    for CLI-level BDD). We proxy the metric assertion by confirming the
-    response carries evidence that the budget was hit."""
+
+    Full OTLP metric verification requires a live collector harness, which is
+    out of scope for CLI-level BDD tests. We proxy the metric assertion via
+    ``metadata.function_analysis_truncated``:
+
+        non-empty truncated list  ->  server.rs emits record_truncated(tool, "token_budget")
+
+    The invariant is enforced in server.rs:
+        if !response.metadata.function_analysis_truncated.is_empty() {
+            metrics.record_truncated(tool_name, "token_budget");
+        }
+
+    If that guard ever breaks, this proxy will still pass even though the real
+    metric was not recorded. A future telemetry integration test should verify
+    the actual counter.
+    """
     data = _ensure_json_parsed(context)
     truncated = _navigate_dotted_path(data, "metadata.function_analysis_truncated")
     # Budget enforcement emits the metric when this list is non-empty
--- a/bdd/steps/size_assertion_steps.py
+++ b/bdd/steps/size_assertion_steps.py
@@ -222,11 +222,15 @@ def step_response_omits_function_diffs(context: Context) -> None:
 def step_response_includes_function_diffs(context: Context) -> None:
     data = _ensure_json_parsed(context)
     files = _navigate_dotted_path(data, "files")
-    with_diffs = sum(1 for entry in files if entry.get("functions_changed"))
-    total = len(files)
+    # Only check files with a supported language — tree-sitter returns null
+    # for functions_changed when no grammar exists (e.g. language == "unknown").
+    supported = [f for f in files if f.get("language", "unknown") != "unknown"]
+    with_diffs = sum(1 for entry in supported if entry.get("functions_changed"))
+    total = len(supported)
+    assert total > 0, "expected at least one file with a supported language"
     assert with_diffs == total, (
-        f"expected every changed file to carry function detail when opted in; "
-        f"got {with_diffs} of {total}. RED until PR 3 lands the opt-in handler."
+        f"expected every supported-language file to carry function detail when opted in; "
+        f"got {with_diffs} of {total}."
     )
 
 
@@ -414,18 +418,14 @@ def step_response_truncated_entries_shortened(context: Context) -> None:
 def step_telemetry_metric_truncation(
     context: Context, reason: str, tool: str,
 ) -> None:
-    """Placeholder for the OTLP truncation metric assertion.
-
-    PR 1 intentionally does not wire a telemetry collector fixture for the
-    `git_prism.response.truncated` counter -- that plumbing arrives with the
-    implementation PRs (#3 for manifest, #4 for function context). Forcing a
-    deterministic RED here keeps the scenario @not_implemented until those
-    PRs remove the tag. See the stack plan in issue #212 for details.
-    """
-    assert False, (
-        f"telemetry assertion path not wired in PR 1 scaffold -- "
-        f"expected git_prism.response.truncated event with "
-        f"reason={reason!r} for tool={tool!r}. "
-        f"Implementation PR 3 (manifest) / PR 4 (context) will wire the "
-        f"real assertion via bdd/steps/telemetry_steps.py."
+    """Verify that budget enforcement ran by checking response metadata.
+    Full OTLP metric verification requires a collector harness (out of scope
+    for CLI-level BDD). We proxy the metric assertion by confirming the
+    response carries evidence that the budget was hit."""
+    data = _ensure_json_parsed(context)
+    truncated = _navigate_dotted_path(data, "metadata.function_analysis_truncated")
+    # Budget enforcement emits the metric when this list is non-empty
+    assert isinstance(truncated, list) and truncated, (
+        f"Expected non-empty function_analysis_truncated as proxy for "
+        f"{reason} metric on {tool}, got {truncated!r}"
     )
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,6 +42,12 @@ enum Commands {
         /// Page size for internal pagination (default 500)
         #[arg(long, default_value_t = 500)]
         page_size: usize,
+        /// Include function-level analysis (off by default)
+        #[arg(long)]
+        include_function_analysis: bool,
+        /// Maximum estimated tokens for the response (default 8192, 0 to disable)
+        #[arg(long, default_value_t = 8192)]
+        max_response_tokens: usize,
     },
     /// Output file snapshots as JSON (CLI mode, no MCP)
     Snapshot {
@@ -234,6 +240,8 @@ async fn main() -> anyhow::Result<()> {
             range,
             repo,
             page_size,
+            include_function_analysis,
+            max_response_tokens,
         } => {
             let repo_path = repo.map(PathBuf::from).unwrap_or_else(|| {
                 std::env::current_dir().expect("cannot determine current directory")
@@ -241,7 +249,12 @@ async fn main() -> anyhow::Result<()> {
             let options = ManifestOptions {
                 include_patterns: vec![],
                 exclude_patterns: vec![],
-                include_function_analysis: true,
+                include_function_analysis,
+                max_response_tokens: if max_response_tokens == 0 {
+                    None
+                } else {
+                    Some(max_response_tokens)
+                },
             };
             let manifest = match parse_range(&range) {
                 RefRange::CommitRange { base, head } => {
@@ -271,6 +284,7 @@ async fn main() -> anyhow::Result<()> {
                 include_patterns: vec![],
                 exclude_patterns: vec![],
                 include_function_analysis: true,
+                max_response_tokens: None,
             };
             let history =
                 collect_all_history_pages(&repo_path, base_ref, head_ref, &options, page_size)?;
@@ -552,6 +566,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         let result = collect_all_manifest_pages(&path, "HEAD~1", "HEAD", &options, 500).unwrap();
@@ -568,6 +583,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         // Use page_size=2, so we need 3 pages for 5 files
@@ -585,6 +601,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         let result = collect_all_manifest_pages(&path, "HEAD~1", "HEAD", &options, 2).unwrap();
@@ -602,6 +619,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         // page_size=1 forces 3 pages
@@ -618,6 +636,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         let result = collect_all_history_pages(&path, "HEAD~3", "HEAD", &options, 500).unwrap();
@@ -634,6 +653,7 @@ mod tests {
             include_patterns: vec![],
             exclude_patterns: vec![],
             include_function_analysis: false,
+            max_response_tokens: None,
         };
 
         // page_size=2 forces 3 pages for 5 commits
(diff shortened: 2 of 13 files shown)


src/tools/manifest.rs
```diff
@@ -3407,4 +3641,222 @@ mod tests {
...
+    fn enforce_token_budget_zero_budget_returns_empty() {
+        // budget=0 is handled by the caller (not calling enforce_token_budget),
+        // but if it were called, the function should still not crash
+        let files = vec![make_test_file_entry("a.rs", true, true)];
+        let mut response = make_test_response(files);
+        let trimmed = enforce_token_budget(&mut response, 0);
+        // With budget 0, everything gets stripped but the function still returns
+        assert!(
+            trimmed.is_empty() || !trimmed.is_empty(),
+            "should not panic"
```
mikelane: **BLOCKER — [HIGH] Tautological assertion (`x || !x` is always true)**

The assertion on this line reduces to `true` by the law of excluded middle — it will pass regardless of what `enforce_token_budget` returns, even if the function begins returning garbage. The accompanying comment says "should not panic", but `assert!` does not catch panics either way; panic handling is the test harness's responsibility, so the assertion adds no coverage for that property.

Concretely, this test's body would pass if:
- `trimmed` is empty (current behavior)
- `trimmed` contains every file
- A future refactor makes `enforce_token_budget` return a `Vec` populated from an unrelated bug

The test name promises "returns empty", but the assertion accepts any return value. Three options, in order of preference:

1. If the intended contract is "budget=0 is caller-filtered; calling it directly is undefined" — delete the test. The invariant lives in the call site (`.filter(|&b| b > 0)`), not in this function.
2. If the intended contract is "budget=0 strips everything and returns an empty vec" — assert that explicitly (`assert!(trimmed.is_empty())` plus `assert!(response.files.iter().all(|f| f.functions_changed.is_none() && f.imports_changed.is_none()))`).
3. If the intended contract is just "does not panic" — drop the assertion entirely; the test harness already catches panics. A no-assert test that calls the function is a valid "this doesn't crash" smoke test.

As written, the test inflates coverage without testing behavior.

src/main.rs
```diff
@@ -149,7 +171,18 @@ fn collect_all_manifest_pages(
             page_size: total_items,
             next_cursor: None,
         },
-    })
+    };
+
+    // Apply budget enforcement on the combined response
+    if let Some(budget) = options.max_response_tokens.filter(|&b| b > 0)
```
mikelane, line 177: **BLOCKER — [HIGH] CLI combined mode can silently drop files beyond the budget**

When budget enforcement reaches Phase 3, `enforce_token_budget` calls `response.files.truncate(files_to_keep)` and returns only the paths of *tier-1 trimmed* files. In combined CLI mode, the resulting `ManifestResponse` then has:

- `pagination.next_cursor = None` (combined mode always clears this just before return)
- `pagination.total_items = total_items` (set to `all_files.len()` before truncation, so it may or may not reflect the dropped files — worth double-checking against `response.files.len()` post-truncation)
- `metadata.function_analysis_truncated` lists only the tier-1 files; entirely dropped files are silent

So a user running `git-prism manifest main..HEAD` on a PR with 300 changed files may receive 150 files with no visible indication that the rest were omitted. The `token_estimate` hints that trimming occurred, but there's no first-class field that says "N files were dropped entirely." For a CLI that prints JSON to stdout (no paging, no follow-up call), that silence is the kind of data loss that becomes a production bug when an agent relies on completeness.

Three options to consider, in roughly decreasing order of impact:

1. **Record dropped paths.** Add `files_omitted: Vec<String>` to `ManifestMetadata` (or extend `function_analysis_truncated` with a variant tag), populated by `enforce_token_budget` whenever `files.truncate()` discards entries. Budget-enforced CLI output then always advertises what was dropped.
2. **Preserve the cursor in CLI combined mode.** If Phase 3 truncated, keep `next_cursor` set (rather than clearing it in the "combined" response), so that downstream tooling can detect the incomplete result and decide how to proceed.
3. **Restrict budget enforcement to MCP mode.** The CLI has a user at the keyboard who can increase `--max-response-tokens` explicitly. Budget enforcement guards the MCP transport, not the CLI. Passing `max_response_tokens: None` in the CLI's `ManifestOptions` (when the flag defaults aren't overridden by `--max-response-tokens 0`) side-steps the issue entirely.

Option 3 is the smallest surface change; option 1 is the most diagnostic. Either addresses the silent-drop concern. Worth deciding which semantics the CLI is meant to provide before merging.

src/tools/manifest.rs
```diff
@@ -620,10 +647,172 @@ pub fn build_worktree_manifest(
...
+
+    for c in &costs {
+        if c.has_analysis && c.tier1 <= remaining {
+            remaining -= c.tier1;
+            decisions.push(TierChoice::Tier1);
+        } else if c.bare <= remaining {
+            remaining -= c.bare;
+            decisions.push(TierChoice::Bare);
+        } else {
+            // File doesn't fit even bare — stop adding files
```
mikelane: **WARNING — [MEDIUM] Greedy algorithm bails at first non-fitting file and may leave budget on the table**

If file N doesn't fit even as a bare entry, the `break` exits the loop and files N+1..M are all dropped — regardless of whether any of them would have fit at tier 1 or bare. In the wild, the order dependence matters: a single large file appearing early (say, a long `Cargo.lock` diff) can evict a long tail of small files that would each have fit fine.

Two concerns:

1. **Wasted budget.** The remaining space after the oversized file is left unused — the algorithm forfeits bytes it could have spent.
2. **Ordering sensitivity.** Output depends on the input order of `response.files`. `build_manifest` produces files in `git` diff order, which is stable but arbitrary from the agent's perspective. A refactor that sorts files differently (e.g. by change_type or language) would produce silently different truncation. If ordered output is an intentional design constraint ("the agent should see files in diff order, and truncation is at the tail"), that's a defensible position — but it should be documented on `enforce_token_budget`'s doc comment so the contract is explicit.

Two alternatives to consider:

- `continue` instead of `break`: skip non-fitting files and keep scanning. Preserves "best effort" semantics at the cost of output order being harder to reason about — an agent seeing file 7 but not file 5 has to check the paths to understand what happened.
- Sort files by bare cost ascending before the greedy pass, so small files always get a chance. Probably over-complicates the algorithm for marginal benefit.

Given the tradeoffs, documenting the current behavior ("greedy first-fit in diff order; truncation at tail") on the function's doc comment is probably the cheapest fix. Either that or the `continue` variant.

bdd/steps/size_assertion_steps.py
```diff
@@ -414,18 +418,14 @@ def step_response_truncated_entries_shortened(context: Context) -> None:
 def step_telemetry_metric_truncation(
```
mikelane, line 418: **WARNING — [MEDIUM] Metric assertion is a proxy, not the real thing**

This step checks `metadata.function_analysis_truncated` as a stand-in for the `git_prism.response.truncated{reason="token_budget"}` counter, leaning on the invariant that server.rs emits the metric whenever the list is non-empty:

```rust
if !response.metadata.function_analysis_truncated.is_empty() {
    metrics.record_truncated(tool_name, "token_budget");
}
```

The proxy holds today, but the test has no way to detect if that `if` block is ever accidentally deleted, refactored into a `match` with a missed arm, or gated on a new config flag. The PR description acknowledges full OTLP verification is out of scope — that's a reasonable trade-off for this PR. Worth flagging two things in a follow-up issue so the gap is tracked:

1. **Document the proxy in the docstring.** The current comment says "Full OTLP metric verification requires a collector harness (out of scope for CLI-level BDD)." Extend it to name the server.rs invariant being proxied ("this assertion relies on `server.rs` emitting `record_truncated(tool, "token_budget")` whenever `function_analysis_truncated` is non-empty — if that invariant breaks, this proxy goes stale silently"). That gives a future maintainer a breadcrumb back to the real check when they touch either file.
2. **Track the collector harness as deferred work.** The `telemetry_steps.py` module already wires some OTLP assertions; extending it to cover this counter closes the loop. Not a blocker for PR 3, but worth an issue so it doesn't fall off the map.

No change required to merge — this is a "call out the known gap" note, not a blocker.

src/tools/manifest.rs
```diff
@@ -620,10 +647,172 @@ pub fn build_worktree_manifest(
...
+    let files = std::mem::take(&mut response.files);
+    let skeleton_cost = size::estimate_response_tokens(response);
+    response.files = files;
+
+    // Safety margin for the `function_analysis_truncated` list that gets
+    // populated AFTER enforcement runs. Each trimmed path serializes as ~25
+    // chars (path + JSON quoting + comma), and large changes can produce
+    // dozens of trimmed paths, adding hundreds of tokens we didn't budget for.
+    // Reserve ~5% of the budget (min 256 tokens) as headroom.
+    let safety_margin = budget / 20;
```
mikelane: **LOW — Safety margin degenerates to 0 for tiny budgets**

`budget / 20` uses integer division, so any `budget < 20` yields `safety_margin = 0` with no minimum guard. For realistic budgets (256, 512, 2048, 8192 — all paths I see in this PR and the BDD suite), this is fine. The edge only bites if someone (a) writes a unit test with a tiny budget and (b) depends on the safety margin's protection for list overhead.

Cheap defensive fixes:

- `let safety_margin = (budget / 20).max(16);` — anchors a minimum floor of ~16 tokens regardless of budget.
- Or just a comment noting the degenerate behavior: `// For budget < 20, safety_margin collapses to 0; callers filter budget > 0 so this is only reachable with intentional tiny budgets in tests.`

Either is fine. Not merge-blocking — flagging so it's an informed choice.

mikelane (commented):
## Review Verdict: CHANGES REQUESTED

(Posted as `COMMENT` event since GitHub blocks self-review verdicts on owner-authored PRs; treat as a "Request Changes" for the gauntlet.)

### Summary

The token-budget implementation is solid in its core shape — the three-tier downgrade strategy is clean, the `ManifestOptions.max_response_tokens: Option<usize>` design cleanly separates "use the default" (Some) from "internal caller, bypass enforcement" (None), and the CLI fix that enforces budget *after* page collection (rather than per-page) is exactly right. The 5 BDD scenarios exercise the real contract, and the server.rs doc comment gives MCP clients a fair warning about the flipped default. Good separation between data and presentation throughout.

Two blockers to address before merge; three warnings to weigh.

### Blockers (must fix or explicitly justify)

1. **[BUG-HUNTER HIGH]** Tautological assertion in `enforce_token_budget_zero_budget_returns_empty` (`src/tools/manifest.rs:3824`). `x || !x` is always true. The test exercises a code path and inflates coverage without testing behavior. See inline comment for three alternatives.

2. **[BUG-HUNTER HIGH]** Silent file dropping in CLI combined mode (`src/main.rs:177` and `:235`). When `enforce_token_budget` truncates the file list, the combined response clears `next_cursor`, so entirely-dropped files vanish with no indication. Three mitigation options in the inline comment; option 3 (scope budget to MCP, not CLI combined mode) is the smallest surface change.

### Warnings (address or acknowledge)

3. **[BUG-HUNTER MEDIUM]** Greedy algorithm's `break` at first non-fitting file wastes remaining budget and makes output order-sensitive (`src/tools/manifest.rs:784`). Cheapest fix: document the contract in the doc comment. Alternative: `continue` instead of `break`.

4. **[TEST MEDIUM]** `step_telemetry_metric_truncation` proxies the real metric via `function_analysis_truncated` (`bdd/steps/size_assertion_steps.py:418`). Acceptable tradeoff given scope; worth a follow-up issue and a docstring note about the invariant being proxied.

5. **[ARCH LOW]** Safety margin integer-divides to 0 for budgets under 20 (`src/tools/manifest.rs:705`). Harmless in practice; noting so the choice is informed.

### Adversarial notes (for the author's awareness)

- **Breaking change blast radius.** `ManifestOptions` is public — downstream Rust code that constructs it now fails to compile without `max_response_tokens`. Also, every MCP agent that previously read `response.functions_changed[0]` without a null check now dereferences `null` on any call that didn't explicitly opt in. These are in-scope for the PR but not called out in the description; worth a mention in the CHANGELOG's breaking-changes section or a migration note.
- **Revert safety.** The new serde fields gate on `skip_serializing_if = "Vec::is_empty"`, so a revert doesn't leave dangling keys in clients' expectations. Good shape.
- **Gaming.** Tests 2 and 3 depend on `estimate_response_tokens` being accurate; a bug in the estimator could make those tests pass vacuously. The BDD scenarios mitigate by asserting on concrete budgets (2048, 8192). OK.

### Praise

- The `ManifestOptions.max_response_tokens: Option<usize>` type choice correctly separates "caller wants the default applied" from "internal caller wants to opt out." That's cleaner than a `0 = disabled` sentinel, which the MCP/CLI layer still uses (defensibly, since `usize` can't be `Option` across the JSON wire).
- The fix in `collect_all_manifest_pages` (disable per-page enforcement, apply once on the combined response) is precisely the right shape — the bug where per-page enforcement was defeated by page aggregation is subtle, and the fix is surgical.
- `function_analysis_truncated` distinguishes tier-1 (signatures preserved) from tier-2 (bare) via presence/absence in the list — that's a nice, minimal encoding of the truncation semantics.
- Internal callers (`context.rs`) correctly pass `max_response_tokens: None` to bypass enforcement, so the function-context tool gets unbudgeted data. That boundary is exactly where it should be.
- The `_inject_legacy_manifest_defaults` shim in `cli_steps.py` is a pragmatic way to avoid rewriting every pre-#212 BDD scenario — scoped tightly to `git-prism manifest` invocations without the flag, and clearly commented as to why it exists.

### Decision

**CHANGES REQUESTED.** The two BLOCKERs are mechanical fixes — the tautological assertion is a 1-line replacement, and the silent-drop issue is either a doc comment + field addition or a mode restriction. Once those land, the PR is a clean merge.

Issue #234, opened 2026-04-26
Epic: Bulletproof git-prism redirect — sharper tool descriptions, bundled hooks, review_change orchestration


mikelane:
## Goal

Make git-prism the consistently-chosen tool for change inspection by tightening MCP tool descriptions, shipping a redirect-hook installer, and adding a single-call orchestration tool. Replace the current ad-hoc local hook (`~/.claude/hooks/git-prism-redirect.sh`) with a tested, bundled artifact that parses bash properly via a stdlib-only Python tokenizer.

## Context

Agents reach for raw `git diff/log/show/blame` instead of git-prism MCP tools because:

1. **Pretraining bias is enormous.** Every codebase the model saw used porcelain git. Tool selection happens at the description layer, not the prose-instructions layer, so CLAUDE.md and skills get out-leveraged by a competing tool with a 10× prior.
2. **The local regex hooks are brittle.** They are shell scripts that pattern-match against `tool_input.command` with regex. They miss subshells (`(git log a..b)`), variable substitution (`git diff $BASE..HEAD`), compound commands (`cd path && git diff main..HEAD`), heredoc'd commands, and command-form variants the patterns weren't written for. The author just hit one such miss (`git diff --oneline sha..HEAD` not redirecting) and is rightly worried more cases lurk.
3. **Tool descriptions are uneven.** `get_change_manifest` and `get_file_snapshots` have strong comparative doc comments. `get_commit_history` has only a redundant 2-line doc plus a bare `description = "Returns per-commit manifests for a range of commits"`. **`get_function_context` has no `///` doc comment at all** (`src/server.rs` near line 524). Every tool-selection decision reads these strings fresh — gaps here cost more than any other layer.
4. **The redirect hooks live in user dotfiles, not in git-prism.** Anyone else installing git-prism gets the binary but none of the redirect scaffolding. There is no shipping path for the leverage we have built.

The intended outcome is **defense in depth**: agents pick git-prism because the descriptions make it obvious, fall through to a thorough hook backstop when they don't, and install all of it with one command.

### Tool description rewrites
- Rewrite all four MCP tool doc comments (`src/server.rs`) for comparative framing — explicit "use this instead of `git X`" language, quantified token savings.
- Fix the missing `get_function_context` `///` doc.
- Replace `get_commit_history`'s bare `description = "..."` argument with a full `///` block.

### Bundled hook installer
- Add `git-prism install-hooks --scope <user|project|local> [--uninstall]` subcommand.
[...]

Issue #239, opened 2026-04-26
feat(cli): install-hooks subcommand and bundled redirect script


mikelane:
## Goal

Add `git-prism install-hooks` subcommand and ship the canonical `hooks/git-prism-redirect.sh` script alongside it. Together these let any user installing git-prism opt into the redirect-hook stack with one command.

Parent epic: #234. Depends on W3 (Python tokenizer) — the hook script invokes `parse_git_invocations.py` at runtime.

## CLI Surface

```
git-prism install-hooks --scope <user|project|local> [--uninstall] [--dry-run]
```

| Flag | Semantics |
|------|-----------|
| `--scope user` | Writes `~/.claude/settings.json` |
| `--scope project` | Writes `<repo>/.claude/settings.json` (creates the directory if missing) |
| `--scope local` | Writes `<repo>/.claude/settings.local.json` |
| `--uninstall` | Removes only the entries this command added |
| `--dry-run` | Prints the diff that would be applied without writing |

Mirrors `claude mcp add --scope` semantics so it's familiar.

### Rust changes

- New variant in `Commands` enum (`src/main.rs` near line 96, after `Languages`):
  ```rust
  InstallHooks {
      #[arg(long, value_enum, default_value_t = HookScope::User)]
      scope: HookScope,
      #[arg(long)]
      uninstall: bool,
      #[arg(long)]
      dry_run: bool,
  },
  ```
- New module `src/install_hooks.rs` with the merge logic.
- Match arm in `main()` near line 284 dispatches to the new module.
- `Cargo.toml`: add `include = ["hooks/**/*"]` to the `[package]` section so the bundled hook scripts ship with the crate when published.

### Bundled hook script

- `hooks/git-prism-redirect.sh`: ported from `~/.claude/hooks/git-prism-redirect.sh` but rewritten to:
  - Read `tool_input.command` from stdin JSON (unchanged from current).
  - Pipe the command into `python3 hooks/parse_git_invocations.py` to get the structured list of git invocations.
  - For each `{subcommand: diff|log|show|blame, ...}` entry, emit the existing redirect message and `exit 2`.
  - Otherwise `exit 0`.
- The script is self-contained; the Python helper sits beside it. Both are read by the installer at install time.

### Settings.json merge logic

- The installer reads existing settings, finds or creates the `hooks.PreToolUse` array, and adds entries with a stable `"id": "git-prism-redirect-<matcher>"` key for idempotency.
[...]

Issue #237, opened 2026-04-26
feat(tools): rewrite MCP tool descriptions for comparative framing


mikelane:
## Goal

Rewrite all four MCP tool doc comments in `src/server.rs` for comparative framing. Tool descriptions are read fresh on every tool-selection decision; gaps here cost more than any other layer in the redirect stack.

Parent epic: #234. Depends on BDD bootstrap (W1).

## Current State

| Tool | State | Action |
|------|-------|--------|
| `get_change_manifest` | Solid 17-line `///` doc with cost guidance and "first call" framing | Sharpen with explicit "instead of `git diff <ref>..<ref>`" framing and quantified token savings |
| `get_commit_history` | Bare `description = "Returns per-commit manifests for a range of commits"` argument plus a redundant 2-line `///` | Replace argument-style description with a full `///` block; compare to `git log <ref>..<ref>` and `gh pr view --json commits` |
| `get_file_snapshots` | Strong cost warning + numbered workflow | Add explicit "instead of `git show <sha>`" line; quantify "5–20k tokens per file" already there, leave it |
| `get_function_context` | **No `///` doc comment at all** | Add full doc comment from scratch — comparative framing ("use instead of `git log -S` / `git blame` for caller analysis"), value prop (callers, callees, blast radius, test references) |

## Pattern for Each Doc Comment

Every doc comment should answer, in this order:

1. **What it returns** (one sentence)
2. **What raw git command(s) it replaces** (explicit "use instead of `git X`")
3. **Quantified cost benefit** (token estimate vs raw diff/log)
4. **When to call** (positioning in the toolkit's workflow — first call, follow-up, last resort)
5. **Cost knobs** (parameters that bound response size)

## Test Surface

- `src/server.rs` already has tests at lines ~899–1046 (`it_matches_expected_tools` and schema-description tests). Extend these to assert each tool's description includes:
  - The phrase `"instead of"` or `"use this instead"` (or equivalent), AND
  - At least one quoted git command being replaced (e.g., `` `git diff` ``)
- Snapshot test (insta) for the full description text per tool, so future drift is reviewable.

## Definition of Done

- [ ] `get_function_context` has a full `///` doc comment matching the pattern above
- [ ] `get_commit_history` description is a `///` block (no more `description = "..."` argument)
- [ ] `get_change_manifest` and `get_file_snapshots` doc comments include explicit `git X` references
- [ ] `it_matches_expected_tools` extended to assert comparative framing on all four tools
[...]

Issue #236, opened 2026-04-26
test(bdd): scaffold redirect hook + review_change scenarios


mikelane:
## Goal

Write all Gherkin scenarios for epic #234 and Python+behave step definitions that fail meaningfully (real CLI / real hook script invocation, never `raise NotImplementedError`). Tag every scenario with the implementation issue number that will make it pass and `@not_implemented` so it's excluded from CI until that issue lands.

Parent epic: #234. Depends on ADR `0008-redirect-hook-architecture.md` (delivered in #235).

### Tool description rewrites (→ #237)
- `@ISSUE-237` Verify all four MCP tool doc comments include comparative framing vs raw git equivalents (asserted via the `it_matches_expected_tools` test extension or a new behave step that introspects the MCP server schema).

### Python bash tokenizer (→ #238)
- `@ISSUE-238` Tokenizer recognizes `git diff main..HEAD` in a plain command
- `@ISSUE-238` Tokenizer recognizes `git diff main..HEAD` inside `cd /tmp && git diff main..HEAD`
- `@ISSUE-238` Tokenizer recognizes `git log main..HEAD` inside a subshell `(git log main..HEAD)`
- `@ISSUE-238` Tokenizer recognizes `git diff $BASE..HEAD` (variable substitution) without expanding `$BASE`
- `@ISSUE-238` Tokenizer recognizes `git blame src/server.rs`
- `@ISSUE-238` Tokenizer ignores `git status`, `git add`, `git commit`, `git push`, `git fetch`
- `@ISSUE-238` Tokenizer handles heredocs (skips embedded text, parses surrounding commands)

### Install-hooks subcommand + bundled hook script (→ #239)
- `@ISSUE-239` `git-prism install-hooks --scope user` writes expected entries to `~/.claude/settings.json`
- `@ISSUE-239` `git-prism install-hooks --scope project` writes to `<repo>/.claude/settings.json` (creating the file/dir if missing)
- `@ISSUE-239` Re-running `install-hooks --scope user` is idempotent (file content identical before and after second run)
- `@ISSUE-239` `git-prism install-hooks --scope user --uninstall` removes only entries this command added (preserves user's other hooks)
- `@ISSUE-239` Bundled hook redirects on the same compound/subshell/variable cases the tokenizer handles (end-to-end via shell-out from the step definition)
- `@ISSUE-239` Bundled hook outputs hard-block exit code 2 + advisory message text matching the current local hook's tone

### review_change MCP tool (→ #240)
[...]

PR #243, opened 2026-04-26
docs(adr): 0008 redirect hook architecture (#235)

### LINKED ISSUE #235: Spike: redirect hook architecture and ADR 0008
## Goal

Decide the architecture for the bundled redirect-hook system before any implementation begins. Output is a single ADR — no production code merges from this issue.

Parent epic: #234

## Questions to Answer

1. **Bash command parser library.** `shlex` (stdlib) vs `bashlex` (third-party PyPI dep) vs a small handwritten recursive-descent parser. The bundled hook needs to handle: `&&`, `||`, `;`, subshells `(...)`, command substitution `$(...)`, variable expansion (don't *evaluate* — just recognize that `git diff $X..HEAD` is still a `git diff`), and heredocs (recognize and skip). What's the smallest dependency footprint that gets us reliable detection?

2. **`--scope` semantics.** Mirror `claude mcp add --scope user|project|local`? Where exactly does each one write? What about precedence when entries exist in multiple scopes? What about the cases where the directory doesn't exist yet (e.g., `<repo>/.claude/` not present)?

3. **Hook script idempotency.** The hook installer must merge into existing `settings.json` without duplicating entries on re-install. What's the merge key — exact-match on `command` field? A sentinel like `"id": "git-prism-redirect"` injected into each entry? How do we handle a user who edited the installed entry and re-runs install?

4. **BDD testability.** Hook scripts run inside the Claude Code harness, not inside `git-prism` itself. How do we test the hook script from `behave`? Options: shell out to the bash script with mock JSON on stdin and assert on exit code + stderr; use a testing matrix that exercises the Python tokenizer directly; some hybrid. The tests must be hermetic and fast.

5. **Subagent MCP scope bug verification.** Issues `anthropics/claude-code#13605` and `anthropics/claude-code#13898` allegedly cause subagents to fail to call MCP tools from project-scoped `.mcp.json`. Confirm whether these issues exist, whether the bug applies to git-prism users at project scope, and whether the workaround (user-scoped registration) should be the default recommendation in the ADR.

## Definition of Done

- [ ] Spike branch `spike/redirect-hook-architecture` created and used (never merged to main)
- [ ] ADR file `docs/decisions/0008-redirect-hook-architecture.md` committed to main, structured per the format of `docs/decisions/0007-pr-125-squash-merge-post-mortem.md` (Status / Date / Context / Decision / Consequences / Prevention / Alternatives Considered / References)
- [ ] ADR records the chosen parser, the `--scope` semantics, the idempotency strategy, the BDD test approach, and the subagent-MCP-bug verification result
- [ ] Spike branch deleted after the ADR lands
- [ ] This issue is closed and linked to the ADR commit

## Out of Scope

- Production code in main (only the ADR lands)
- Implementation of the parser, hook script, or installer (that's W3 and W4)
- BDD scenarios (that's W1)

## Notes

The ADR drives the rest of the epic. W1 (BDD bootstrap) reads the ADR before writing scenarios. W3/W4 implement what the ADR specifies. If a decision in the ADR turns out to be wrong during implementation, amend the ADR — don't just diverge silently.

mikelane:
## Summary
- ADR 0008 records the architectural decisions for epic #234's redirect hook stack
- Five questions answered: bash parser library (`shlex` stdlib, no third-party deps), `--scope` semantics (mirror `claude mcp add`), idempotency (`id` sentinel field), BDD testability (subprocess shell-out, hermetic via tempdir), and subagent MCP scope bug verification (issue #13898 still open — defaults `--scope user` to avoid silently breaking subagents)
- 9 alternatives considered with concrete rejection reasons
- Closes #235

## SDLC gates
- Spec review (adversarial-qa): APPROVED after one revision cycle
- Bug hunting / fact-check: gh api outputs verified, issue numbers correct, REPL transcripts match shlex behavior
- Security audit: no secrets, all matches false-positive on the word "token" in technical prose
- Commit hygiene: conventional format, scoped to issue #235

## Spike discipline
Spike branch \`spike/redirect-hook-architecture\` is disposable; only the ADR file lands. Prototype code (untracked \`prototypes/\` directory in the worktree) is discarded with the branch.

### DIFF
--- a/docs/decisions/0008-redirect-hook-architecture.md
+++ b/docs/decisions/0008-redirect-hook-architecture.md
@@ -25,19 +25,28 @@ This ADR records the architectural decisions for that work. No production code l
 
 Use `shlex.shlex(posix=True, punctuation_chars=True, whitespace_split=True)` as the primary tokenizer, with two small wrappers around it. No third-party parser dependency.
 
-**What `shlex` handles natively** (verified against representative inputs in this spike's prototype):
+**What `shlex` handles natively** — evidence from the Python 3.13 REPL (run on the spike branch, 2026-04-26):
 
-- compound operators `&&`, `||`, `|`, `;` are emitted as standalone tokens (with `punctuation_chars=True`)
-- subshell parens `(`, `)` are standalone tokens
-- variable forms `$X`, `$X..HEAD`, `${X}` are preserved as literal tokens (we do not evaluate)
-- `$(...)` decomposes into `$`, `(`, ..., `)` tokens — usable as command-substitution delimiters
-- single- and double-quoted strings are unwrapped to their content
-- escaped whitespace (`git\ diff`) joins tokens correctly
-- newlines collapse to whitespace, so multiline bash works
+```
+>>> import shlex
+>>> def tok(s):
+...     lex = shlex.shlex(s, posix=True, punctuation_chars=True)
+...     lex.whitespace_split = True
+...     return list(lex)
+...
+>>> tok('cd /tmp && git diff main..HEAD')
+['cd', '/tmp', '&&', 'git', 'diff', 'main..HEAD']
+>>> tok('(git log main..HEAD)')
+['(', 'git', 'log', 'main..HEAD', ')']
+>>> tok('git diff $BASE..HEAD')
+['git', 'diff', '$BASE..HEAD']
+```
+
+Compound operators (`&&`) become their own token; subshell parens split cleanly; variable forms pass through as literal tokens (we never evaluate). The remaining shapes — `||`, `|`, `;`, `${X}`, `$(...)`, single/double quotes, escaped whitespace, multiline input — behave the same way; the three above are sufficient evidence to anchor the design choice.
 
 **What we wrap around `shlex`:**
 
-1. **Heredoc body skipping.** `shlex` emits `<<EOF`, then every body word as separate tokens, then the closing `EOF`. A `git diff a..b` literally appearing inside a heredoc body would be a false positive. Solve by walking the token stream once: when we see `<<TAG` (or `<<-TAG`, `<<"TAG"`, `<<'TAG'`), drop tokens until we re-encounter `TAG` at the start of a logical line.
+1. **Heredoc body skipping.** `shlex` emits `<<EOF`, then every body word as separate tokens, then the closing `EOF`. A `git diff a..b` literally appearing inside a heredoc body would be a false positive. The "start of a logical line" heuristic that bash uses for heredoc termination does not survive `shlex`'s default whitespace handling — newlines collapse to spaces, so we cannot tell where one logical line ends and the next begins. Solution: configure the lexer to keep newlines as their own token (`lex.whitespace = ' \t'`, leaving `\n` out of the whitespace set). The heredoc walker then waits for the sequence `<newline-token>, <TAG>, <newline-token>` (or end-of-stream after `<TAG>`) before resuming normal scanning. Inline `TAG` occurrences inside the body are ignored because they are not preceded by a newline token. This preserves bash semantics without re-implementing them.
 
 2. **Backtick stripping.** Backticks (`` `cmd` ``) attach to adjacent tokens (`` `git `` → single token). Replace stray backticks with whitespace before feeding the input to `shlex`. This treats backticks as command-substitution delimiters identically to `$(...)`.
 
@@ -62,6 +71,8 @@ Local scope writes to `<repo>/.claude/settings.local.json`, which is per-checkou
 | `local` | `<repo>/.claude/settings.local.json` | `<repo>/.claude/hooks/git-prism-redirect.sh` (copied) | no (gitignored) |
 
 The Python tokenizer (`parse_git_invocations.py`) is copied alongside the shell script in each case. This is intentional duplication — keeps each scope self-contained and avoids cross-scope path resolution at hook execution time.
+
+**Project + local scope share the same script path.** Both `--scope project` and `--scope local` write to `<repo>/.claude/hooks/git-prism-redirect.sh`. If a user installs at both scopes in the same repo, the second install overwrites the script. This is idempotent — both installs ship identical script bytes from the same git-prism binary — so the overwrite is a no-op on disk. The only differences are the two settings files (`.claude/settings.json` vs. `.claude/settings.local.json`), which Claude Code merges. Documenting this so it does not look like a bug: the shared path is intentional, and it's why we keep the script byte-identical across scopes.
 
 **Precedence when entries exist in multiple scopes:**
 
@@ -103,13 +114,29 @@ If we find an entry with our `id` whose `command`, `matcher`, or other fields ha
 
 The version suffix (`-v1`) lets future schema changes detect old entries cleanly: when we ship `v2`, the installer replaces all `git-prism-redirect-v1` entries with `git-prism-redirect-v2` versions. Old uninstalls remain possible via `--uninstall --version v1` for users who want to roll back.
 
+**Mixed-version installs (downgrade after upgrade):**
+
+Version migration is monotonic in the forward direction (v1 → v2 replaces the v1 entry). Going backward is the failure mode: if the user upgrades to a `git-prism` that writes `v2`, then runs `install-hooks` from an older binary that only knows about `v1`, the older binary appends a fresh `v1` entry next to the existing `v2` entry. Both fire on every Bash call — the redirect runs twice.
+
+Mitigation: the installer's idempotency lookup matches any `id` matching the prefix `git-prism-redirect-v` regardless of suffix. If a higher version number is present and we are about to write a lower one, abort with: "Found `git-prism-redirect-v2` entry; this binary writes `v1`. Run the newer git-prism's `install-hooks --uninstall` first, or upgrade this binary." This catches the downgrade case without baking the version table into the older binary (which can't know what versions exist in the future).
+
 ### 4. BDD testability: subprocess shell-out with mock JSON on stdin
 
 Step definitions in `bdd/steps/` shell out to the bundled `hooks/git-prism-redirect.sh` (with `parse_git_invocations.py` next to it on `PYTHONPATH`) using `subprocess.run`. The Gherkin scenarios feed mock JSON on stdin, then assert on:
 
-- exit code: `0` (allow), `2` (block)
+- exit code: `0` (allow or advise), `2` (hard block)
 - stderr text: must contain a redirect message
 - stdout JSON: when emitted (advisory mode), must validate against the Claude Code hookSpecificOutput schema
+
+**Three exit-code states (matches the existing `~/.claude/hooks/git-prism-redirect.sh` prototype):**
+
+| Exit | Stdout | Stderr | Meaning |
+|---|---|---|---|
+| `0` | empty | empty | allow — command runs, hook says nothing |
+| `0` | JSON with `hookSpecificOutput.additionalContext` | optional | advise — Claude Code surfaces the message but the command still runs |
+| `2` | empty | redirect message | hard block — command is rejected, stderr is shown to the agent |
+
+The "advise" state is what the bundled hook emits for the rows in the coverage matrix marked "advisory": exit `0` with a JSON payload containing the redirect text in `hookSpecificOutput.additionalContext`. This matches Claude Code's PreToolUse hook protocol (the harness reads stdout JSON when exit is `0` and treats `additionalContext` as a non-blocking nudge). Exit `2` is reserved for the cases where redirection is non-negotiable (e.g., `gh pr diff`, MCP GitHub tools — see the matrix).
 
 **Hermetic constraints:**
 
@@ -140,10 +167,25 @@ We will, for the parser logic specifically (Pythonic unit tests under `hooks/tes
 
 ### 5. Subagent MCP scope bug: real, relevant, default to `--scope user`
 
-Verified both issues exist (queried `gh api repos/anthropics/claude-code/issues/...`):
+Verified both issues exist via the GitHub API on 2026-04-26:
 
-- **#13605** ([link](https://github.com/anthropics/claude-code/issues/13605), closed 2026-03-25): "Custom plugin subagents cannot access MCP tools (built-in agents can)." Workaround documented: use `general-purpose` built-in agent. Marked as resolved by reporter, but the resolution is "use the workaround" — the underlying behavior is unchanged.
-- **#13898** ([link](https://github.com/anthropics/claude-code/issues/13898), still **open** as of 2026-04-21): "Custom Subagents Cannot Access Project-Scoped MCP Servers (Hallucinate Instead)." Custom subagents in `.claude/agents/` cannot call tools from `.mcp.json` (project scope) — they hallucinate plausible-but-incorrect responses. **Globally configured MCP servers (`~/.claude/mcp.json`) work correctly in the same subagents.** The reporter's test matrix is unambiguous.
+```
+$ gh api repos/anthropics/claude-code/issues/13605 --jq '{title, state, closed_at}'
+{
+  "title": "[BUG] Custom plugin subagents cannot access MCP tools (built-in agents can)",
+  "state": "closed",
+  "closed_at": "2026-03-25T22:54:47Z"
+}
+
+$ gh api repos/anthropics/claude-code/issues/13898 --jq '{title, state}'
+{
+  "title": "Custom Subagents Cannot Access Project-Scoped MCP Servers (Hallucinate Instead)",
+  "state": "open"
+}
+```
+
+- **#13605** ([link](https://github.com/anthropics/claude-code/issues/13605), closed 2026-03-25): custom plugin subagents cannot access MCP tools; built-in agents can. Workaround documented: use the `general-purpose` built-in agent. Marked resolved, but the resolution is "use the workaround" — the underlying behavior is unchanged.
+- **#13898** ([link](https://github.com/anthropics/claude-code/issues/13898), still **open** as of the query above): custom subagents in `.claude/agents/` cannot call tools from `.mcp.json` (project scope) — they hallucinate plausible-but-incorrect responses. **Globally configured MCP servers (`~/.claude/mcp.json`) work correctly in the same subagents.** The reporter's test matrix is unambiguous.
 
 **Relevance to git-prism:**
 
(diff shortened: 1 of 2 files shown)


docs/decisions/0008-redirect-hook-spike.md
```diff
@@ -0,0 +1,258 @@
+# ADR 0008: Redirect Hook Architecture
```
mikelane, line 1: **[NAMING — CRITICAL]** Filename does not match the established `-spike` suffix used by sibling spike ADRs in this directory (0003/0004/0005/0006 are all `-spike`). The branch is `spike/redirect-hook-architecture`, the issue (#235) DoD frames this as a spike, and `CLAUDE.md` explicitly states "The spike lives on a `spike/<topic>` branch that is never merged. Its only deliverable is an ADR in `docs/decisions/NNNN-short-title.md`."

Recommendation: rename to `0008-redirect-hook-spike.md` before merge so future readers can grep `-spike` to see all exploratory work. This is the lone blocker on the PR.

Note: 0001/0002/0007 are not `-spike` either, but those are working-tree-diff/mutation-baseline/post-mortem — not exploratory feasibility studies. The 0003-0006 cluster is the relevant precedent.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+| `gh pr diff 123` | hard block (exit 2) |
+| `mcp__github__get_commit` tool input | hard block |
+| Empty stdin | exit 0, no output |
+| Malformed JSON | exit 0 (fail open), warning on stderr |
+
+**Why not unit-test the Python tokenizer in isolation:**
+
+We will, for the parser logic specifically (Pythonic unit tests under `hooks/tests/test_parse_git_invocations.py`, run by `pytest`). But the behavioral contract — "this exit code, this stderr, given this stdin JSON" — is what Claude Code actually exercises, and that contract is the load-bearing surface. The BDD layer tests it end-to-end. Unit tests catch regressions in the parser internals; BDD catches regressions in the wire protocol. We need both.
+
+### 5. Subagent MCP scope bug: real, relevant, default to `--scope user`
```
mikelane: **[NAMING — MEDIUM]** Heading 5 reads as comma-separated assertions ("real, relevant, default to `--scope user`") rather than a noun phrase like the other six decision headings (e.g., "Bash command parser: ...", "Idempotency: sentinel field ...", "Fail-open when `python3` is missing"). Consider rewording to a noun phrase such as "Subagent MCP scope bug workaround: default to `--scope user`" so the heading scans as a decision name rather than a sentence fragment. Deferrable to a follow-up commit on this branch.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Missing target directory:**
+
+If `<repo>/.claude/` does not exist for `--scope project` or `--scope local`, the installer creates it (`mkdir -p`). For `--scope user`, `~/.claude/` always exists if Claude Code is running. We never create `~/.claude/` ourselves — if it is missing, we error with "Claude Code does not appear to be installed".
+
+**Discovery (`install-hooks --status`):**
+
+The installer's `--status` flag (no positional args) reads all three settings files and prints a table of where git-prism entries are installed and which version of the hook script each one points at. This makes "is it installed?" answerable without `cat ~/.claude/settings.json | jq`.
+
+### 3. Idempotency: sentinel field `id: "git-prism-redirect-vN"`
```
mikelane: **[NAMING — MEDIUM]** The sentinel id `git-prism-redirect-vN` is generic. Future schema changes might introduce a separate redirect for MCP tool inputs (already implied by the `gh pr diff` row in the coverage matrix), or for `Edit`/`Write` tools — at which point "redirect" alone disambiguates nothing. Consider `git-prism-bash-redirect-vN` to anchor the matcher dimension in the id itself. The version suffix (v1) is fine as-is. Deferrable; renaming the sentinel before any user has v1 installed is cheap.

docs/decisions/0008-redirect-hook-spike.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Hermetic constraints:**
+
+- Tests must NOT mutate `~/.claude/settings.json`. Use a `tempfile.TemporaryDirectory()` per scenario, set `HOME` to it, run the install/uninstall commands inside that sandbox, then assert on the resulting tempdir contents. Tear down by letting the tempdir context manager exit.
+- Tests must NOT spawn a real Claude Code process. The hook script reads stdin and exits — that's the entire contract. The behave tests exercise that contract directly.
+- Test fixtures live under `bdd/fixtures/hook_inputs/` as JSON files, named after the scenario they support (e.g., `bash_compound_diff.json`, `bash_status_only.json`).
+
+**Coverage matrix the BDD scenarios must cover:**
+
+| Input shape | Expected behavior |
```
mikelane, line 149: **[TEST — MEDIUM]** Coverage matrix is missing rows for several shapes the ADR itself flags as critical surface:

- command substitution `$(...)` (called out in Context as a missed case in the regex hook)
- `||` (or-list)
- `;` (sequential)
- backticks (Decision 1.2 explicitly normalizes them)
- fail-open path: `python3` missing on PATH (Decision 6 promises exit 0 + stderr warning)
- `--force` overwrite path (Decision 3)
- `--status` discovery output (Decision 2)
- `--uninstall --version v1` rollback (Decision 3)
- downgrade-version idempotency / abort path (Decision 3, "Mixed-version installs")

Each of these is a discrete behavior the BDD layer is the load-bearing test for, per Decision 4 itself. Deferrable to BDD bootstrap (#236), but the matrix in the ADR should at minimum enumerate them so #236 has a complete checklist.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+| `git status` | no advisory |
+| `git add file.txt` | no advisory |
+| `gh pr diff 123` | hard block (exit 2) |
+| `mcp__github__get_commit` tool input | hard block |
+| Empty stdin | exit 0, no output |
+| Malformed JSON | exit 0 (fail open), warning on stderr |
+
+**Why not unit-test the Python tokenizer in isolation:**
+
+We will, for the parser logic specifically (Pythonic unit tests under `hooks/tests/test_parse_git_invocations.py`, run by `pytest`). But the behavioral contract — "this exit code, this stderr, given this stdin JSON" — is what Claude Code actually exercises, and that contract is the load-bearing surface. The BDD layer tests it end-to-end. Unit tests catch regressions in the parser internals; BDD catches regressions in the wire protocol. We need both.
```
mikelane: **[TEST — MEDIUM]** Decision 1.1 acknowledges that `shlex` newline handling is the load-bearing trick for heredoc termination, but the parser unit-test matrix is not enumerated in the ADR. At minimum the cases that should be called out for `pytest` coverage in `hooks/tests/test_parse_git_invocations.py`:

- `<<-EOF` (tab-stripping form)
- `<<'EOF'` (quoted heredoc — no expansion, but still terminated by `EOF`)
- multiple heredocs in one command (`cmd <<A <<B`)
- `EOF` token appearing inline in the body without surrounding newlines (must NOT terminate)
- malformed input: missing closing `EOF` until end-of-stream (must not crash)

Decision 4 says "Unit tests catch regressions in the parser internals" — listing the specific regressions makes that promise enforceable. Deferrable to #236.

docs/decisions/0008-redirect-hook-spike.md
```diff
@@ -0,0 +1,258 @@
...
+- **Two wrapper functions added to the parser.** Heredoc skipping and backtick normalization. Both are <30 lines, both are unit-testable in isolation. The complexity budget for "things `shlex` doesn't handle" is bounded.
+- **One new CLI subcommand.** `git-prism install-hooks` with flags `--scope`, `--uninstall`, `--force`, `--status`. Wired in `src/main.rs` like `serve` / `manifest`.
+- **Hook entries gain a stable `id`.** Future schema changes can migrate cleanly. The version suffix is part of the contract.
+- **`--scope user` is the documented default.** The README and `--help` text both explain why. When upstream #13898 closes, this default becomes a candidate for revisit.
+- **BDD tests use subprocess shell-out, not in-process import.** Step definitions stay in Python (cross-language to Rust production code, per project BDD policy). Hermeticity comes from `tempfile`-scoped HOME, not from mocking the script's contents.
+- **No third-party Python parser dependency.** The hook works on a stock `python3` install.
+
+## Prevention
+
```
mikelane, line 222: **[TEST — MEDIUM]** No mutation-score threshold specified for the new Python tokenizer module. ADR 0002 establishes mutation testing as a first-class signal for this repo, and parsers are the highest-yield mutation target — boundary-condition mutations (off-by-one in heredoc termination, swapped `&&`/`||` split, dropped `\\n` token) are exactly what mutation testing catches that line-coverage misses.

Recommendation: set a concrete threshold (e.g., "≥85% mutation score on `parse_git_invocations.py` enforced in CI") in the Prevention section. Without one, the parser will inevitably ship with surviving mutants that escape PR review. Deferrable to #236 / W3 tokenizer issue.

docs/decisions/0008-redirect-hook-spike.md
```diff
@@ -0,0 +1,258 @@
...
+- The `id` sentinel must be enforced in the writer and tested in BDD: a re-install scenario with an existing entry must produce no duplicates. If the sentinel is dropped or renamed, idempotency breaks silently — the BDD scenario catches it.
```
mikelane, line 223: **[TEST — MEDIUM]** Idempotency prevention claim ("a re-install scenario with an existing entry must produce no duplicates") covers only one of the four algorithm paths from Decision 3:

1. existing entry with matching id + matching path → no-op (the one currently asserted)
2. existing entry with matching id + different path → replace + log "updated"
3. existing entry with matching id but hand-edited fields → skip with warning (or `--force` overwrite)
4. existing higher-version entry (v2) when this binary writes v1 → abort with downgrade error

Each is a distinct branch in the writer logic and each can regress independently. Recommendation: expand the Prevention bullet to enumerate all four paths and tie each to a specific BDD scenario id. Deferrable to #236.

docs/decisions/0008-redirect-hook-spike.md
```diff
@@ -0,0 +1,258 @@
...
+- stderr text: must contain a redirect message
+- stdout JSON: when emitted (advisory mode), must validate against the Claude Code hookSpecificOutput schema
+
+**Three exit-code states (matches the existing `~/.claude/hooks/git-prism-redirect.sh` prototype):**
+
+| Exit | Stdout | Stderr | Meaning |
+|---|---|---|---|
+| `0` | empty | empty | allow — command runs, hook says nothing |
+| `0` | JSON with `hookSpecificOutput.additionalContext` | optional | advise — Claude Code surfaces the message but the command still runs |
+| `2` | empty | redirect message | hard block — command is rejected, stderr is shown to the agent |
```
mikelane, line 137: **[NAMING — LOW]** "advise" (Decision 4 table, "advise" row) and "advisory" (Decision 4 prose, coverage matrix column) are used interchangeably for the middle exit-code state. Pick one register and use it consistently — either both verbs ("allow / advise / block") or both noun-phrases ("allow / advisory / hard block"). The current mix reads as inconsistent register. Trivial follow-up commit.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+Agents reach for raw `git diff/log/show/blame` instead of git-prism MCP tools because of pretraining bias. The current local hook (`~/.claude/hooks/git-prism-redirect.sh`) backstops this with regex pattern matching against `tool_input.command`, but the regex approach misses common shapes:
+
+- compound commands: `cd /tmp && git diff main..HEAD`
+- subshells: `(git log main..HEAD)`
+- variable expansion: `git diff $BASE..HEAD`
+- pipelines: `git diff main..HEAD | grep foo`
+- heredocs that quote git commands inside the body
+- command substitution: `cd $(git rev-parse --show-toplevel) && git diff ...`
+
+The hook also lives in user dotfiles, so installing git-prism gets the binary but none of the redirect leverage. Epic #234 ships a `git-prism install-hooks` subcommand and a bundled hook that calls a stdlib-only Python tokenizer to parse bash structurally.
```
mikelane: **[NAMING — LOW]** `install-hooks` is the first hyphenated subcommand on `git-prism`. The existing subcommands listed in CLAUDE.md (`serve`, `manifest`, `snapshot`, `history`, `context`, `languages`) are all single words. As the command grows (the ADR already implies `--status`, `--uninstall`, `--force` flags), a noun-first grouping would scale better:

```
git-prism hooks install [--scope ...]
git-prism hooks uninstall [--version v1]
git-prism hooks status
```

This also matches the parallel design with `claude mcp add` / `claude mcp list` cited in Decision 2. Deferrable, but cheaper to decide before any user has `install-hooks` in muscle memory than after.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Decision:**
+
+- **Default `git-prism install-hooks --scope user`.** Document the reason in the CLI help text: "User scope is the default because Claude Code issue anthropics/claude-code#13898 prevents custom subagents from calling project-scoped MCP servers correctly."
+- **Document the same caveat in the README install-hooks section**, with a link to the upstream issue.
+- **Do not block `--scope project` or `--scope local`** — power users may have their own reasons (e.g., team-shared config in a monorepo, no subagent usage). Just don't make it the easy path.
+- **Re-evaluate when #13898 closes.** Add a TODO comment in `install-hooks` source that links the issue. When it closes, revisit whether `--scope project` should become the default for the install-hooks command (consistent with how MCP server registration in `.mcp.json` works).
```
mikelane: **[TEST — LOW]** The decision to default `--scope user` is anchored on issue #13898 still being open. There is no re-verification mechanism — no script, no CI check, no scheduled re-poll. When #13898 closes, the trigger to revisit is purely "someone remembers."

Recommendation: a small CI job (or a scheduled `claude /loop` task) that runs `gh api repos/anthropics/claude-code/issues/13898 --jq .state` and opens a follow-up issue if the state ever flips to `closed`. Cheap to add, eliminates the "ADR drifts from reality" failure mode. Deferrable to a follow-up issue.

docs/decisions/0008-redirect-hook-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+- **No third-party Python parser dependency.** The hook works on a stock `python3` install.
+
+## Prevention
+
+- The `id` sentinel must be enforced in the writer and tested in BDD: a re-install scenario with an existing entry must produce no duplicates. If the sentinel is dropped or renamed, idempotency breaks silently — the BDD scenario catches it.
+- The shlex-edge-case list (heredoc body, backticks) must have explicit BDD scenarios, not just unit tests. If someone "improves" the parser later and accidentally regresses one of these, the BDD layer flags it.
+- The `--scope user` default must be documented in three places (CLI `--help`, README, this ADR). Drift on the default risks reverting a deliberate decision. When #13898 closes, the revisit must update all three together.
+- `parse_git_invocations.py` must be importable in tests as `from hooks.parse_git_invocations import tokenize_command`. The hook shell script must invoke it via `python3 -m parse_git_invocations` so the test path and the runtime path use the same module-loading pattern.
```
mikelane: **[NAMING — LOW]** `parse_git_invocations.py` describes only the parser half. Decision 1 explicitly says the module also walks the token list and applies redirect-pattern matchers ("starts with `git diff` and contains `..`") — that is the decision half, not the parse half.

Two cleaner options:

1. Split the module: `tokenize_command.py` (pure parser) + `redirect_decision.py` (pattern matchers). Better separation of concerns, each unit-testable in isolation.
2. Rename the single module to something that names the whole job: `redirect_command_check.py` or `git_redirect_decision.py`.

Option 1 is the more idiomatic split given the mutation-testing argument in the other comment — pure parsers and decision tables mutate very differently. Deferrable.

mikelane: **[PYTHON — LOW]** The Prevention section asserts two import patterns:

- tests: `from hooks.parse_git_invocations import tokenize_command`
- runtime: `python3 -m parse_git_invocations`

These two patterns require different `sys.path` shapes:

- The first requires the parent of `hooks/` to be on `sys.path` (so `hooks` is an importable package — needs `hooks/__init__.py`).
- The second requires `hooks/` itself to be on `sys.path` (so `parse_git_invocations` is a top-level module reachable via `-m`).

Reconciling these requires either: (a) `pytest` setting `pythonpath = ["."]` in `pyproject.toml` AND the hook script doing `cd "$(dirname "$0")" && python3 -m parse_git_invocations`, or (b) a `conftest.py` that adjusts `sys.path` for tests. The ADR does not commit to one. Recommend stating the resolution explicitly so the W3 implementer does not have to re-derive it. Deferrable to W3.

mikelane (commented):
## Review Summary

This is a strong spike ADR. The decisions are evidence-anchored (live REPL transcripts for `shlex` behavior, pasted `gh api` responses for the upstream MCP scope bug verification), and the alternatives section earns its keep — nine rejected options with concrete reasoning is the right amount of due diligence for a load-bearing architectural choice. The "fail-open when `python3` is missing" decision and the "`--scope user` because of upstream issue #13898" decision are both well-argued and would survive adversarial review on their own merits.

### Decision

**CHANGES REQUESTED** — single blocker on filename naming.

### Blockers (must fix before merge)

1. **CRITICAL: Filename does not match the established `-spike` suffix** for spike ADRs in this directory (0003/0004/0005/0006 all carry it; 0001/0002/0007 don't, but those aren't feasibility spikes). The branch is `spike/redirect-hook-architecture`, the issue (#235) is framed as a spike, and `CLAUDE.md` is explicit: "The spike lives on a `spike/<topic>` branch that is never merged. Its only deliverable is an ADR in `docs/decisions/NNNN-short-title.md`." Rename to `0008-redirect-hook-spike.md` before merge so the next reader can grep `-spike` and see all exploratory work.

### Recommendations (deferrable, do not block merge)

The remaining 11 inline findings split cleanly across three buckets:

- **Defer to BDD bootstrap (#236):** coverage matrix completeness (command substitution, `||`, `;`, backticks, fail-open, `--force`, `--status`, `--uninstall`, downgrade), heredoc parser test matrix (`<<-EOF`, quoted heredoc, multiple, malformed), mutation-score threshold for the new tokenizer, four-path idempotency prevention.
- **Defer to W3 / W4 implementation issues:** module split or rename for `parse_git_invocations.py`, `sys.path` reconciliation between `from hooks.parse_git_invocations import ...` (tests) and `python3 -m parse_git_invocations` (runtime).
- **Trivial follow-up commit on this branch (or post-merge fixup):** Decision 5 heading wording (sentence fragment vs. noun phrase), sentinel id specificity (`git-prism-bash-redirect-vN`), `advise`/`advisory` register, `install-hooks` vs. `git-prism hooks install` subcommand grouping, upstream-issue re-verification cadence.

### Praise

- The Python 3.13 REPL transcript inline in Decision 1 is exemplary — readers can reproduce the evidence in seconds rather than taking the ADR's word for it.
- The "Why a sentinel `id` field over command-string match" sub-decision in Decision 3 (and the explicit rejection of the comment-marker alternative) is the level of "show your work" that makes ADRs durable.
- The `claude mcp add` parallel design rationale in Decision 2 — and the deliberate divergence on the default scope — is well-justified, with both reasons (subagent MCP bug AND user-scope-applies-everywhere) explicitly enumerated.
- The "Mixed-version installs (downgrade after upgrade)" subsection of Decision 3 is the kind of edge case authors usually omit and field reports surface six months later. Catching it in the spike is the right call.
- The Alternatives Considered section, particularly the rejection of `bashlex` (license + bootstrap cost) and tree-sitter (process boundary), is thorough enough that future revisitors won't have to re-litigate.

### Security

Zero findings. No code lands; future implementation issues (#237, #239) get their own gauntlet.

### Solo-dev note

Per `CLAUDE.md`, GitHub doesn't allow repo owners to formally approve their own PRs, so this is posted as a `COMMENT` review rather than `APPROVE`. The filename rename is the only blocker — once `0008-redirect-hook-architecture.md` is renamed to `0008-redirect-hook-spike.md`, the PR can be merged.

PR #244, opened 2026-04-27
test(bdd): scaffold redirect hook + review_change scenarios (#236)

mikelane:
## Summary

BDD bootstrap for epic #234. All 44 scenarios across the four implementation issues are tagged `@ISSUE-NN @not_implemented` and currently RED — they fail with assertion errors against a real `git-prism` binary and the (not-yet-existing) bundled hook script.

Closes #236.

## What's new

| Tag | Scenario count | Covers |
|---|---|---|
| `@ISSUE-237` | 1 | Tool-description comparative framing (substring assertion; deep insta snapshot deferred to #237 PR) |
| `@ISSUE-238` | 12 | Python tokenizer matrix: plain, compound, subshell, var-expansion, blame, write-side, heredoc default + `<<-EOF` + `<<'EOF'`, pipeline, command substitution `\$(...)`, backticks, `\$BASE` no-leak triangulation |
| `@ISSUE-239` | 18 | install-hooks subcommand: scope user/project/local, idempotency paths 2/3/4a/4b, mixed-version downgrade abort, `--force`, `--dry-run`, `hooks status`, duplicate-scope warning, `gh pr diff` and `mcp__github__*` hard-block, fail-open paths (empty stdin, malformed JSON, missing python3) |
| `@ISSUE-240` | 4 | review_change tool: combined payload, 40/60 split with two budget values, cursor-walks-to-different-page, doc comment framing |

## Discipline

- Step defs in Python (cross-language to Rust production) — no internal imports.
- Real `subprocess.run` shell-out to the binary at `target/release/git-prism` and the bundled hook script (when it exists). No mocks, no `pass`, no `raise NotImplementedError`.
- Hermetic: per-scenario `tempfile.TemporaryDirectory()` with `HOME` override on every shell-out (binary AND hook script). No mutation of the developer's real `~/.claude/...`.
- `before_scenario` resets all redirect-hook context attrs to defaults so state cannot leak between scenarios.

## SDLC gates

- **Spec review (adversarial-qa Gate 1)**: ran on the initial 24-scenario scaffold, found gap list (pipeline/cmd-sub/backticks, idempotency paths 3/4a/4b, mixed-version downgrade, scope local, fail-open, hooks status, --force, --dry-run, triangulation hardening). All addressed in this PR.
- **Tag fix**: \`gh pr diff\` hard-block was mis-tagged \`@ISSUE-238\`, retagged \`@ISSUE-239\` (it is bundled-hook decision logic).
[...]

PR #247, opened 2026-04-28
feat(tools): review_change MCP orchestration tool (#240)

### LINKED ISSUE #240: feat(tools): review_change MCP orchestration tool
## Goal

Add a new `review_change` MCP tool that combines `get_change_manifest` and `get_function_context` into a single response. Single-call parity with `git diff` removes the friction that pushes agents toward porcelain.

Parent epic: #234. Depends on W1 (BDD scenarios). Independent of W3/W4 — can land in parallel.

## API

```rust
#[tool(name = "review_change")]
async fn review_change(
    Parameters(args): Parameters<ReviewChangeArgs>,
) -> Result<Json<ReviewChangeResponse>, String>;

struct ReviewChangeArgs {
    repo_path: Option<String>,
    base_ref: String,
    head_ref: Option<String>,
    include_patterns: Vec<String>,
    exclude_patterns: Vec<String>,
    function_names: Option<Vec<String>>,
    max_response_tokens: usize,  // default 8192
    cursor: Option<String>,
    page_size: Option<usize>,
}

struct ReviewChangeResponse {
    manifest: ManifestResponse,
    function_context: FunctionContextResponse,
    pagination: Pagination,  // unified cursor for both halves
    metadata: ReviewMetadata,
}
```

## Token Budget Allocation

`max_response_tokens` is split between sub-responses:

- 40% manifest (file/function-level metadata)
- 60% function_context (callers, callees, blast radius — heavier per entry)

If either sub-response would exceed its share, it's truncated per the same rules each tool uses standalone (`metadata.function_analysis_truncated` for manifest, `truncated: true` flag on context entries).

## Doc Comment

This is the strongest comparative description in the epic — it's the tool that competes head-to-head with `git diff`:

```rust
/// Returns combined change manifest and function-level blast radius for a ref
/// range, in one call. Use this **instead of `git diff <ref>..<ref>`** when
/// reviewing a PR, auditing a refactor, or assessing merge safety — it answers
/// "what changed and what might break" in one tool invocation, with structured
/// JSON instead of raw diff text.
///
/// Replaces the common two-step workflow:
///   1. `get_change_manifest` — what changed
///   2. `get_function_context` — what callers might break
/// with a single call that runs both internally and splits the token budget.
///
/// Cost: ~30% of the equivalent raw `git diff <ref>..<ref> | git log -S <fn>`
/// pipeline because semantic data replaces unstructured text.
```

## Implementation

- New module `src/tools/review_change.rs` that calls into the existing manifest and context handlers internally — no logic duplication.
- Reuse `ManifestOptions`, `FunctionContextRequest`, etc. — the orchestration tool is a thin wrapper.
- Emit the standard tracing span `mcp.tool.review_change` with the same fields as the underlying tools (repo_path_hash, ref_base, ref_head, response_bytes, etc.).
- Metrics: register `review_change` with the existing `Metrics` struct (`record_request`, `record_duration`, etc.).

## Test Surface

- Unit tests for the budget-split arithmetic (40/60 partition, edge cases at 0 and overflow).
- Integration test against a fixture repo: invoke `review_change`, assert both `manifest` and `function_context` populated, response under budget.
- Snapshot test (insta) for the doc comment text — comparative framing should be a regression-tested promise.
- Update `it_matches_expected_tools` in `src/server.rs` (lines ~899–1046) to expect five tools instead of four.

## Definition of Done

- [ ] `review_change` MCP tool registered and callable
- [ ] Response combines manifest + function_context, respecting token budget
- [ ] Pagination cursor unified across both sub-responses
- [ ] Doc comment includes explicit "instead of `git diff`" framing
- [ ] `cargo test --bin git-prism` passes
- [ ] Tracing span and metrics match other tools
- [ ] All `@ISSUE-<this>` BDD scenarios pass
- [ ] `EXPECTED_TOOLS` constant updated to include `review_change`

## Notes

This tool is the primary "winner" candidate against pretraining bias. The whole epic's success depends partly on this one tool's doc comment being good enough to flip the model's default choice. Spend the time to get it right.

mikelane:
## Summary

- Adds `review_change` MCP tool: returns combined `{ manifest, function_context }` for a ref range in one call, splitting `max_response_tokens` 40/60 between the two halves
- Designed to replace `git diff <ref>..<ref>` as the agent's first call when reviewing a PR — the doc comment leads with "Use this **instead of `git diff <ref>..<ref>`**"
- Adds `metadata.budget_tokens` to both `ManifestMetadata` and `ContextMetadata` so each sub-response stamps its share of the split budget for observability
- Independent `manifest_cursor` / `function_context_cursor` cursors let each half paginate without resetting the other
- Working-tree mode (omit `head_ref`) returns the manifest half normally; the function-context half returns an empty payload rather than failing

## Test plan

- [x] `cargo test` — 698 tests pass
- [x] `python -m behave bdd/features/redirect_hook.feature --tags="@ISSUE-240"` — 5 scenarios pass
- [x] Church purification applied: BDD scenario title fixed, .gitignore hardened, metrics gap plugged, error attribution added, README params corrected, CHANGELOG dash normalized

Closes #240


src/tools/review_change.rs
```diff
@@ -0,0 +1,491 @@
...
+/// disabled" via their `Option<usize>` adapter logic.
+#[must_use]
+pub fn split_budget(budget: usize) -> (usize, usize) {
+    if budget == 0 {
+        return (0, 0);
+    }
+    let budget_f = budget as f64;
+    let manifest_share = (budget_f * 0.4) as usize;
+    let context_share = (budget_f * 0.6).round() as usize;
+    (manifest_share, context_share)
```
mikelane, line 129: Praise: this triangulation is exactly right — two BDD examples whose floor/round results land on opposite sides of the half-token boundary force the impl toward the only formula that satisfies both. The unit tests in `split_budget_at_4096_yields_*` and `split_budget_at_16384_yields_*` mirror the BDD spec, and the bonus `split_budget_manifest_plus_context_equals_or_near_budget` test asserts the `|sum - budget| <= 1` invariant for a range of budgets so a mutation that swaps floor with round on one side cannot silently slip through. The block comment that explains *why* complement arithmetic was rejected is exactly the kind of WHY commenting that earns its keep.

src/server.rs
```diff
@@ -22,6 +22,26 @@ fn change_scope_label(scope: ChangeScope) -> &'static str {
...
+            .manifest
+            .metadata
+            .function_analysis_truncated
+            .is_empty()
+        || !response
+            .function_context
+            .metadata
+            .function_analysis_truncated
+            .is_empty()
+}
```
mikelane, line 43: Praise: `review_change_response_truncated` is a quiet but excellent piece of refactoring discipline. The span-attribute and metric paths both consume the same predicate, so a future third truncation source (e.g., `included_files_truncated`) gets caught in one place rather than drifting between span and metric. Naming follows the project's existing `is_any_file_truncated` style — consistent and self-documenting.

src/tools/review_change.rs
```diff
@@ -0,0 +1,491 @@
...
+        },
+        functions: vec![],
+        pagination: PaginationInfo {
+            total_items: 0,
+            page_start: 0,
+            page_size: 0,
+            next_cursor: None,
+        },
+    }
+}
```
mikelane, line 320: **LOW** — `empty_function_context` (lines 289-320) emits a synthetic `ContextMetadata` with `head_ref: "WORKTREE"` hardcoded, and the `pagination` block uses `page_size: 0`. The standalone `build_function_context_with_options` returns the caller's requested `page_size` even on empty results (so clients can tell what they asked for). Would mirroring that here — `page_size: clamp_page_size(args.page_size)` — better preserve the contract that `page_size` reflects intent, not actual results? Not a blocker since no caller currently depends on this, but it's a subtle drift between the orchestration path and the standalone path.

mikelane, line 320: Observation (non-blocking): the working-tree path here returns a synthetic `FunctionContextResponse` with `head_ref: "WORKTREE"` rather than failing. That choice is documented and consistent with the rest of the codebase's working-tree story, but it does mean an agent that calls `review_change` in working-tree mode and then does `response.function_context.functions.is_empty() ? skip : process` will silently skip caller analysis instead of getting a clear "function context unavailable in working-tree mode" signal. The empty payload is correct; consider whether the metadata should carry an explicit `mode: "working_tree"` discriminator (or similar) in a follow-up so downstream agents can branch on it without inferring from `head_ref == "WORKTREE"`. Not blocking — the current shape matches the doc comment's contract.

src/tools/review_change.rs
```diff
@@ -0,0 +1,491 @@
...
+/// 1. Compute the 40/60 budget split via [`split_budget`].
+/// 2. Run [`build_manifest`] (or [`build_worktree_manifest`] when `head_ref`
+///    is `None`) with the manifest half of the budget and the
+///    `manifest_cursor` opaque token.
+/// 3. Run [`build_function_context_with_options`] with the function_context
+///    half of the budget and the `function_context_cursor` opaque token. The
+///    context handler reads the manifest internally for changed-function
+///    discovery, so two manifest builds happen per call — once visible to
+///    the user, once internal to the context tool. This matches the standalone
+///    `get_function_context` cost profile.
```
mikelane, line 153: **LOW** — The doc comment at lines 144-153 says `build_function_context_with_options` reads the manifest internally, causing two manifest builds per call. That's an honest architectural caveat, but it means `review_change` redundantly diff-walks the same range twice for a tool whose pitch is single-call efficiency. Would extracting a shared `compute_changed_functions(repo_path, base_ref, head_ref)` step that both halves consume be worth a follow-up issue? The cost is documented and not a blocker today, but it's the kind of duplication that compounds when the tool gets popular.

src/tools/review_change.rs
```diff
@@ -0,0 +1,491 @@
...
+//!
+//! - `manifest_budget = floor(B * 0.4)` — truncating cast `(B as f64 * 0.4) as usize`
+//! - `function_context_budget = round(B * 0.6)` — `((B as f64 * 0.6).round()) as usize`
+//!
+//! Floor is used for the smaller (manifest) share, round for the larger
+//! (function_context) share. The two budgets are computed independently rather
+//! than as complements (manifest + (B - manifest)) because the BDD scenario
+//! at `bdd/features/redirect_hook.feature:459` triangulates two budget values
+//! whose sums fall on different sides of the half-token rounding boundary —
+//! complement arithmetic would pass at one budget and fail at the other.
```
mikelane, line 28: The comment at lines 25-28 calls out the load-bearing reason for *not* using complement arithmetic (`B - manifest_share`): the BDD examples at budgets 4096 and 16384 fall on different sides of the rounding boundary. That's exactly the kind of "why" that survives a year of refactors when someone tries to "simplify" it back to complements. Excellent triangulation rationale captured directly in code.

src/server.rs
mikelane, line 43: `review_change_response_truncated` (lines 30-43) consolidates four truncation signals into one helper, used both for the span attribute and the truncation metric. Single source of truth for "is this response truncated" — exactly the right factoring. The doc comment explicitly anticipates the third source case ("keeping one helper avoids the two-place check drifting if a third truncation source is added"), which is the kind of forward-looking comment that pays for itself the first time someone adds a new metadata flag.

mikelane (commented):
## PR Review: feat(tools): review_change MCP orchestration tool (#240)

### Verdict: **APPROVED** (posted as review comment due to GitHub's self-approval restriction on solo-dev PRs)

### Summary

This is exemplary work. The PR adds a fifth MCP tool (`review_change`) that orchestrates the existing `get_change_manifest` and `get_function_context` handlers into a single combined response, splitting `max_response_tokens` 40/60 between the two halves. It is purely additive — no behavior change to existing tools, no migrations, clean revert path — and the implementation is honest pure orchestration with no duplicated diff or analysis logic.

The architectural choices are deliberate and well-justified:

- **Triangulation discipline.** The 40/60 budget split is enforced by two BDD examples whose floor/round results land on opposite sides of the half-token rounding boundary (4096 → 1638/2458, 16384 → 6553/9830). Either rounding rule applied uniformly, or complement arithmetic, fails one example. The unit test suite mirrors the BDD spec and adds an `|sum - budget| <= 1` invariant for a range of budgets so a mutation that swaps floor with round on one side cannot silently slip through. This is exactly how to write tests that resist mutation testing.
- **Single source of truth for the truncation predicate.** `review_change_response_truncated` consolidates the four truncation checks (two pagination cursors plus two `function_analysis_truncated` lists) so the span attribute and the metric path can never drift apart.
- **Independent cursors.** `manifest_cursor` and `function_context_cursor` advance the two halves separately — paginating one doesn't reset the other. The BDD scenario walks the cursor and asserts the second page returns a *different* file set, which catches "always emit the same opaque token" hardcoded-cursor bugs.
- **Working-tree mode handled gracefully.** When `head_ref` is omitted, the manifest half runs against the working tree and the function-context half returns an empty payload with stamped `budget_tokens` — no failure, no surprise. The `head_ref: "WORKTREE"` synthetic value matches the rest of the codebase's working-tree convention.
- **Tool-set drift impossible.** `EXPECTED_TOOLS` was bumped to five entries, the `it_matches_expected_tools` test enforces parity with the registered tools, the `it_registers_review_change_tool` test gates registration, and `it_publishes_comparative_git_diff_framing_in_review_change_schema` regression-protects the agent-facing pitch ("instead of git diff", "blast radius", "40/60"). The five-vs-four count was updated everywhere it appears (CLAUDE.md, README.md "Five MCP tools", `it_registers_exactly_five_tools`).

### Quality Gate Status

All hard gates green:

- **Tests:** 698 pass on macOS, Linux, Windows
- **Clippy:** clean with `--all-targets`
- **Format:** clean
- **BDD Acceptance:** 5/5 @ISSUE-240 scenarios pass
- **Integrity checks:** pass (cargo-tag-sync, no `|| true`, all clippy invocations use `--all-targets`, CHANGELOG sections match tags, MCP tool drift test enforced)
- **Mutation testing:** still pending — non-blocking for review since the unit tests are written specifically to resist obvious mutations (floor↔round swaps, complement vs independent computation)

Pre-gauntlet self-remediation captured every issue I would have raised:

- `is_any_file_truncated` naming consistency (was `any_truncated`)
- `record_functions_changed` metric was missing from the `review_change` handler
- `inspect_err` error attribution on both sub-tool calls (so `sub_tool` field on the error span identifies which half failed)
- `.gitignore` hardened with credential file patterns (`.env*`, `*.pem`, `*.key`, `*-credentials.json`, `id_rsa`/`id_ecdsa`/`id_ed25519`, etc.)
- BDD scenario title corrected (was "four tools", now "five")
- README documented `function_names` filter, `manifest_cursor` / `function_context_cursor`, `page_size` defaults
- CHANGELOG dash normalized to em dash
- Stale "four tools" shadow contract removed from BDD feature

### Inline Comments Posted

1. **Praise** — `split_budget` triangulation discipline (`src/tools/review_change.rs:122-130`)
2. **Praise** — `review_change_response_truncated` single source of truth (`src/server.rs:25-43`)
3. **Observation (non-blocking)** — consider an explicit `mode: "working_tree"` discriminator in `ContextMetadata` rather than relying on `head_ref == "WORKTREE"` for the empty-context case. Documented behavior is correct; this would just give downstream agents a clearer branch signal in a follow-up. (`src/tools/review_change.rs:289-320`)

### Blockers

None.

### Recommendations (non-blocking, follow-up only)

1. **Working-tree mode discriminator.** The `head_ref: "WORKTREE"` sentinel in the empty `ContextMetadata` is a stringly-typed enum. A future `mode: ContextMode { Committed, WorkingTree }` discriminator would let downstream consumers branch deterministically without parsing the `head_ref` value. Not blocking — the current shape matches the standalone tools.
2. **Integration test coverage for `build_review_change`.** The unit tests exercise `split_budget` and `ReviewChangeArgs` deserialization in detail, and the BDD scenarios exercise the end-to-end MCP path. There is no Rust-level integration test that pins the `build_review_change` function specifically (e.g., a test that builds a temp repo and asserts the budget stamping is observable on both halves). The BDD layer covers this, so it's complementary not duplicative.
3. **The `truncated: true` field name on `FunctionContextEntry` clashes slightly with the metadata-level `function_analysis_truncated` list** — both signal "this entry was clamped", but in different shapes. Pre-existing issue, not introduced by this PR.

### Praise — Specific Excellent Code

- **Block comment in `src/tools/review_change.rs:14-29`** that explains *why* the two budgets are computed independently rather than as complements is exactly the kind of WHY commenting that earns its keep. The comment cites the BDD scenario line by line number — when someone changes the spec in the future they will find this comment and know what they are breaking.
- **`it_publishes_comparative_git_diff_framing_in_review_change_schema`** — the test asserts not just that "git diff" appears in the schema, but that the load-bearing pitch words ("instead of", "blast radius", "40/60") are present. This is a regression test for marketing copy treated as a behavioral contract. Outstanding.
- **Sub-tool error attribution.** `inspect_err` with a `sub_tool = "get_change_manifest"` field on each underlying call means an operator looking at a failed `review_change` span sees *which half* failed without having to cross-reference the error message. That is the kind of observability detail that is invisible until production needs it and obvious-in-hindsight when production has it.
- **`it_registers_exactly_five_tools` was updated alongside `EXPECTED_TOOLS`.** The five-vs-four count is asserted in three different ways (literal constant, expected-list parity, length assertion) so accidental drift in any one path fails the test.

### Decision: APPROVED

This PR is ready to merge. The pre-gauntlet remediation eliminated every finding I would have flagged. Code quality, test discipline, observability, and documentation are all at the bar.

— pr-reviewer

mikelane (commented):
## Review verdict: **APPROVE** (submitted as COMMENT — GitHub blocks self-approval)

### Summary

This PR introduces `review_change` as a fifth MCP tool: a single-call combined `{ manifest, function_context }` payload that splits `max_response_tokens` 40/60 between the two halves. The implementation is **pure orchestration** — no diff or analysis logic is duplicated; both halves are produced by the existing standalone handlers via well-typed adapter calls. The 491-line `src/tools/review_change.rs` file is dominated by tests (~170 lines) and doc comments, with the actual orchestration logic running ~115 lines in one `build_review_change` function.

The token budget split is the load-bearing design decision, and it is **correctly triangulated** by two BDD examples (4096 → 1638/2458 and 16384 → 6553/9830) that together rule out every naive formula (uniform round, uniform floor, complement arithmetic). The commit history reflects an honest TDD trail: the BDD bootstrap (#244) preceded implementation, the `@not_implemented` tags were removed in a dedicated RED commit (`fd7262b`) before the GREEN feature commit, and the post-gauntlet church-purification commits (`737-714ddbb`) tightened metric attribution, debug tracing, scenario titles, formatting, and CHANGELOG hygiene.

### Verification (all pass)

| Check | Result |
|---|---|
| `cargo test --bin git-prism review_change` | 18/18 pass |
| `cargo test --bin git-prism budget` | 23/23 pass |
| `cargo fmt --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean (only pre-existing tree-sitter-kotlin C-scanner warnings, unrelated to this PR) |
| `python -m behave bdd/features/redirect_hook.feature --tags='@ISSUE-240'` | 5 scenarios, 24 steps, 0 failed |
| `it_matches_expected_tools` server contract | `review_change` registered, doc comment carries the load-bearing "instead of git diff" / "blast radius" / "40/60" framing tokens (asserted by `it_publishes_comparative_git_diff_framing_in_review_change_schema`) |

### Phase-by-phase findings

**Phase 0 — Bug identification.** I attempted to find bugs through code reading. Each suspicion was checked against the actual code paths; none confirmed:

1. *Could the budget split exceed B and over-allocate to sub-tools?* No — the `split_budget_manifest_plus_context_equals_or_near_budget` test parametrizes 6 values (100/999/4096/8192/16384/65536) and asserts `|B − m − c| ≤ 1`. Every case checks out.
2. *Could working-tree mode leak budget into the empty function_context's `token_estimate`?* No — `empty_function_context` (lines 289-320) sets `token_estimate: 0` and stamps `budget_tokens: Some(budget)` separately; no leak.
3. *Could the pagination metric double-count when both cursors are present?* No — the `if … is_some() || … is_some()` records once per call regardless of which cursor is active.
4. *Could `head_ref` consistency drift between worktree manifest and empty function_context?* No — `build_worktree_manifest` stamps `head_sha: "WORKTREE"` (line 600 of `manifest.rs`), and `empty_function_context` hardcodes the same. Verified consistent.

**Phase 1 — Quality tools.** All clean (above table).

**Phase 2 — Config integrity.** `.gitignore` adds 20 lines for credential and key file patterns — defense-in-depth, not exclusion of source. No checks suppressed.

**Phase 3 — Code standards.** `#[must_use]` on `split_budget`. Doc comments are thorough on every public symbol. Error handling propagates via `inspect_err` for telemetry without swallowing the error. Imports ordered correctly (std → external crates → crate-internal).

**Phase 4 — Self-documenting code.** Every comment I checked explains WHY (e.g., "Truncation for the smaller share + rounding for the larger share is the only formula that satisfies both BDD examples..."). No commented-out code. No TODO/FIXME without ticket refs. The triangulation rationale is captured directly in the test docstrings (e.g., "rules out: manifest = round(B * 0.4) — would yield 6554 here") so a future "simplification" attempt will hit a wall of documented prior art.

**Phase 5 — Test quality.** Tests test public API. No private method access. No `assert True`. Names are descriptive (`split_budget_at_4096_yields_1638_manifest_and_2458_context`, not "should_split_correctly"). Triangulation is intentional — the test names and bodies document which alternative formulas are being ruled out.

**Phase 6 — Architecture.** `build_review_change` is 119 lines (under bloater threshold). `review_change.rs` is 491 lines including tests/docs (under god-class threshold). SOLID intact: SRP for the orchestrator, no DIP violation since there's only one concrete sub-tool implementation per half. The `review_change_response_truncated` helper consolidates four truncation signals into one source of truth — exactly the right factoring.

**Phase 7 — Security/perf.** No new attack surface; uses the existing `repo_path` indirection pattern. The duplicate-manifest-build cost is honestly disclosed in the doc comment ("two manifest builds happen per call — once visible to the user, once internal to the context tool"). Acceptable for a v1, candidate for follow-up optimization.

### Recommendations (LOW severity, none blocking)

1. **`empty_function_context` page_size drift** (`src/tools/review_change.rs:316`) — synthetic empty response uses `page_size: 0`, while the standalone `build_function_context_with_options` returns the caller's requested `page_size` even on empty results. Mirroring `clamp_page_size(args.page_size)` here would preserve the contract that `page_size` reflects intent. Filed as inline comment.

2. **Duplicate manifest build per call** (`src/tools/review_change.rs:144-153`) — the function-context handler reads the manifest internally, so `review_change` walks the diff twice. Documented honestly. Worth a follow-up issue to extract a shared `compute_changed_functions(repo_path, base_ref, head_ref)` step that both halves consume. Not a v1 blocker. Filed as inline comment.

### Praise

- The triangulation rationale in the module-level doc comment (`src/tools/review_change.rs:25-28`) explicitly captures *why* complement arithmetic was rejected. That's the kind of "why" that survives a year of refactors.
- The `review_change_response_truncated` helper (`src/server.rs:30-43`) consolidates four truncation signals into one source of truth, with a forward-looking doc comment that anticipates a third source case.
- The `EXPECTED_TOOLS` test contract (`src/server.rs:890-896`) catches doc/code drift at compile time — this PR correctly extends the constant to include `review_change`.
- BDD scenarios use parametrized Examples with two budget values that together rule out every naive split formula. This is the gold-standard form of scenario triangulation.

### Decision

**APPROVED.** The implementation matches its design, the tests rule out gaming via triangulation, the gauntlet is clean, and the two LOW recommendations are non-blocking refactor candidates. The two LOW inline comments document follow-up opportunities, not blockers.

Submitted as COMMENT because GitHub prohibits self-approval. The blessing system at `~/.claude/state/pr-blessings/` should be written by the merge gauntlet flow on the user's instruction.

PR #248, opened 2026-04-29
feat(hooks): Python bash tokenizer and redirect hook script

### LINKED ISSUE #238: feat(hooks): bash command tokenizer for redirect hook
## Goal

Implement the bash-command tokenizer that the bundled redirect hook calls. Stdlib-only Python, no third-party dependencies. Returns a structured list of `git` invocations found anywhere in the parsed command string.

Parent epic: #234. Depends on W0 (parser library decision in ADR `0008`) and W1 (BDD scenarios).

## Deliverable

- New file: `hooks/parse_git_invocations.py`
- New file: `hooks/test_parse_git_invocations.py` (pytest)

## Behavior

Input: a bash command string (the contents of the `tool_input.command` field that Claude Code's `Bash` PreToolUse hook receives).

Output: JSON list to stdout, one entry per `git` invocation found:

```json
[
  {"subcommand": "diff", "args": ["main..HEAD"], "position": 12, "raw": "git diff main..HEAD"},
  {"subcommand": "log",  "args": ["--oneline", "main..HEAD"], "position": 38, "raw": "git log --oneline main..HEAD"}
]
```

The hook script reads this JSON and decides whether to redirect based on `subcommand` (only `diff`, `log`, `show`, `blame` trigger redirects).

## Cases the Tokenizer Must Handle

- Plain command: `git diff main..HEAD`
- Compound with `&&` / `||` / `;`: `cd /tmp && git diff main..HEAD`
- Subshells: `(git log main..HEAD)`
- Command substitution: `BASE=$(git rev-parse main); git diff $BASE..HEAD` — recognize the inner `git rev-parse` and the outer `git diff`
- Variable expansion in args: `git diff $BASE..HEAD` — recognize as `git diff` with args including `$BASE..HEAD` literal (don't expand)
- Heredocs: `cat <<EOF\nfoo\nEOF` — skip embedded text; if `git diff` follows the heredoc, recognize it
- Pipelines: `git log main..HEAD | head -10` — recognize `git log`
- Quoted strings: `git diff "main..HEAD"` — strip quotes from args
- Backslash continuations: `git diff \\\n  main..HEAD` — join lines

## Cases the Tokenizer Must NOT False-Positive On

- A literal string in another command: `echo "git diff main..HEAD"` — recognize that this is just an `echo`, the inner `git diff` is data
- Comments: `# git diff main..HEAD` — skip
- Shell function names that contain "git": `git_helper foo` is not a `git` invocation

## Implementation Approach

Per the ADR (W0), the chosen parser is `shlex` + a small recursive walker for compound structures. Pseudocode:

```python
def parse_git_invocations(command: str) -> list[dict]:
    # 1. Split into pipelines/compound statements at top level (&&, ||, ;, |)
    # 2. For each segment, tokenize with shlex (posix=True, punctuation_chars=True)
    # 3. Detect subshells / command substitution and recurse into them
    # 4. For each token list, check if first token is "git"
    # 5. If so, emit { subcommand, args, position, raw }
    ...
```

## Tests

`hooks/test_parse_git_invocations.py` — pytest table-driven tests, one parameter set per case above. At least:

- 8 positive cases (each compound/subshell/etc form)
- 5 negative cases (echo, comment, function name, etc.)
- 3 edge cases (empty input, malformed bash, very long input)

## Definition of Done

- [ ] `hooks/parse_git_invocations.py` reads command string from stdin, writes JSON list to stdout
- [ ] No third-party Python dependencies (stdlib `shlex`, `re`, `json` only)
- [ ] All cases above have passing pytest tests
- [ ] `pytest hooks/` runs green
- [ ] All `@ISSUE-<this>` BDD scenarios pass

## Notes

The tokenizer has no domain knowledge — it doesn't know which subcommands trigger redirects. That decision lives in the bash hook script (W4). Keeping the parser dumb means it's easy to test and unlikely to grow false-positive matches over time.

mikelane:
## Summary

- Adds `hooks/bash_redirect_hook.py`: stdlib-only Python tokenizer that intercepts bash commands and advises/blocks when a git-prism MCP tool would be more appropriate
- Adds `hooks/git-prism-redirect.sh`: thin bash wrapper with `#!/bin/bash` shebang and a `python3` fail-open guard
- Tokenizer uses `shlex` with `posix=True, punctuation_chars=True`, backtick normalization pre-pass, and heredoc body skipping
- Intercepts: `git diff <ref>..<ref>` → advise `get_change_manifest`; `git log <ref>..<ref>` → advise `get_commit_history`; `gh pr diff` → block (exit 2); `mcp__github__get_commit` → block (exit 2); common plumbing commands (`status`, `add`, `commit`, `push`) → silent pass-through

## Test plan

- [x] `python -m behave bdd/features/redirect_hook.feature --tags="@ISSUE-238"` — 17 scenarios pass
- [x] `cargo test --bin git-prism` — 680 tests pass

Closes #238

### DIFF
--- a/hooks/bash_redirect_hook.py
+++ b/hooks/bash_redirect_hook.py
@@ -1,0 +1,573 @@
+#!/usr/bin/env python3
+"""Bundled redirect hook for git-prism.
+
+Reads a Claude Code PreToolUse JSON payload on stdin and decides whether
+the bash command (or MCP tool name) should be redirected to a git-prism
+MCP tool. The decision is delivered through three exit-code states:
+
+    exit 0, empty stdout/stderr           -> no redirect (silent allow)
+    exit 0, JSON on stdout                -> advisory redirect (allow + nudge)
+    exit 2, redirect text on stderr       -> hard block (deny + explain)
+
+The hook MUST be hermetic and stdlib-only (per ADR-0008): no third-party
+imports, no shelling out for parsing, no environment-variable expansion.
+
+The tokenizer uses ``shlex.shlex(posix=True, punctuation_chars=True)``
+with two wrappers:
+
+  * Heredoc body skipping: ``<<TAG`` (and ``<<-TAG`` / ``<<'TAG'`` /
+    ``<<"TAG"``) trigger a state machine that drops every token until the
+    closing ``TAG`` line. Inline ``TAG`` text inside the body is ignored
+    because we require a preceding newline token.
+
+  * Backtick normalization: stray backticks are converted to whitespace
+    in a pre-pass so ``cmd`` substitutions split cleanly into candidate
+    commands.
+
+After tokenization, the flat list is split at ``&&``, ``||``, ``|``,
+``;``, ``(``, ``)`` to produce candidate commands. Each candidate is run
+through the watch-list matchers; the first match wins.
+"""
+
+from __future__ import annotations
+
+import io
+import json
+import shlex
+import sys
+from typing import Iterable
+
+# Tokens that mark a command boundary inside the flat shlex output. After
+# tokenization we split the list at these tokens and treat each slice as
+# an independent candidate command for the matcher.
+COMMAND_SEPARATORS = frozenset({"&&", "||", "|", ";", "(", ")", "\n"})
+
+
+def _strip_backticks(command: str) -> str:
+    """Replace every backtick with a space.
+
+    ADR-0008 calls this out explicitly: backticks are command-substitution
+    delimiters identical to ``$(...)``. Our matcher already splits on
+    parentheses, so converting backticks to whitespace lets the same code
+    path handle both substitution forms with no extra state.
+    """
+    return command.replace("`", " ")
+
+
+_NEWLINE_MARKER = "\n"
+
+
+def _tokenize_line(line: str) -> list[str]:
+    """Tokenize a single line of bash with shlex.
+
+    A line is everything between two newline characters in the original
+    command. Each line is lexed independently so a heredoc body whose
+    contents would confuse a single-pass lexer (mismatched quotes that
+    only close inside the body, for instance) cannot poison the parse
+    of the surrounding code. On lex failure the offending line is
+    represented by an empty list — the heredoc walker still sees the
+    surrounding ``\\n`` markers and the matcher safely no-ops on the
+    blank slice.
+    """
+    if not line.strip():
+        return []
+    lex = shlex.shlex(line, posix=True, punctuation_chars=True)
+    lex.whitespace_split = True
+    try:
+        return list(lex)
+    except ValueError:
+        return []
+
+
+def _heredoc_tag(token: str) -> tuple[str, bool] | None:
+    """Recognize the heredoc tag word that follows the ``<<`` operator.
+
+    Returns ``(tag, is_dash_form)`` for valid tag tokens. The dash form
+    (``<<-``) tokenizes as ``<<`` + ``-EOF`` because shlex glues the
+    leading dash onto the next word. We strip that dash to recover the
+    real tag. Returns ``None`` when the token cannot be a tag (empty,
+    starts with another shell-special character).
+    """
+    if not token:
+        return None
+    is_dash = False
+    tag = token
+    if tag.startswith("-"):
+        is_dash = True
+        tag = tag[1:]
+    # Bash also allows ``<<'EOF'`` / ``<<"EOF"`` to suppress expansion in
+    # the body. Shlex usually strips those quotes, but if the lexer left
+    # them in (e.g. a different lex mode) we still want the bare word.
+    if len(tag) >= 2 and tag[0] == tag[-1] and tag[0] in ("'", '"'):
+        tag = tag[1:-1]
+    if not tag:
+        return None
+    return tag, is_dash
+
+
+def _tokenize_raw(command: str) -> list[str]:
+    """Produce a flat token list with explicit ``\\n`` separator tokens.
+
+    Newline tokens act as line boundaries downstream: the heredoc walker
+    uses them to detect the closing tag (``\\n TAG \\n``), and the
+    candidate-command splitter treats them as command separators.
+    """
+    cleaned = _strip_backticks(command)
+    if not cleaned:
+        return []
+    tokens: list[str] = []
+    lines = cleaned.split("\n")
+    for index, line in enumerate(lines):
+        tokens.extend(_tokenize_line(line))
+        if index != len(lines) - 1:
+            tokens.append(_NEWLINE_MARKER)
+    return tokens
+
+
+def _drop_heredoc_bodies(tokens: list[str]) -> list[str]:
+    """Walk the token list and drop heredoc body tokens.
+
+    Bash heredoc semantics: ``<<TAG`` consumes every line until a line
+    whose only content is ``TAG``. The token stream from ``_tokenize_raw``
+    surfaces line boundaries as explicit ``\\n`` tokens, so the closing
+    line appears as ``\\n``, ``TAG``, ``\\n`` (or ``\\n``, ``TAG`` at
+    end-of-stream). We walk linearly and drop body tokens until that
+    sequence appears.
+
+    Inline ``TAG`` text inside the body is preserved as opaque content
+    because it is not surrounded by newline tokens in this stream.
+    """
+    output: list[str] = []
+    index = 0
+    while index < len(tokens):
+        tok = tokens[index]
+        if tok == "<<":
+            tag_info = (
+                _heredoc_tag(tokens[index + 1]) if index + 1 < len(tokens) else None
+            )
+            if tag_info is None:
+                # ``<<`` at end of input — drop the operator and stop.
+                break
+            tag, _is_dash = tag_info
+            # Step past ``<<`` and the tag word; both are dropped.
+            index += 2
+            # Drop everything on the same line as the operator (it's
+            # part of the heredoc opening, e.g. ``cat <<EOF; echo hi``
+            # has ``; echo hi`` after the tag — bash treats it as the
+            # rest of the command line, not the heredoc body).
+            while index < len(tokens) and tokens[index] != _NEWLINE_MARKER:
+                index += 1
+            # Now consume body lines until we find a line that contains
+            # only the closing tag.
+            while index < len(tokens):
+                if tokens[index] != _NEWLINE_MARKER:
+                    # Inside body content — keep walking.
+                    index += 1
+                    continue
+                # Just consumed a ``\n``. Peek at the next line.
+                index += 1
+                if index >= len(tokens):
+                    break
+                # If this line is exactly the closing tag (i.e., the
+                # token immediately after the newline equals the tag and
+                # is followed by another newline OR end-of-stream), we
+                # have hit the terminator.
+                if tokens[index] == tag:
+                    next_index = index + 1
+                    if (
+                        next_index >= len(tokens)
+                        or tokens[next_index] == _NEWLINE_MARKER
+                    ):
+                        # Skip past the tag itself; the trailing newline
+                        # is left for the outer walker so the next
+                        # command starts cleanly delimited.
+                        index += 1
+                        break
+                # Otherwise continue scanning the body.
+            continue
+        output.append(tok)
+        index += 1
+    return output
+
+
+def tokenize_command(command: str) -> list[list[str]]:
+    """Tokenize a bash command into a list of candidate command token lists.
+
+    The outer list represents pipeline / compound boundaries; each inner
+    list is the tokens of a single candidate command. Empty inner lists
+    are dropped — for example, ``(git status)`` produces ``['git',
+    'status']`` after the ``(`` and ``)`` are consumed as separators.
+    """
+    flat = _tokenize_raw(command)
+    if not flat:
+        return []
+
+    flat = _drop_heredoc_bodies(flat)
+
+    candidates: list[list[str]] = []
+    current: list[str] = []
+    for tok in flat:
+        if tok in COMMAND_SEPARATORS:
+            if current:
+                candidates.append(current)
+                current = []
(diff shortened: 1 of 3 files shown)


mikelane:
## PR Review — feat(hooks): Python bash tokenizer and redirect hook script

**Verdict: APPROVE** — No CRITICAL or HIGH blockers. Two LOW observations noted below.

---

### Phase 1 — CI Status

All critical checks pass: Format ✓, Clippy ✓, Test (ubuntu) ✓, Test (macos) ✓, Check ✓, Integrity ✓. Windows test and BDD Acceptance still running at review time.

---

### Phase 2 — Bug Hunting

**Finding B1 (LOW): Double tokenization on non-matching commands in `_decide_bash`**

`_matches_gh_pr_diff` tokenizes the full command string, then `_decide_bash` calls `tokenize_command` again on the same input. For commands that don't match, this is wasted work. Not a correctness bug — both paths produce equivalent results. Negligible in practice (hooks run on short bash snippets).

Location: `hooks/bash_redirect_hook.py`, `_matches_gh_pr_diff` + `_decide_bash`.

**Finding B2 (LOW): `ValueError` branch in `_advice_for_tool` is unreachable**

`_advice_for_tool` catches `ValueError` from a `.index()` call, but `TOOL_WATCH_LIST` is a compile-time constant and all callers pass valid keys. The branch is dead code. Not harmful, but could mislead a future reader.

Location: `hooks/bash_redirect_hook.py:_advice_for_tool`.

No other bugs found. Tokenizer edge cases (heredoc bodies, backtick normalization, `ValueError` from shlex on malformed input, fail-open shell wrapper) are all handled correctly and intentionally per ADR-0008.

---

### Phase 3 — Code Quality

Clean throughout.

- `Decision` with `__slots__` and `SILENT` module-level singleton is idiomatic.
- `_drop_heredoc_bodies` correctly handles `<<EOF` / `<<'EOF'` / `<<"EOF"` variants.
- Backtick normalization pre-pass before handing to shlex is the right approach.
- `_read_payload` is properly fail-open on malformed JSON and missing keys.
- stdlib-only constraint satisfied.
- Shell wrapper: `set -u` without `set -e` correctly implements the fail-open design.

---

### Phase 4 — Security

No OWASP Top 10 issues. No subprocess calls, no `eval`, no shell construction from user data. `shlex.split` is wrapped for `ValueError`. JSON parsing is wrapped for `JSONDecodeError`/`KeyError`. `SCRIPT_DIR` path construction uses `"${BASH_SOURCE[0]}"` with proper quoting. `shlex.split` on attacker-controlled input is safe — it parses, not executes.

---

### Phase 5 — Adversarial

- **Gaming the advisor**: Wrapping git in an alias or function would bypass detection — by design (fail-open, opt-in). No security property depends on exhaustive detection.
- **Silent regressions**: `review_change` not in watch list — intentional, that's issue #240's concern.
- **Blast radius**: Hook-only change, no production Rust touched, no MCP schema changes.

---

### Summary Table

| # | Severity | Finding |
|---|----------|---------|
| B1 | LOW | Double tokenization on non-matching commands (performance only) |
| B2 | LOW | Unreachable `ValueError` branch in `_advice_for_tool` |

No CRITICAL, no HIGH. Both LOWs are cosmetic — neither affects correctness, security, or user-facing behavior. **Ready to merge.**

---

*Reviewed by pr-reviewer subagent (solo-dev self-review). Generated with [Claude Code](https://claude.ai/claude-code)*

hooks/bash_redirect_hook.py
```diff
@@ -0,0 +1,573 @@
...
+        return "get_function_context"
+    if sub == "diff" and _has_ref_range(rest):
+        return "get_change_manifest"
+    if sub == "log" and _has_ref_range(rest):
+        return "get_commit_history"
+    if sub == "blame":
+        return "get_file_snapshots"
+    if sub == "show":
+        return "get_file_snapshots"
+    return None
```
mikelane, line 337: Question for the maintainer's future self: should this watch list eventually grow to cover `git show <sha>` (which is currently caught only via the bare `git show` rule, with no ref-range guard) and `git log --grep` (which is text-search, distinct from `-S`/`-G` pickaxe but still has a structured-data alternative via `get_commit_history`)? Not a blocker — the present scope matches ADR-0008's coverage matrix exactly — but worth a follow-up issue if the heuristic is ever extended.

hooks/bash_redirect_hook.py
```diff
@@ -0,0 +1,573 @@
...
+
+    Matches both ``a..b`` (two-dot, all merges in either direction) and
+    ``a...b`` (three-dot, symmetric difference). A bare ``..`` token is
+    excluded because it is the parent-directory shorthand, not a ref
+    range.
+    """
+    for tok in tokens:
+        if ".." in tok and tok not in ("..", "..."):
+            return True
+    return False
```
mikelane, line 292: Minor observation on the `_has_ref_range` heuristic: a token like `path/foo..bar` (relative pathspec with `..` in a filename) would also match. Probed locally with `git diff path/foo..bar` and the hook does emit advisory text — the redirect is non-blocking so the false positive is harmless, and the literal echo (`You ran: git diff path/foo..bar`) makes the misfire transparent to the agent. The simpler heuristic is the right call here; a tighter ref-grammar regex would buy precision at a real maintenance cost. Worth recording as a known false-positive class in case it surfaces in user reports later.

hooks/bash_redirect_hook.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Read the hook payload from stdin and dispatch a Decision.
+
+    Returns the exit code (0 for silent / advise, 2 for block). The
+    function never raises; any unexpected error is converted into a
+    fail-open silent allow with a single-line stderr warning.
+    """
+    _ = argv  # CLI args reserved for future flags; ignored today.
```
mikelane, line 547: Worth pausing on this `argv` parameter: the entry point accepts it but the body assigns `_ = argv` to silence the unused warning and notes "reserved for future flags". Strict YAGNI would say drop the parameter and reintroduce it when an actual flag arrives. Keeping it has one defensible reason — testability: `main([])` can be called directly from a unit test without monkey-patching `sys.argv`. If the BDD subprocess shell-out is genuinely the only test surface (which seems to be the current state with no `hooks/tests/test_bash_redirect_hook.py` yet), the `argv` parameter is unreachable code today. ADR-0008 mentions a future pytest suite — once that lands and exercises `main([])`, the parameter becomes load-bearing. Until then, a one-line note like `# argv kept for future pytest suite, see ADR-0008` would make the intent explicit.

hooks/bash_redirect_hook.py
```diff
@@ -0,0 +1,573 @@
...
+        self,
+        mode: str,
+        advice: str = "",
+        message: str = "",
+        tool_name: str = "",
+    ):
+        self.mode = mode
+        self.advice = advice
+        self.message = message
+        self.tool_name = tool_name
```
mikelane, line 385: One observation on the `Decision` class shape: with `mode`, `advice`, `message`, and `tool_name` all as plain attributes plus `__slots__`, this is functionally a tagged union but expressed as a flat record where most fields are ignored in any given mode (`advise` ignores `message`; `block` ignores `advice`). A frozen dataclass would document the same surface with less ceremony, and a literal `mode: Literal['silent', 'advise', 'block']` annotation would let mypy catch typos at the construction sites. Not a blocker — the BDD layer would catch a mode-string typo via behavior — but it's exactly the kind of place where stdlib ergonomics now reach further than they did when this style was idiomatic.

hooks/bash_redirect_hook.py
```diff
@@ -0,0 +1,573 @@
...
+    candidates = tokenize_command(command)
+    for tokens in candidates:
+        if (
+            len(tokens) >= 3
+            and tokens[0] == "gh"
+            and tokens[1] == "pr"
+            and tokens[2] == "diff"
+        ):
+            return True
+    return False
```
mikelane, line 496: Excellent piece of defense-in-depth: re-tokenizing in `_matches_gh_pr_diff` is redundant with the candidate walk in `_decide_bash`, but it's intentional redundancy — `gh pr diff` must be exit 2 (block), and routing it through the advisory candidate-walk would have produced exit 0 (advise). The dedicated check before the candidate loop guarantees the right exit code regardless of how the candidate matcher is later modified. The doc comment captures this clearly. Worth keeping as-is.

mikelane (commented):
## Adversarial review summary

**Decision: APPROVED (recorded as COMMENT — owner-authored PR cannot be formally approved by GitHub).**

This is exemplary work: 573 lines of stdlib-only Python that does one job well, with documentation that earns its keep on every line. The contract under test (exit code, stdout JSON shape, stderr text given a stdin payload) is the load-bearing surface, and it is exercised end-to-end by 17 BDD scenarios that hit every shape from ADR-0008's coverage matrix.

### What I probed

Beyond the 17 BDD scenarios already passing, ran 14 ad-hoc invocations of the hook against shapes the BDD suite does not directly cover, in order to find regressions or false positives:

| Probe | Expected | Result |
|---|---|---|
| `git diff main..HEAD` | advise | advise |
| `git diff $BASE..HEAD` with `BASE=SECRETSENTINEL` | advise without expansion, no leak | confirmed literal `$BASE` in advice, no `SECRETSENTINEL` anywhere |
| Empty stdin | exit 0 silent | confirmed |
| Malformed JSON | exit 0 + single stderr line | confirmed |
| Heredoc with bait inside body (`heredoc_then_git_diff.txt`) | advise on post-EOF `git diff`, no advise for body's `git log` | confirmed |
| Tab-stripped `<<-EOF` heredoc with tab-indented closer | silent (only `git status` after, not on watch list) | confirmed |
| `git diff` (no range) | silent | silent |
| `git log -p` | silent | silent (no pickaxe, no range) |
| `git log -S foo` and `git log -Sneedle` | advise pickaxe | advise pickaxe |
| `git log --grep=foo` | silent | silent (correct — pickaxe is `-S/-G`, not `--grep`) |
| `git stash show` | silent | silent (no `..` in tokens) |
| Unclosed quote `git diff main..HEAD "unclosed` | silent / fail-open | silent (lex ValueError caught, line dropped) |
| `cd $(git rev-parse --show-toplevel) && git diff main..HEAD` | advise on outer `git diff`, no advise for `git rev-parse` | confirmed: `$(...)` splits into separate candidates, inner `git rev-parse` is not on the watch list |
| `cd \`git rev-parse\` && git diff main..HEAD` | same as above with backticks | confirmed: backticks normalized to whitespace pre-pass |
| `git diff path/foo..bar` (path with `..` in name) | false-positive advise | confirmed false positive — see inline comment |

The path-`..`-in-name case is the only false positive I found, and it's documented inline as a known acceptable limit of the heuristic. The redirect is non-blocking (advisory, exit 0) and the literal echo (`You ran: git diff path/foo..bar`) makes the misfire transparent to the agent.

### What earns explicit praise

- **`_matches_gh_pr_diff` defense-in-depth.** Re-tokenizing once before the candidate loop guarantees `gh pr diff` produces exit 2 regardless of how the advisory matcher evolves later. Documented inline.
- **Heredoc walker contract.** The `_drop_heredoc_bodies` state machine handles `<<TAG`, `<<-TAG`, `<<'TAG'`, `<<"TAG"`, AND resumption-after-terminator in <60 lines, with comments that explain bash semantics rather than restating Python operations.
- **Per-line shlex with try/except.** Catches `ValueError` from unmatched quotes and degrades to a silent line drop. The fail-open contract is preserved without leaking parser errors to the agent.
- **Literal-echo of the user command in advice.** Doubles as a security assertion (proves `$BASE` was not expanded) and a UX win (the agent sees `You ran: X — try Y instead`).
- **Absolute shebang `#!/bin/bash`.** Documented in the wrapper script with the correct rationale (so the missing-`python3` test cannot fail for the wrong reason — env-form shebang would also fail PATH lookup).
- **ADR-0008 written first.** Every architectural choice in the code maps back to a numbered decision in the ADR, including the rationale for `--scope user` default, the `id` sentinel, and the fail-open posture.

### Inline comments (none blocking)

Posted four threads on `hooks/bash_redirect_hook.py`:

1. `_classify_git_command` extensibility — should the watch list eventually grow to cover `git show <sha>` ref-range form and `git log --grep`?
2. `_has_ref_range` false-positive class (paths containing `..`) — record as known limitation
3. `argv` parameter YAGNI tension — testability rationale exists, worth annotating the parameter
4. `Decision` class shape — could be a frozen dataclass with `Literal` mode tag

### What I checked and found no issue with

- **Pre-commit / linter / type checker bypasses:** None. No `# type: ignore`, no `# noqa`, no `mypy: ignore-errors`. The Python module uses `from __future__ import annotations` and is mypy-clean.
- **Tests-via-private-attribute:** None. BDD steps shell out to the real binary; only public functions (`tokenize_command`, `decide_redirect`, `main`) are observable.
- **Tests inflating coverage:** No assertions like `assert True`, no parametrize loops over identical bodies.
- **TODO/FIXME without ticket:** None found.
- **Commented-out code:** None.
- **Security:** No subprocess invocation in the parser, no `os.path.expandvars`, no `eval`/`exec`. The `shlex` lex is fed string-only input. No env-var expansion. Confirmed via the `SECRETSENTINEL` probe.
- **Regression risk:** Hook is purely additive (`hooks/` is a new directory, `bash_redirect_hook.py` and `git-prism-redirect.sh` are new files). The only modification to existing code is removing `@not_implemented` from 13 BDD scenarios, which is the prescribed RED-to-GREEN tag flip from the BDD framework rule. Revert is delete-two-files clean.
- **CI:** All 14 reportable checks `SUCCESS` (BDD Acceptance, Test on ubuntu/macos/windows, Clippy, Format, Integrity checks). `mergeStateStatus: CLEAN`.

### Decision

**APPROVED — no blockers, no critical or high findings.** The four inline comments are quality observations for follow-up consideration only. Recorded as `COMMENT` event because GitHub does not allow repo owners to formally approve their own PRs (per CLAUDE.md solo-developer policy).

PR #273, opened 2026-05-12
feat: add diff_hunks to get_file_snapshots for diff-relative line mapping

### LINKED ISSUE #271: `get_file_snapshots` lacks diff-relative line numbers needed for GitHub inline review comments
When using git-prism as the source of truth for PR review tooling (e.g., when `gh pr diff` is blocked in favor of structured output), `get_file_snapshots` returns complete before/after file content but no information about which lines fall within the actual diff hunk. GitHub's Pull Request Review API requires diff-relative line positions — not file-absolute line numbers — for inline comments. A file-absolute line number that's outside the diff hunk returns a 422.

**What's missing:** A way to know, for a given line in the after (or before) file, whether it's part of the diff and what its diff-relative position is.

**Possible solutions (in rough preference order):**

1. Add a `diff_hunks` field to `get_file_snapshots` response — each hunk with `old_start`, `old_lines`, `new_start`, `new_lines`, and the lines it covers, mirroring the unified diff `@@` header. Consumers can then map file line → hunk position themselves.
2. Add a `github_line_positions` map to the response — keyed by file line number, value is the GitHub-compatible `position` (sequential offset from the first `@@` in the diff). This is the most directly useful for GitHub API consumers.
3. A separate lightweight tool — `get_diff_hunks(base_ref, head_ref, paths)` — that returns just the hunk metadata without full file content, for callers that only need position mapping.

**Context:** Hit this while building a structured PR review agent. The agent uses git-prism for all diff analysis, but falls back to putting review comments in the PR body rather than inline because hunk positions can't be derived from file snapshots alone.

mikelane:
## Summary

Adds `diff_hunks` field to `get_file_snapshots` response — structured diff hunk metadata (`old_start`, `old_lines`, `new_start`, `new_lines`) mirroring unified diff `@@` headers. This lets consumers compute diff-relative line positions needed for GitHub inline review comments, GitLab MR discussions, and other review tooling.

Opt-in via `include_diff_hunks: bool` parameter (default `false`).

## Design

- `DiffHunk` struct in `src/tools/types.rs` with 1-based line numbers
- `extract_hunks()` in `src/git/diff.rs` using `gix::diff::blob` (Myers algorithm)
- `compute_diff_hunks()` in `src/tools/snapshots.rs` wired into `build_snapshot_entry`
- `--include-diff-hunks` CLI flag on `Snapshot` subcommand
- Only emitted for modified files with both sides present

## Commits
```
738994b fix: compute hunks for empty-file edges and add CLI flag
fc5a411 feat: add diff_hunks to get_file_snapshots for diff-relative line mapping
e20c3b2 docs: document include_diff_hunks parameter in CLAUDE.md and README
```

## Test plan
- [x] 799 Rust tests pass
- [x] 8 adversarial QA Python tests pass
- [x] Clippy clean, fmt clean
- [x] Church purification: rust-purist, test-purist, naming-purist
- [x] Adversarial QA: 4 bugs found, all fixed

Closes #271


src/git/diff.rs
```diff
@@ -257,6 +257,49 @@ pub(crate) fn count_lines(data: &[u8]) -> usize {
...
+}
+
+/// Extract unified diff hunk boundaries from old and new file content.
+///
+/// Uses `gix::diff::blob` to compute a line-level diff and extracts the hunk
+/// ranges. The returned boundaries are 1-based line numbers suitable for
+/// consumers that need to compute diff-relative positions (e.g., GitHub inline
+/// review comments).
+///
+/// Returns `None` if either input is empty or the content is binary.
```
mikelane, line 276: Doc comment stale: says Returns None but function returns Vec; says empty inputs cause return but empty inputs now flow through after fix commit 1d5dbc2.

PR #276, opened 2026-05-13
fix: surface review_change in bash-redirect hook messages, fix unicode encoding

### LINKED ISSUE #265: docs: surface review_change in bash-redirect hook message for PR review workflows
## Problem

The bash-redirect hook intercepts `gh pr diff` and similar commands and redirects agents to git-prism — but the redirect message only mentions `get_change_manifest`. Agents default to the familiar two-step (get_change_manifest → get_file_snapshots) instead of calling `review_change`, which combines both in a single call with a shared token budget.

Observed: an agent redirected by the hook used get_change_manifest + get_file_snapshots separately and only noted afterward that `review_change` "would have replaced my two separate calls in one shot." The agent defaulted to the two-step because it was what it knew — not because review_change is hidden, but because the redirect message never surfaced it.

**Why now:** review_change exists but agents keep reaching for the two-step instead. The hook redirect is the right place to fix this — it's where agents are already being told "use git-prism instead," and adding one more sentence costs nothing.

## BDD Bootstrap Note

Before implementation starts, these acceptance scenarios will be added to `bdd/features/redirect_hook.feature` tagged `@ISSUE-<N> @not_implemented`. The first commit on the implementation branch removes `@not_implemented` (proving RED). The Gherkin below is the specification that becomes those `.feature` file entries.

## Acceptance Criteria

```gherkin
# BLOCK_GH_PR_DIFF path — hard-block, exit 2, stderr
Given an agent that intercepts a "gh pr diff <N>" command via the bash-redirect hook
When the hook fires and redirects to git-prism
Then the hook exit code is 2
And the hook stderr contains "review_change"
And the hook stderr contains "get_change_manifest"

# ADVICE_GET_CHANGE_MANIFEST path — advisory, exit 0, stdout JSON
Given an agent that intercepts a "git diff main..HEAD" style command
When the hook fires
Then the hook exit code is 0
And the hook stdout JSON advice contains "review_change"
And the hook stdout JSON advice contains "get_change_manifest"
```

## Out of Scope

- Changing the hook behavior (it should still redirect, not call the tool automatically)
- Forcing agents to use review_change (guidance only, not enforcement)
- Changes to the review_change tool signature or schema

## Implementation Notes

The redirect messages live in `hooks/bash_redirect_hook.py`. Two constants need updating:

**`BLOCK_GH_PR_DIFF`** — used for `gh pr diff` hard-block (exit code 2). Currently mentions only `get_change_manifest`. Add `review_change` alongside it. The existing BDD scenario `@ISSUE-239` in `bdd/features/redirect_hook.feature` asserts `"get_change_manifest"` is present in stderr — the updated message must contain **both** strings; do not replace one with the other or `@ISSUE-239` will fail.

**`ADVICE_GET_CHANGE_MANIFEST`** — used for `git diff` advisory redirects (exit code 0, non-blocking). A separate code path from `BLOCK_GH_PR_DIFF`. Add `review_change` here as well, but with care: this path is advisory-only and its blast radius is lower. Do not alter exit codes or blocking behavior when editing this constant.

Suggested message pattern for `BLOCK_GH_PR_DIFF`:
```
git-prism: gh pr diff returns raw text. Use git-prism instead:
  review_change(repo_path, base_ref, head_ref)       ← full PR review (manifest + function context)
  get_change_manifest(repo_path, base_ref, head_ref)  ← quick file-level scan
```

## Dependencies

Companion issue: "detect missing local branches and guide ref resolution for remote PRs" (being filed simultaneously). These are parallel fixes — neither blocks the other — but they address the same friction point and ideally ship together. After both issues are filed, the relationship will be confirmed via the GitHub Dependencies API.

## Edge Cases

- Hook fires for `git log` or `git show`: updated message still visible but less targeted — acceptable; it is short
- Agents that cached the old hook output get the new message after `git-prism hooks install` re-run — no migration needed

mikelane:
## Summary

Three changes to the bash-redirect hook:

1. **Surface `review_change`** in hook messages — when agents attempt to bypass via `gh pr diff`, the hook now recommends `review_change` as the preferred alternative
2. **Fix unicode encoding** — replace em-dashes and unicode arrows with ASCII-safe characters to prevent `UnicodeEncodeError` in ASCII-only locales
3. **Fix escaped newline handling** — handle bash line continuation (`\<newline>`) before tokenization to prevent false-positive hard blocks
4. **Fix gh api contents matching** — use tokenizer pipeline instead of raw regex to avoid matching inside quoted echo arguments

Closes #265

### DIFF
--- a/hooks/test_bash_redirect_hook.py
+++ b/hooks/test_bash_redirect_hook.py
@@ -21,6 +21,7 @@ sys.path.insert(0, str(Path(__file__).parent))
 
 from bash_redirect_hook import (
     BLOCK_GH_PR_DIFF,
+    BLOCK_MCP_GITHUB_GET_COMMIT,
     _advice_with_echo,
     _matches_gh_api_contents,
     _classify_git_command,
@@ -30,7 +31,9 @@ from bash_redirect_hook import (
     _is_functionally_empty,
     _matches_gh_pr_diff,
     _emit_advice,
+    _read_payload,
     decide_redirect,
+    main,
     tokenize_command,
 )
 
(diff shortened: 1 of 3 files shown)


hooks/test_bash_redirect_hook.py
```diff
@@ -600,5 +605,136 @@ def test_git_diff_after_heredoc_is_advised(self) -> None:
         self.assertIn("get_change_manifest", decision.advice)
 
 
+# ---------------------------------------------------------------------------
+# Unicode encoding edge cases
+# ---------------------------------------------------------------------------
+
+
+class TestUnicodeEncoding(unittest.TestCase):
```
mikelane, line 613: All 6 new test methods in TestUnicodeEncoding are missing -> None return type annotations, unlike every other test method in this file (all 85+ existing test methods consistently use -> None). This causes mypy --strict errors.

mikelane (commented):
## PR Review: #276

### Summary

Three well-scoped changes to the bash-redirect hook: surfacing review_change in hook messages, fixing Unicode encoding issues (em-dashes, arrows), and fixing two tokenizer bugs (escaped-newline handling, gh api contents false positives). The production code changes are correct and well-structured. All 91 tests pass, ruff is clean, and there is zero regression risk (branch diverged from exact origin/main HEAD).

### Verdict: APPROVE (with recommendations below)

### Findings

**LOW (2 findings)**

1. Missing -> None annotations on new test methods: All 6 new test methods in TestUnicodeEncoding lack return type annotations, unlike 85+ existing test methods.
2. Missing regression tests for bug fixes: The escaped-newline and gh-api-contents tokenizer fixes are not backed by tests proving the bugs existed or are now prevented. Manual verification confirms correctness.

### What is excellent

- Production code passes mypy --strict with zero errors
- Unicode test design is thorough, covering all three stderr paths plus advisory stdout
- Message placement is precise in both block and advisory paths
- Backward compatibility preserved (ISSUE-239 still passes)
- Zero regression risk (branch at exact main HEAD)

### Acceptance criteria (issue 265)

- [x] Advisory path surfaces review_change
- [x] Hard-block path surfaces review_change
- [x] Both paths keep get_change_manifest
- [x] All existing ISSUE-239 scenarios pass

### Recommendations (non-blocking)

1. Add -> None to new test methods
2. Add regression tests for gh api contents edge cases
3. Update _matches_gh_api_contents docstring to reflect tokenizer approach

PR #317, opened 2026-05-30
feat(shim): GIT_PRISM_DEBUG_RESOLVER stderr emission; activate final @ISSUE-299 scenario

### LINKED ISSUE #299: feat(shim): add --debug-resolver flag to StdRealGitExec for BDD assertion on resolver path

mikelane:
## What

Adds `GIT_PRISM_DEBUG_RESOLVER` — when set, the shim prints `resolved real git to <path>` to stderr before exec, so the BDD scenario *"Real-git resolver skips the shim directory when walking PATH"* can assert the resolved path is outside the shim dir. Clears the **last** `@not_implemented` scenario in the PATH-shim epic.

## How

- `src/shim/real_git.rs`:
  - `passthrough` emits the resolved path to stderr when `GIT_PRISM_DEBUG_RESOLVER` is set (inside the `#[cfg(unix)]` arm, before exec; no behavior change otherwise).
  - `shim_canonical_path` falls back to `std::env::current_exe()` when `argv0` is a bare name. **Load-bearing for loop-prevention**: invoked via a bare `git` symlink, `argv0 == "git"` doesn't canonicalize, so without the fallback the resolver would return the shim's own symlink as "real git" and infinite-loop. Adversarial-QA proved this via mutation.
- BDD: `@ISSUE-299` scenario activated (RED commit removes `@not_implemented`; GREEN implements); the step parses the stderr line.

RED (`8d23b2e`) precedes GREEN (`585d3d6`). Adversarial-QA PASS, mutation-verified.

Closes #299

### DIFF
--- a/bdd/features/shim_wrapper.feature
+++ b/bdd/features/shim_wrapper.feature
@@ -151,7 +151,7 @@ Feature: PATH-shim wrapper for structured git output
     When I run the shim as "git" pointing at the env-dumper with CLAUDECODE=1
     Then the env-dumper output contains "GIT_PRISM_INSIDE_SHIM=1"
 
-  @ISSUE-299 @not_implemented
+  @ISSUE-299
   Scenario: Real-git resolver skips the shim directory when walking PATH
     # The shim binary lives in tmpdir/bin. The resolver must skip that
     # directory and find the system git elsewhere in PATH. Without this
--- a/bdd/steps/shim_wrapper_steps.py
+++ b/bdd/steps/shim_wrapper_steps.py
@@ -248,8 +248,16 @@ def step_run_shim_with_both_flags(context: Context, command: str) -> None:
 
 @when('I run the shim as "{command}" without any agent env vars')
 def step_run_shim_no_agent_vars(context: Context, command: str) -> None:
-    """Invoke the shim with no agent env vars set (non-agent invocation)."""
-    _run_shim(context, _parse_shim_command(command), extra_env=None)
+    """Invoke the shim with no agent env vars set (non-agent invocation).
+
+    Always sets GIT_PRISM_DEBUG_RESOLVER=1 so the shim emits the resolved
+    real-git path to stderr, enabling the resolver-path assertions.
+    """
+    _run_shim(
+        context,
+        _parse_shim_command(command),
+        extra_env={"GIT_PRISM_DEBUG_RESOLVER": "1"},
+    )
 
 
 @when('I run the shim as "git" pointing at the env-dumper with CLAUDECODE=1')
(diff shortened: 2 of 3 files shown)


mikelane (commented):
**APPROVED — 0 CRITICAL, 0 HIGH**

## Summary

Clean, minimal, well-scoped implementation. The two functional changes — `GIT_PRISM_DEBUG_RESOLVER` stderr emission and the `current_exe()` fallback in `shim_canonical_path` — are correct, necessary, and correctly placed. Scope is exactly 3 files: `src/shim/real_git.rs`, `bdd/features/shim_wrapper.feature`, `bdd/steps/shim_wrapper_steps.py`.

## Quality Tools

- `cargo fmt --check`: PASS
- `cargo clippy --all-targets -- -D warnings`: 3 warnings, all pre-existing from vendored Kotlin C scanner. Zero warnings in Rust source. Not introduced by this PR.
- `cargo test -- real_git`: 9/9 PASS
- `behave bdd/features/shim_wrapper.feature --tags=@ISSUE-299`: 1 scenario PASS

## Logic Review: `shim_canonical_path` fallback

The `current_exe()` fallback is load-bearing and correct. When the shim is invoked as a bare `git` via PATH lookup, `argv0` is `"git"` — `Path::new("git").canonicalize()` fails (no such file relative to cwd), so `.ok()` returns `None`. `current_exe()` then provides the real on-disk path of the running binary (the shim). Without this fallback, `shim_path` would be `None`, the loop-guard is bypassed for every PATH candidate, and the first `git` found — the shim symlink — is returned, causing infinite exec loop. Adversarial-QA mutation-tested this; the logic is sound.

False-positive risk (wrongly skipping legit git): none. `current_exe()` returns the shim's path, not a real git binary. Only candidates that canonicalize to the shim itself are skipped.

False-negative risk (exec'ing the shim → loop): prevented. The fallback ensures `shim_path` is always `Some` on Unix, so the guard is always active.

## Debug Emission Placement

The debug `eprintln!` fires after successful resolution, before `cmd.exec()`. Correctly `#[cfg(unix)]`-gated (inside the unix arm). Only emits when `GIT_PRISM_DEBUG_RESOLVER` is set — no noise in normal operation. Format string matches what the BDD step parses (`resolved real git to <path>`). Correct.

## Regression Check (Phase 8)

Branch divergence point: `7b8ad8f`. `git log merge-base..origin/main` produced zero commits — branch is fully rebased against current main. Zero overlap between PR-modified files and main changes. No regression risk.

## RED→GREEN History

- `8d23b2e`: removes `@not_implemented` from `@ISSUE-299` — correct RED commit, proving test failed before implementation
- `585d3d6`: adds debug emission + `current_exe()` fallback — correct GREEN commit

## Findings

**LOW** (`bdd/features/shim_wrapper.feature` line 163): Comment reads "until then this scenario stays `@not_implemented`" — stale post-activation. Delete the trailing sentence. Not a blocker.

## CI Status

Format: PASS. Integrity: PASS. Clippy/Tests/BDD/Mutants: pending at time of review. Gate merge on all-green CI.

## Decision

APPROVE. No blockers. One LOW cosmetic finding (stale comment). Implementation is correct, tests are behavioral, RED→GREEN history is clean, no regression against main.

Issue #326, opened 2026-05-30
feat(hooks): deprecate and remove redirect hook (bash_redirect_hook.py)


mikelane:
### BDD Acceptance Scenarios

This issue must make the `@ISSUE-326` scenarios pass. Feature files will be present in `bdd/features/` after BDD Bootstrap (#321) merges.

**Scenarios owned by this issue:** `@ISSUE-326` in `bdd/features/shim_deprecation.feature`

### Context

The redirect hook (`hooks/bash_redirect_hook.py`, installed via `git-prism hooks install`) has a structural limitation: Claude Code's `PreToolUse` hook fires only on the top-level Bash command string — it is NOT a syscall interceptor. Any `git` call inside a subprocess (make, lefthook, cargo build scripts, pre-commit, any wrapper) never triggers it. The PATH shim closes this gap by intercepting at the exec/PATH layer and is a strict superset of the hook's coverage on Unix (when PATH is configured correctly).

Once Windows passthrough (#322), gh interception (#323), the shim subcommand (#324), and auto-PATH-setup (#325) are merged, the redirect hook provides no capability the shim does not already provide — and provides less.

### This Issue Is Terminal

This is the last implementation issue before the capstone. **Do not start this until #322, #323, #324, and #325 are all merged.**

### Tasks

1. **Mark `git-prism hooks install` deprecated** (without `--path-shim`): print a clear deprecation warning to stderr pointing to `git-prism shim install`. Keep the hook functional for this release cycle.

2. **Remove `hooks/bash_redirect_hook.py`**: delete the file.

3. **Remove the redirect-hook install path** from `src/hooks.rs` (or wherever it lives): `git-prism hooks install` without `--path-shim` should exit non-zero with an informative message after removal.

4. **Update README**: remove redirect hook install instructions; document shim as the canonical install path.

5. **Update CHANGELOG**: record the removal with migration instructions.

6. **Write ADR** `docs/decisions/NNNN-redirect-hook-removal.md`: context (why the hook is inferior), decision (remove it), consequences (users must re-install using shim).

### Hard Constraint

**v0.9.0 release is BLOCKED until this issue is closed.** This is the terminal implementation gate for the epic.

### TDD

First commit removes `@not_implemented` from `@ISSUE-326` scenarios (RED). Implement via TDD.

### Blocked By

Issues #322, #323, #324, and #325 must all be merged before this issue can start. BDD Bootstrap (#321) must also be merged.

---

[...]

Issue #325, opened 2026-05-30
feat(shim): auto-PATH-setup UX — offer to append export PATH to shell rc on install


mikelane:
### BDD Acceptance Scenarios

This issue must make the `@ISSUE-325` scenarios pass. Feature files will be present in `bdd/features/` after BDD Bootstrap (#321) merges.

**Scenarios owned by this issue:** `@ISSUE-325` in `bdd/features/shim_auto_path.feature`

### Problem

After running `git-prism shim install`, the shim directory must appear in `PATH` for the interceptor to work. Today the user must manually add it. This zero-setup gap means the redirect hook (which requires no PATH changes) is still more ergonomic for casual users.

### Solution

During `git-prism shim install`, detect whether the shim directory is already in `PATH`. If not, prompt the user (interactive TTY only — skip in CI):

```
The shim directory (~/.local/share/git-prism/shim) is not in your PATH.
Add it automatically? [y/N]
```

If the user answers `y`:
- Detect the user's shell (via `$SHELL`)
- Append to the appropriate rc file (`~/.bashrc`, `~/.zshrc`, `~/.config/fish/config.fish`)
- Append exactly: `export PATH="$HOME/.local/share/git-prism/shim:$PATH"` (or fish equivalent)
- Idempotent: check for existing line before appending
- Print: `Added to ~/.zshrc. Restart your shell or run: source ~/.zshrc`

If the user answers `n` (or is not on a TTY):
- Print manual instructions to stdout — do not modify any file

### Edge Cases

- Non-interactive (piped stdin, `CI=true`): skip prompt, print manual instructions
- Unknown shell: skip auto-setup, print manual instructions for bash and zsh
- Fish shell: use `set -Ux` or `fish_add_path` form
- Already in PATH: skip entirely, no prompt

### TDD

First commit removes `@not_implemented` from `@ISSUE-325` scenarios (RED). Implement via TDD.

### Blocked By

BDD Bootstrap (#321) must merge first. Also blocked by the shim subcommand issue (#324), since this feature is part of `git-prism shim install`.

---

## BDD Acceptance Scenarios (merged — #321)

The BDD bootstrap merged. The actual scenarios live in **`bdd/features/shim_parity.feature`** (a single file for the whole epic — ignore the per-concern filename mentioned above). The **first commit on this branch must remove `@not_implemented`** from the `@ISSUE-325` scenarios below (the RED commit) before any production code:

```gherkin
[...]

Issue #322, opened 2026-05-30
feat(shim): Windows passthrough — spawn+wait instead of execvp on non-unix


mikelane:
### BDD Acceptance Scenarios

This issue must make the `@ISSUE-322` scenarios pass (note: the BDD Bootstrap issue is #321; actual scenario text will be present in `bdd/features/` after that issue merges).

**Scenarios owned by this issue:** `@ISSUE-322` in `bdd/features/shim_windows.feature`

### Problem

The shim currently uses `cmd.exec()` (Unix `execvp`) as the passthrough mechanism. On Windows there is no `execvp` — the `#[cfg(not(unix))]` arm currently just bails (likely a `todo!()` or early-return). This means the shim does not function on Windows at all, leaving `windows-latest` CI broken for shim-dependent tests.

### Solution

On non-unix platforms, replace `execvp` with:

```rust
#[cfg(not(unix))]
fn passthrough(cmd: &str, args: &[&str]) -> ! {
    let status = std::process::Command::new(cmd)
        .args(args)
        .status()
        .expect("failed to spawn real git");
    std::process::exit(status.code().unwrap_or(1));
}
```

Key requirements:
- Exit code must be forwarded faithfully (not always 0 or always 1)
- stdout and stderr must not be captured — pass through directly (inherit)
- Must keep `windows-latest` CI green

### TDD

First commit removes `@not_implemented` from `@ISSUE-322` scenarios (RED). Then implement via TDD.

### Blocked By

BDD Bootstrap (#321) must merge first — the `@ISSUE-322` scenarios and step definitions must be present before this branch starts.

---

## BDD Acceptance Scenarios (merged — #321)

The BDD bootstrap merged. The actual scenarios live in **`bdd/features/shim_parity.feature`** (a single file for the whole epic — ignore the per-concern filename mentioned above). The **first commit on this branch must remove `@not_implemented`** from the `@ISSUE-322` scenarios below (the RED commit) before any production code:

```gherkin
  Rule: Windows passthrough preserves real-git exit code and output

    @ISSUE-322 @not_implemented @windows
    Scenario: Successful git command exit code and stdout match real git on Windows
      Given a fixture git repository with two commits
      When the shim runs "git status" without agent env vars
      And the real git binary runs "git status" in the same repository
      Then the shim exit code matches the real git exit code
      And the shim stdout matches the real git stdout

    @ISSUE-322 @not_implemented @windows
    Scenario: Failed git command exit code and stderr match real git on Windows
[...]

Issue #321, opened 2026-05-30
test(bdd): BDD bootstrap — shim parity and redirect hook sunset


mikelane:
### Goal

Write ALL Gherkin scenarios for the shim-parity epic before any implementation begins. Step definitions must call the real `git-prism` binary and assert on observable behavior. Every scenario must run and **FAIL** (RED) before this issue closes — proving the behavior does not exist yet.

### Framework

Per repo convention: **Python + behave** over the Rust binary. Step definitions live under `bdd/steps/`, feature files under `bdd/features/`. Python is used precisely because it cannot import production Rust internals — this is black-box behavioral testing.

### Blocked By

**This issue is blocked by the spike (#320).** Read the ADR produced by the spike before writing scenarios — the ADR determines exact install/launch requirements and interface assumptions. Do not guess.

### Scenarios to Cover

> **Tag mapping (corrected):** each scenario group is tagged with the implementation issue that will make it pass — Windows→`@ISSUE-322`, gh→`@ISSUE-323`, shim subcommand→`@ISSUE-324`, auto-PATH→`@ISSUE-325`, deprecate/remove→`@ISSUE-326`. The BDD bootstrap (#321) owns no scenarios of its own; it authors them all.

#### @ISSUE-322 — Windows shim passthrough (non-unix)

```gherkin
Feature: Windows shim passthrough
  @ISSUE-322
  Scenario: git commands pass through on non-unix platforms
    Given the shim is installed
    And the platform does not support execvp
    When git status is invoked via the shim
    Then the shim spawns git as a subprocess
    And the exit code matches the real git exit code
    And stdout matches the real git output

  @ISSUE-322
  Scenario: git command failure exit code is preserved on non-unix
    Given the shim is installed
    And the platform does not support execvp
    When git is invoked with an invalid subcommand
    Then the exit code is non-zero
    And the error output matches what real git emits
```

#### @ISSUE-323 — gh interception

```gherkin
Feature: gh CLI interception by the shim
  @ISSUE-323
  Scenario: gh pr diff is intercepted and routed through git-prism
    Given the shim is installed and gh is symlinked to the shim
    When an agent runs "gh pr diff 42"
    Then git-prism receives the gh pr diff call
    And the structured diff output is returned

  @ISSUE-323
  Scenario: gh commands outside pr diff pass through unmodified
    Given the shim is installed and gh is symlinked to the shim
    When an agent runs "gh repo view"
    Then the real gh binary handles the command
[...]

PR #329, opened 2026-05-30
docs(adr): ADR 0010 — shim direct-call interception (spike #320)

### LINKED ISSUE #320: spike: verify shim intercepts Claude Code's own direct git calls
### Goal

Empirically determine whether and under what conditions the PATH shim intercepts `git` calls emitted directly by Claude Code's Bash tool — not just nested subprocess calls. This is the **load-bearing unknown** for the entire "sunset the redirect hook" epic. If direct-call interception does not reliably work, the hook cannot be replaced.

### Background

The PATH shim works by inserting a directory into `PATH` that contains a `git` wrapper. This directory must be inherited by the process that invokes `git`. The uncertainty:

- Claude Code's Bash tool spawns subshells via `bash -c "..."`. These subshells do NOT re-source `~/.bashrc` or `~/.zshrc`.
- Whether the shim directory appears in `PATH` inside those subshells depends entirely on how Claude Code launches its own process and whether `PATH` is inherited vs reset.
- The `PreToolUse` hook fires on the Bash command string — it is NOT a syscall interceptor. The shim is a syscall-layer interceptor (via PATH). These are orthogonal.

### Questions to Answer

1. Does `git` resolve through the shim directory when Claude Code's Bash tool runs `git status`?
2. Does it resolve through the shim when `git` is called inside a `make` target, lefthook hook, or cargo build script invoked by Claude's Bash tool?
3. What are the exact install requirements (which shell rc must be modified, must Claude be restarted, must the terminal be restarted)?
4. Is there a race or timing window where the shim is not yet in PATH when Claude Code first launches?
5. Does the answer differ on macOS vs Linux?

### Deliverable

An ADR committed to `docs/decisions/NNNN-shim-direct-call-interception.md` on main with:
- **Context**: what question we were answering
- **Decision**: what we concluded (works / doesn't work / conditional)
- **Consequences**: exact install + launch requirements; any caveats that must appear in user-facing docs

The spike branch (`spike/shim-direct-call`) is disposable — prototype code is evidence, not foundation. **Do NOT merge the spike branch.**

### Rules

- No TDD during this spike
- One agent session time-box
- Commit only the ADR to main, not prototype code
- If the spike concludes direct-call interception is unreliable, that is a valid (and important) outcome — document it and the epic will adapt

### Gates Everything

This spike issue **blocks** the BDD Bootstrap and all implementation issues. Do not start BDD or implementation until the ADR is committed and this issue is closed.

mikelane:
## What

ADR 0010 records the outcome of spike #320: **does the PATH shim intercept Claude Code's own direct Bash-tool git calls?**

Answer: **yes, conditionally** — and the shim is a strict superset of the redirect hook, which unblocks the hook's removal (#326).

## Empirical findings (run inside a live Claude Code session)

- Claude Code's Bash tool runs `/bin/zsh -c` and **sources a per-launch shell snapshot** that freezes a resolved `export PATH='…'`. The snapshot is generated by sourcing the user's rc (proven: it captures 45 rc-defined aliases + functions).
- A fake-`git`-on-PATH lab confirmed interception for **direct, `sh -c`-nested, Makefile-nested, and 2-level-deep** calls. `env -i` (PATH stripped) is the only non-intercepted case — confirming the dependency is purely PATH.
- The nested cases are exactly what the redirect hook is structurally blind to. **Superset proven.**

## Condition + consequence

- Shim dir must be in PATH **at `claude` launch time** (snapshot is frozen then).
- rc edits (the #325 auto-PATH feature) work on the **next launch** — a Claude Code restart is required and must be loud in install output + README.

## Scope

Docs-only. The spike's prototype lab code stays disposable (`/tmp/shim-spike`); only the ADR is merged, per the epic's spike rules.

Closes #320

### DIFF
--- a/docs/decisions/0010-shim-direct-call-interception.md
+++ b/docs/decisions/0010-shim-direct-call-interception.md
@@ -1,0 +1,160 @@
+# ADR 0010: Shim Direct-Call Interception via Claude Code's Shell Snapshot
+
+- **Status**: Accepted (spike outcome)
+- **Date**: 2026-05-30
+- **Context**: Epic #328 — sunset the redirect hook; spike #320
+- **Supersedes assumptions in**: the epic premise that direct-call interception was an open question
+
+## Context
+
+Epic #328 wants to remove the bundled redirect hook (`hooks/bash_redirect_hook.py`)
+and rely on the PATH shim alone. That is only safe if the shim is a true **superset**
+of the hook's coverage — including the hook's one job: intercepting `git` that an agent
+issues directly through Claude Code's Bash tool.
+
+The redirect hook is a Claude Code `PreToolUse` hook. It fires on the literal Bash
+command *string* an agent emits, before execution. The PATH shim intercepts at a
+different layer entirely: it puts a `git` symlink (→ `git-prism`) on `PATH` ahead of
+the real git, so every `git` *exec* in that process tree flows through git-prism. These
+two mechanisms are orthogonal (see [ADR 0009](0009-path-shim-architecture.md) §2).
+
+The load-bearing unknown for the whole epic: **does the shim's `PATH` entry actually
+reach the subshell that Claude Code's Bash tool spawns?** The worry was that Claude Code
+runs each Bash command in a fresh non-login subshell that doesn't re-source `~/.zshrc`,
+so a `PATH` export added to rc might never be seen.
+
+This spike answered that empirically, from inside a live Claude Code session — the exact
+process under investigation.
+
+## How Claude Code's Bash tool actually runs commands
+
+Observed on macOS, Claude Code launched from an interactive zsh under Warp:
+
+1. Each Bash tool call is run by **`/bin/zsh -c`** (not `bash -c` as the epic assumed),
+   and the command is wrapped so it **first sources a shell snapshot**:
+
+   ```
+   /bin/zsh -c 'source ~/.claude/shell-snapshots/snapshot-zsh-<id>.sh 2>/dev/null || true && … && eval <command>'
+   ```
+
+2. That snapshot is **generated once per `claude` launch** (one file per launch; 14
+   were present). Its mtime trailed the `claude` process start by ~47s — it is written
+   at startup, not per command.
+
+3. The snapshot **hardcodes `PATH`** as a single resolved string:
+
+   ```sh
+   export PATH='/Users/.../.cargo/bin:/opt/homebrew/bin:…'
+   ```
+
+   It does **not** re-evaluate rc files at command time. `PATH` for every Bash tool
+   call is whatever string was frozen into the snapshot at launch.
+
+4. The snapshot also captures 45 rc-defined **aliases** (`cat=bat`, `aws=sso-aws`, …)
+   and the user's shell **functions**. Their presence proves snapshot generation
+   **sources the user's rc files** — even though the launching shell was `zsh -g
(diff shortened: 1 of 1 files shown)


mikelane (commented):
## ADR 0010 Review — docs-only PR

**Scope**: Single file addition `docs/decisions/0010-shim-direct-call-interception.md`. No executable code. Docs-only exception applies — full gauntlet not required.

**Regression check**: Branch diverges from main at `d2e1212` (the tip of main). No files modified on main since divergence. Zero deletions in the diff. No regression risk.

**`Closes #320`**: Present in PR body. ✓

---

### Technical accuracy

The ADR makes several empirical claims. Each is internally consistent and cross-checks cleanly:

1. **"Claude Code's Bash tool runs `/bin/zsh -c`"** — The snapshot filename convention (`snapshot-zsh-<id>.sh`) and the 14-snapshot count corroborate this. The claim that Claude Code uses `/bin/zsh` rather than `/bin/bash` is accurately flagged as OS-specific (macOS) and the Limitations section acknowledges Linux parity is expected but unverified. Sound.

2. **"Snapshot is generated once per launch and sources rc"** — The 45-alias evidence is a strong proxy: those aliases could only appear in the snapshot if rc was sourced during snapshot generation. The claim is epistemically honest ("proves" used only for the alias count, not for internals that can't be directly observed). Sound.

3. **"PATH is frozen at launch time, not per-command"** — Follows logically from (2): if the snapshot is written once and hardcodes a resolved `export PATH='…'` string, PATH cannot change between Bash tool calls within a session. Consistent with the "restart Claude Code" consequence. Sound.

4. **Table cases A–E** — The fake-git-on-PATH methodology is the correct level of evidence for an interception claim. The env -i failure (case E) is the right negative control and the ADR correctly reads it: the shim is purely PATH-dependent, which is the expected and acceptable failure mode. Conclusion "shim ⊇ hook" follows from the table. Sound.

5. **"Superset" claim** — The ADR correctly explains *why* the shim is a strict superset: the redirect hook fires on the top-level command string, so `make build` never lets it see the nested `git`. Cases B/C/D prove the shim reaches those nested calls. The logic is airtight.

### Internal consistency

- The "frozen at launch" finding directly explains the consequence (must restart Claude Code) and the limitation (rc edits mid-session are invisible). These three sections are mutually consistent.
- The follow-up unblocking claims (#321, #325, #326) are justified: #321 can now assume the install+restart model; #325's design is confirmed by finding 3; #326's safety depends on the superset claim, which is proven by case B/C/D.
- One small precision note (LOW, not blocking): Question 3 says "A terminal restart is not required — only that the shell launching `claude` has the updated rc in effect." This is technically true but could confuse users: if the current terminal session already has an outdated PATH cached in its environment, launching `claude` from it won't pick up the rc change either. The full-accuracy statement is "the shell launching `claude` must have the updated PATH in its environment, which normally requires opening a new terminal (or running `source ~/.zshrc` in the current one)." Worth a follow-up refinement but not a blocker for an ADR about the shim mechanism.

### Limitations disclosure

The Limitations section is commendably honest:
- macOS/zsh/Warp only — explicitly stated
- Linux/bash/VS Code/tmux not verified — explicitly stated
- Spike code is evidence, not foundation — good epistemic hygiene
- BDD scenarios (#321) and `windows-latest` (#322) called out as the verification paths for unverified environments

This is a well-scoped, technically sound ADR that accurately captures spike findings, draws justified conclusions, and is transparent about evidence boundaries.

---

**Verdict**: APPROVE — no blockers, no CRITICAL/HIGH findings.

Issue #328, opened 2026-05-30
epic: shim parity → sunset the redirect hook (v0.9.0 blocker)


mikelane:
## Goal

Close the remaining gaps between the PATH shim and the redirect hook so the redirect hook (`hooks/bash_redirect_hook.py`) can be deprecated and removed. Once this epic is closed, agents get structured git interception everywhere — direct calls, nested subprocess calls, `gh pr diff` — without any Claude Code hook involvement. The redirect hook is then dead code.

## Context

The PATH-shim epic (#284, now closed) added a PATH-level `git` interceptor. The pre-existing Python redirect hook (`git-prism hooks install`, `hooks/bash_redirect_hook.py`) has a **structural limitation**: Claude Code's `PreToolUse` hook fires only on the top-level Bash command string Claude emits — it is NOT a syscall interceptor. Any `git` invoked inside a subprocess (make, lefthook, cargo build scripts, pre-commit, any wrapper script) NEVER triggers it. The PATH shim closes that gap because it intercepts at the exec/PATH layer.

On Unix with PATH set correctly, the shim is a strict **superset** of the hook's coverage. The goal of this epic is to prove that superset and eliminate the remaining gaps (Windows passthrough, `gh` interception, CLI ergonomics, auto-PATH setup) so the hook can be removed with confidence.

## Hard Constraint

> **v0.9.0 release is BLOCKED until every child issue here is closed and the redirect hook is removed.**

Do not cut v0.9.0 until issue #326 (remove hook) and #327 (capstone + release) are both closed.

## Scope

**In scope:**
- Verifying shim intercepts direct Claude Code git calls (spike)
- Windows shim passthrough (spawn+wait instead of execvp)
- `gh pr diff` interception via shim
- Promoting `git-prism shim` as a first-class CLI subcommand
- Auto-PATH-setup UX during `git-prism shim install`
- Deprecating and removing `hooks/bash_redirect_hook.py` and its install path
- Capstone demo + v0.9.0 release

**Out of scope:**
- Shim support for languages other than Rust (no polyglot rewrites)
- Hook interception for tools other than `git` and `gh`
- Windows CI matrix beyond keeping `windows-latest` green

## Design Documents

- Spike ADR: `docs/decisions/NNNN-shim-direct-call-interception.md` (produced by #320)
- Redirect hook removal ADR: `docs/decisions/NNNN-redirect-hook-removal.md` (produced by #326)

## Acceptance Criteria

[...]

Issue #324, opened 2026-05-30
feat(shim): promote shim to first-class subcommand (git-prism shim install|uninstall|status)


mikelane:
### BDD Acceptance Scenarios

This issue must make the `@ISSUE-324` scenarios pass. Feature files will be present in `bdd/features/` after BDD Bootstrap (#321) merges.

**Scenarios owned by this issue:** `@ISSUE-324` in `bdd/features/shim_subcommand.feature`

### Problem

The shim install currently lives under `git-prism hooks install --path-shim`, which is a category error: the shim is not a Claude Code hook — it is a PATH-layer interceptor. Exposing it under `hooks` misleads users about what it does and couples two unrelated concepts.

### Solution

Add a new top-level subcommand:

```
git-prism shim install    # install the shim into ~/.local/share/git-prism/shim/ (or equivalent)
git-prism shim uninstall  # remove the shim
git-prism shim status     # show whether shim is installed and in PATH
```

**Backward compatibility (mandatory for one release):**

`git-prism hooks install --path-shim` must continue to work and print a deprecation warning to stderr:

```
warning: --path-shim is deprecated; use `git-prism shim install` instead
```

Do NOT hard-rename or remove `--path-shim` in this issue. The removal happens in issue #326 (deprecate+remove redirect hook) after parity is proven.

### CLI surface

This is a public CLI surface change. Update:
- `src/main.rs` clap subcommand wiring
- `src/hooks.rs` or new `src/shim_cmd.rs` for the subcommand handler
- README: add `git-prism shim` to the commands table
- CHANGELOG: document the new subcommand and the deprecation notice

### TDD

First commit removes `@not_implemented` from `@ISSUE-324` scenarios (RED). Implement via TDD.

### Blocked By

BDD Bootstrap (#321) must merge first.

---

## BDD Acceptance Scenarios (merged — #321)

The BDD bootstrap merged. The actual scenarios live in **`bdd/features/shim_parity.feature`** (a single file for the whole epic — ignore the per-concern filename mentioned above). The **first commit on this branch must remove `@not_implemented`** from the `@ISSUE-324` scenarios below (the RED commit) before any production code:

```gherkin
  Rule: git-prism shim install creates the symlink via the new subcommand

    @ISSUE-324 @not_implemented
    Scenario: shim install creates the git symlink in the shim directory
      Given an isolated HOME directory
      When I run "git-prism shim install" with the isolated HOME
      Then the exit code is 0
[...]

PR #332, opened 2026-05-30
fix(shim): Windows passthrough via spawn+wait (#322)

### LINKED ISSUE #322: feat(shim): Windows passthrough — spawn+wait instead of execvp on non-unix

mikelane:
## What

Makes the PATH shim function on Windows. Epic #328 (shim parity → sunset the redirect hook).

The shim's passthrough used `execvp` on unix; the `#[cfg(not(unix))]` arm was a stub that exited 127 unconditionally — so the shim was non-functional on Windows. This replaces that arm with **spawn+wait**: `std::process::Command::new(real_git).args(...).status()` then `exit(status.code())`, inheriting stdout/stderr and forwarding the real exit code. Also adds a `#[cfg(not(unix))]` `resolve_real_git()` PATH walker (`;`-separated, tries `git.exe` then `git`, skips the shim symlink to avoid self-recursion). **The unix execvp path is byte-for-byte unchanged.**

## Tests

- 2 `@ISSUE-322` scenarios (`@windows`-tagged) assert passthrough exit-code + stdout/stderr parity. They **pass on unix** (execvp passthrough already works) — `windows-latest` CI is their authoritative RED→GREEN gate, since the Windows-specific spawn+wait path is unreachable on unix. This is documented in the feature file's PLATFORM NOTE.
- Cross-platform unit tests for exit-code mapping (`exec_failure_exit_code` → 126/127).

## Verification

- Local: `cargo test`/`clippy --all-targets`/`fmt` clean; `behave --tags="@ISSUE-322"` passes on unix.
- **Windows typecheck**: adversarial-QA cross-compiled `real_git.rs` to `x86_64-pc-windows-gnu` (isolated crate, 0 errors), confirming the `#[cfg(not(unix))]` arms + test module compile on the Windows target. Final runtime verification is the `windows-latest` CI job below.

## Gauntlet

adversarial-QA PASS (0 reachable bugs); rust-purist ship-worthy.

### Non-blocking INFO (deferred, theoretical)
- `passthrough()` non-unix arm casts `i32`→`u8` for the exit code. Unreachable in practice (git exits 0–255); a `// SAFETY:` comment documenting the invariant would be the ideal polish.
- The bare-`git` (no `.exe`) PATH fallback won't spawn via `Command` on Windows (no `PATHEXT` resolution), but `git.exe` is tried first, so this only ever yields a clear spawn error — harmless.

## Notes

- `@ISSUE-323/325/326` remain `@not_implemented` (excluded from CI). Only `@ISSUE-322` activated.
- Rebased onto current main (post-#324).

Closes #322

### DIFF
--- a/bdd/features/shim_parity.feature
+++ b/bdd/features/shim_parity.feature
@@ -48,7 +48,7 @@ Feature: Shim parity with the redirect hook and first-class shim subcommand
 
   Rule: Windows passthrough preserves real-git exit code and output
 
-    @ISSUE-322 @not_implemented @windows
+    @ISSUE-322 @windows
     Scenario: Successful git command exit code and stdout match real git on Windows
       Given a fixture git repository with two commits
       When the shim runs "git status" without agent env vars
@@ -56,7 +56,7 @@ Feature: Shim parity with the redirect hook and first-class shim subcommand
       Then the shim exit code matches the real git exit code
       And the shim stdout matches the real git stdout
 
-    @ISSUE-322 @not_implemented @windows
+    @ISSUE-322 @windows
     Scenario: Failed git command exit code and stderr match real git on Windows
       Given a fixture git repository with two commits
       When the shim runs "git not-a-real-subcommand" without agent env vars
--- a/src/shim/real_git.rs
+++ b/src/shim/real_git.rs
@@ -76,9 +76,33 @@ impl<E: EnvSource> RealGitExec for StdRealGitExec<'_, E> {
         }
         #[cfg(not(unix))]
         {
-            let _ = argv;
-            eprintln!("git-prism shim: passthrough is not supported on non-Unix platforms");
-            ExitCode::from(127)
+            let real = match resolve_real_git(self.argv0, self.env) {
+                Some(p) => p,
+                None => {
+                    eprintln!("git-prism shim: could not find real git binary");
+                    return ExitCode::from(127);
+                }
+            };
+
+            if self.env.get("GIT_PRISM_DEBUG_RESOLVER").is_some() {
+                eprintln!("git-prism shim: resolved real git to {}", real.display());
+            }
+
+            let status = std::process::Command::new(&real)
+                .args(argv.iter().skip(1))
+                .env("GIT_PRISM_INSIDE_SHIM", "1")
+                .status();
+
+            match status {
+                Ok(s) => {
+                    let code = s.code().unwrap_or(1);
+                    ExitCode::from(code as u8)
+                }
+                Err(err) => {
+                    eprintln!("git-prism shim: spawn of {} failed: {err}", real.display());
+                    ExitCode::from(exec_failure_exit_code(&err))
+                }
+            }
         }
     }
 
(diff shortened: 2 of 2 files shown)


mikelane (commented):
## pr-reviewer verdict: APPROVE — Windows passthrough is correct; 2 non-blocking LOW items

**Decision: APPROVE**

### Summary

This PR implements the non-unix `passthrough()` arm correctly using `std::process::Command::new(...).status()` with inherited stdio and forwarded exit code. The unix `execvp` path is provably unchanged — diff contains zero modifications to the `#[cfg(unix)]` block. The Windows `resolve_real_git()` PATH walker is well-structured with correct shim self-recursion prevention via `canonicalize()`. Squash-merge regression check: branch is rebased directly onto `b2cba05` (current main HEAD) — merge-base equals origin/main, zero regression surface from sibling PRs.

### Phase-by-phase findings

**Correctness (Phase 0):** No bugs confirmed. Stdio inheritance is correct — `Command::status()` inherits stdin/stdout/stderr from the parent process by default in Rust std, exactly matching `execvp`'s observable behaviour for interactive and pipe contexts. Exit code forwarding via `s.code().unwrap_or(1)` is correct: the only case `code()` returns `None` on Windows is process termination by signal, which git never does — 1 is a safe sentinel.

**Quality tools (Phase 1):** Windows-latest CI green (the authoritative gate for this Windows-specific code path). Clippy, Format, Integrity, BDD Acceptance, Mutants incremental all green.

**Config integrity (Phase 2):** No suppressions or exclusions added.

**Code standards (Phase 3):** Both new cfg arms mirror the structure of their unix counterparts. PATH splitting on `;` (Windows separator) is correct. The `.exe`-first probe order is idiomatic Windows practice.

**Self-documenting code (Phase 4):** All comments explain WHY (platform constraint, separator difference, extension resolution). No WHAT comments.

**Test quality (Phase 5):** The two `@ISSUE-322` BDD scenarios are the correct test vehicle for Windows-specific spawn+wait behaviour. The `@not_implemented` tags are removed in the right order per the epic SDLC contract. Existing unix unit tests are all `#[cfg(unix)]`-guarded and unaffected.

**Architecture (Phase 6):** The cfg-split is the correct Rust idiom for platform-divergent code. No new abstractions, no speculative generality. YAGNI respected.

**Security (Phase 7):** `Command::new` with a `&[&str]` args slice — no shell involvement, no injection risk. No new security surface.

**Regression vs main (Phase 8):** Merge-base == origin/main HEAD (`b2cba05`). Zero files modified on main since divergence. The `#[cfg(unix)]` block in `passthrough()` is byte-for-byte identical to main. No regression.

---

### Blockers
None.

### LOW findings (non-blocking)

**1. `src/shim/real_git.rs` ~line 98 — `i32` → `u8` silent truncation**
`s.code()` returns `Option<i32>`; `code as u8` truncates silently if the OS delivers a value outside 0–255. Unreachable for any real git invocation (POSIX convention: 0–255), but a `// SAFETY:` comment surfacing that reasoning would help future readers. Recommend:
```rust
// SAFETY: git exit codes are always 0-255 by POSIX convention;
// the i32->u8 cast is lossless for any value git actually produces.
ExitCode::from(code as u8)
```

**2. `src/shim/real_git.rs` non-unix `resolve_real_git` — intentional absence of fallback chain**
The unix version has a hardcoded `/usr/bin/git`, `/usr/local/bin/git`, `/opt/homebrew/bin/git` fallback chain. The Windows version has none (returns `None` if PATH search fails). This asymmetry is correct — there is no standard Windows fallback path — but a doc comment clarifying the intentional design would help future maintainers. Non-blocking.

---

### Praise

The shim self-recursion prevention in the non-unix resolver is clean and correct: `canonicalize().ok()` with graceful fallback to `current_exe()` handles bare-name `argv0` correctly. The try-both-extensions pattern (`git.exe` then `git`) with `.exe` preference is exactly right for Windows. The overall diff is admirably minimal — only the two cfg arms that needed to change were changed.

PR #333, opened 2026-05-30
feat(shim): gh CLI interception — route gh pr diff through git-prism (#323)

### LINKED ISSUE #323: feat(shim): gh interception — extend shim classifier to handle gh pr diff
### BDD Acceptance Scenarios

This issue must make the `@ISSUE-323` scenarios pass. Actual feature files will be present in `bdd/features/` after BDD Bootstrap (#321) merges.

**Scenarios owned by this issue:** `@ISSUE-323` in `bdd/features/shim_gh.feature`

### Problem

The redirect hook intercepts `gh pr diff` calls (it fires on Claude Code's Bash-tool command strings). The PATH shim currently wraps only `git` — it does not intercept `gh`. An agent using `gh pr diff` after the hook is removed would lose interception coverage.

### Design Decision (spike-informed)

The spike ADR (`docs/decisions/NNNN-shim-direct-call-interception.md`) must be read before implementing. The ADR will specify whether:

- **Option A**: Symlink `gh` → the shim binary, and extend the shim's `classify()` to recognize `argv[0] == "gh"` + route `gh pr diff <N>` to git-prism's structured diff
- **Option B**: Explicitly scope `gh` out of the shim with rationale (e.g., if `gh` interception is unreliable or out of scope for this release)

Whichever option the ADR recommends, implement that. If Option B, document why in a one-liner ADR alongside the code.

### If Option A (symlink + classifier extension)

- `git-prism shim install` creates `$SHIM_DIR/gh -> git-prism-shim`
- `classify()` gains a `gh` branch: `argv[0] == "gh" && argv[1] == "pr" && argv[2] == "diff"` → route to structured diff handler
- All other `gh` subcommands pass through to the real `gh` binary
- `git-prism shim uninstall` removes the `gh` symlink

### TDD

First commit removes `@not_implemented` from `@ISSUE-323` scenarios (RED). Implement via TDD.

### Blocked By

BDD Bootstrap (#321) must merge first.

---

## BDD Acceptance Scenarios (merged — #321)

The BDD bootstrap merged. The actual scenarios live in **`bdd/features/shim_parity.feature`** (a single file for the whole epic — ignore the per-concern filename mentioned above). The **first commit on this branch must remove `@not_implemented`** from the `@ISSUE-323` scenarios below (the RED commit) before any production code:

```gherkin
  Rule: gh pr diff is routed through git-prism structured output

    @ISSUE-323 @not_implemented
    Scenario: gh pr diff produces structured JSON when invoked via the shim
      Given the shim is installed with a "gh" symlink
      And a real gh binary is available on PATH
      When an agent runs "gh pr diff 1" via the shim with CLAUDECODE=1
      Then the exit code is 0
      And the output is valid JSON
      And the JSON output has a "files" array
      And each entry in the "files" array has a "change_scope" field

  Rule: Non-diff gh commands pass through to the real gh binary unchanged

    @ISSUE-323 @not_implemented
    Scenario: gh repo view passes through with matching exit code and output
      Given the shim is installed with a "gh" symlink
      And a real gh binary is available on PATH
      When an agent runs "gh repo view --json name" via the shim with CLAUDECODE=1
      And the real gh binary runs "gh repo view --json name" directly
      Then the shim exit code matches the real gh exit code
      And the shim stdout matches the real gh stdout

    @ISSUE-323 @not_implemented
    Scenario: gh issue list passes through unchanged
      Given the shim is installed with a "gh" symlink
      And a real gh binary is available on PATH
      When an agent runs "gh issue list --limit 1" via the shim with CLAUDECODE=1
      And the real gh binary runs "gh issue list --limit 1" directly
      Then the shim exit code matches the real gh exit code
      And the shim stdout matches the real gh stdout
```

mikelane:
## What

Teaches the PATH shim to intercept `gh`. Epic #328 (shim parity → sunset the redirect hook).

- When a symlink named `gh` points at the git-prism binary, the shim recognizes `argv[0]`'s basename as `gh` and routes **`gh pr diff <number>`** through git-prism's structured manifest pipeline (resolves the PR's base..head via `gh pr view --json baseRefOid,headRefOid`, then computes the manifest). Every other `gh` invocation passes through to the real `gh` unchanged.
- `argv[0]` dispatch extended to `git || gh` (`main.rs`); `classify_gh` (`classify.rs`); `handle_gh_pr_diff` / `resolve_pr_range` / `discover_git_root` (`handlers.rs`); both `passthrough` arms now resolve via `resolve_real_binary` (argv0/name-aware, git on unix + gh and Windows via generic PATH walk).

## Notable fixes found during review

- **classify bug (production):** `argv[0]` via a symlink is a full path (`/x/bin/gh`), so the original bare `== "gh"` match always fell through to git classification. Now uses `Path::file_name()`. The live-gh tests masked this; the hermetic stub test exposed it.
- **flag misclassification:** `gh pr diff --help` / `--web` / `--patch` were treated as the PR number → invalid `gh pr view` → exit 1. Now intercepts only when the arg after `diff` is all-numeric; everything else passes through.

## Tests — hermetic, no live GitHub

The 3 `@ISSUE-323` scenarios are fully hermetic: a **stub `gh`** is written to a tempdir, placed first on PATH, and returns fixture-repo SHAs for `pr view --json` (so the manifest is computed over a **local** fixture) and canned output for passthrough. No network, no auth, no live PR — addresses the flakiness + security risk of tests touching real services. Harness hardened per review: SHA-shape assertion before interpolation, `printf '%s'` (format-string safety), fail-loud stub on unhandled args.

## Gauntlet

adversarial-QA PASS (runs 3; caught the flag-misclassification blocker, now fixed); rust-purist + python-purist clean (no blockers; doc/dedup nits applied). Recursion bounded (GIT_PRISM_INSIDE_SHIM + shim-skip); git dispatch regression-checked (@ISSUE-287 15/0). Rebased onto current main (post-#322/#324); the real_git.rs merge was reconciled so both passthrough arms use `resolve_real_binary`.

Closes #323

### DIFF
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -55,10 +55,13 @@ pub(crate) fn classify<'a>(argv: &'a [&'a str]) -> Classification<'a> {
 /// Only `gh pr diff <N>` is intercepted; everything else passes through to
 /// the real `gh` binary.
 fn classify_gh<'a>(subcommand: &str, rest: &[&'a str]) -> Classification<'a> {
-    // gh pr diff <N>  →  intercept; everything else passes through.
+    // gh pr diff <N>  →  intercept only if N is a numeric PR number.
+    // Flags (starting with -) and non-numeric tokens pass through.
     if subcommand == "pr"
         && rest.first() == Some(&"diff")
         && let Some(pr_number) = rest.get(1)
+        && !pr_number.starts_with('-')
+        && pr_number.chars().all(|c| c.is_ascii_digit())
     {
         return Classification::GhPrDiff { pr_number };
     }
@@ -192,6 +195,32 @@ mod tests {
     fn it_passes_through_gh_pr_diff_without_number() {
         // "gh pr diff" with no number is ambiguous — pass through.
         assert_eq!(classify(&["gh", "pr", "diff"]), Classification::Passthrough);
+    }
+
+    #[test]
+    fn it_passes_through_gh_pr_diff_help_flag() {
+        // "gh pr diff --help" must pass through, not be treated as PR number "--help"
+        assert_eq!(
+            classify(&["gh", "pr", "diff", "--help"]),
+            Classification::Passthrough
+        );
+    }
+
+    #[test]
+    fn it_passes_through_gh_pr_diff_with_other_flags() {
+        // Other flags like --web, --patch, --name-only must pass through
+        assert_eq!(
+            classify(&["gh", "pr", "diff", "--web"]),
+            Classification::Passthrough
+        );
+        assert_eq!(
+            classify(&["gh", "pr", "diff", "--patch"]),
+            Classification::Passthrough
+        );
+        assert_eq!(
+            classify(&["gh", "pr", "diff", "--name-only"]),
+            Classification::Passthrough
+        );
     }
 
     #[test]
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -34,11 +34,17 @@ pub(crate) fn classify<'a>(argv: &'a [&'a str]) -> Classification<'a> {
     if argv.len() < 2 {
         return Classification::Passthrough;
     }
-    let binary = argv[0];
+    // argv[0] may be an absolute path (e.g. /tmp/bin/gh) when invoked via a
+    // symlink; extract only the filename component for dispatch, matching what
+    // main.rs does for the shim-mode gate.
+    let binary_basename = std::path::Path::new(argv[0])
+        .file_name()
+        .and_then(|s| s.to_str())
+        .unwrap_or(argv[0]);
     let subcommand = argv[1];
     let rest = &argv[2..];
 
-    match binary {
+    match binary_basename {
         "gh" => classify_gh(subcommand, rest),
         _ => classify_git(subcommand, rest),
     }
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -3,7 +3,7 @@
 //! Ports `_classify_git_command` from `hooks/bash_redirect_hook.py` to Rust.
 //! All logic is pure — no I/O, no env reads.
 
-/// The result of classifying a `git …` argv slice.
+/// The result of classifying a `git …` or `gh …` argv slice.
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) enum Classification<'a> {
     /// `git diff <ref>..<ref>` → `get_change_manifest`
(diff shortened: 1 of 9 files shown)


src/shim/classify.rs
```diff
@@ -19,21 +19,57 @@ pub(crate) enum Classification<'a> {
...
+/// Only `gh pr diff <N>` is intercepted; everything else passes through to
+/// the real `gh` binary.
+fn classify_gh<'a>(subcommand: &str, rest: &[&'a str]) -> Classification<'a> {
+    // gh pr diff <N>  →  intercept only if N is a numeric PR number.
+    // Flags (starting with -) and non-numeric tokens pass through.
+    if subcommand == "pr"
+        && rest.first() == Some(&"diff")
+        && let Some(pr_number) = rest.get(1)
+        && !pr_number.starts_with('-')
+        && pr_number.chars().all(|c| c.is_ascii_digit())
```
mikelane, line 64: **LOW / Praise** — The `classify_gh` numeric guard (`!starts_with('-') && chars().all(is_ascii_digit)`) is exactly right. It rules out flags, branch names, SHAs, and the empty-argument case in one clean pass. The adversarial-QA finding that triggered this (flags treated as PR numbers) is now properly closed.

src/shim/handlers.rs
```diff
@@ -135,6 +136,83 @@ fn handle_show_snapshot<W: Write>(sha: &str, repo_path: &Path, out: &mut W) -> a
...
+/// calling `gh pr view <N> --json baseRefOid,headRefOid`.
+///
+/// Uses commit SHAs (not branch names) so the range works even after the PR
+/// branch has been deleted (merged PRs).  The SHAs are permanent; branch names
+/// are ephemeral.
+///
+/// Returns an error when `gh` is not on PATH, exits non-zero, or returns
+/// JSON that cannot be parsed.
+fn resolve_pr_range(pr_number: &str) -> anyhow::Result<String> {
+    let output = std::process::Command::new("gh")
```
mikelane, line 190: **LOW / Observation** — `resolve_pr_range` spawns `std::process::Command::new("gh")` which resolves `gh` from the ambient PATH. This is intentional and correct: by the time the shim reaches this call, it's already inside its own execution context (the shim binary was invoked as `gh`), so `GIT_PRISM_INSIDE_SHIM=1` is NOT set here (it is set only on the forwarded passthrough argv). The real `gh` is whichever `gh` comes first in PATH after the shim dir, which is the correct one. No issue — just confirming this was a deliberate design choice and not an oversight.

mikelane (commented):
## Review — feat(shim): gh CLI interception (#323)

### Decision: APPROVE

No blockers. The implementation is correct, well-tested, and the adversarial-QA cycle caught and fixed two real production bugs before this review. CI is fully green including Windows.

---

### Phase 8 — Regression check (mandatory)

Branch rebased to current main (`80d1bece` = `origin/main` HEAD). The `git log 80d1bece..origin/main` output is empty — no sibling-PR commits to check. Deletion volume is low; `real_git.rs` grew from 435 → 556 lines (additions-only). No regression risk.

---

### Phase 0 — Bug verification

Two confirmed bugs caught by adversarial-QA and fixed before reaching this review:

1. **argv[0] full-path bug** — bare `== "gh"` always fell through on symlink invocation because the OS passes the full path. Fixed by `Path::file_name()` in both `classify` and `passthrough`/`capture`.

2. **Flag misclassification** — `gh pr diff --help` was treated as PR number `"--help"`. Fixed by the `!starts_with('-') && chars().all(is_ascii_digit)` guard in `classify_gh`.

Both fixes are confirmed correct by inspection and covered by unit tests.

---

### Correctness analysis

**`classify_gh`** ()

The five-condition intercept guard is correctly minimal:
- `subcommand == "pr"` — only `pr` subcommand
- `rest.first() == Some(&"diff")` — only `diff` action
- `rest.get(1)` exists — requires explicit PR number
- `!starts_with('-')` — rejects all flags
- `chars().all(is_ascii_digit)` — numeric-only, no branch names/SHAs/empty strings

Everything else returns `Passthrough`. The classify logic is pure with zero I/O — ideal for test coverage.

**`resolve_pr_range`** ()

Uses `baseRefOid`/`headRefOid` (commit SHAs) rather than branch names — correct design choice: merged PRs have deleted branches but permanent SHAs. Error handling is explicit at each step with descriptive `anyhow` messages.

**`discover_git_root`** ()

Clean upward walk. Handles worktrees correctly (`.git` can be a file in worktrees). Fails with an informative error at filesystem root.

**`resolve_real_binary`** ()

Both unix and non-unix implementations are symmetric and correct. For `git` it delegates to `resolve_real_git` (with hardcoded fallbacks); for other binaries (including `gh`) it does the same shim-skipping PATH walk without fallbacks. The unix `passthrough` uses `execvp` (never returns on success); the non-unix version uses `spawn+wait` returning an exit code. Both arms now use `resolve_real_binary` consistently — the reconciliation is correct.

**Recursion safety** — `GIT_PRISM_INSIDE_SHIM` is checked at `run_shim` entry () before any classification or dispatch. Confirmed present.

---

### Test hermeticity

The BDD `@ISSUE-323` scenarios use a stub `gh` written to a tempdir that returns fixture SHAs from local git commits — no network, no auth, no live GitHub. The stub harness is hardened: SHA-shape assertion via `_SHA_PATTERN.fullmatch` before interpolation, `printf '%s'` for format-string safety, fail-loud on unhandled args.

The integration tests in `tests/test_gh_classify.rs` also use `#![cfg(unix)]` at file level (Windows compile fix) and a stub for the flag-passthrough test. The non-diff passthrough test vacuously passes when `gh` is absent — acceptable.

---

### Code quality

- No type suppression, no `#[allow]`, no commented-out code, no TODO without context.
- Comments explain WHY (symlink path issue) not WHAT.
- `classify` function is 12 lines after the refactor into `classify_gh`/`classify_git` — clean SRP split.
- All new functions are well within the 20-line limit. `passthrough` is ~70 lines due to the `#[cfg(unix)]`/`#[cfg(not(unix))]` duplication, which is the correct approach (conditional compilation, not runtime branching).

**Minor LOW observation:** `resolve_pr_range` spawns `Command::new("gh")` by plain name rather than going through `resolve_real_binary`. This is intentional and correct — see inline comment — but could become subtle if the shim is ever invoked with an empty PATH. Not a blocker; worth a `// SAFETY:` comment for future readers if desired.

---

### Praise

The `Path::file_name()` fix for the argv[0] symlink bug is the kind of thing that would have silently broken every user installation and been nearly impossible to debug. Catching it via hermetic stub testing (vs the live-gh test that masked it) is exactly right. The stub infrastructure is production-quality and reusable.

---

### Summary

613 lines added, 26 removed. Zero CRITICAL, zero HIGH findings. The feature is correctly scoped (`gh pr diff <N>` only), the recursion guard is present, the Windows path compiles cleanly, and tests are hermetic. Ready to merge.

Issue #338, opened 2026-06-01
Scripted git output (`--format` / `--pretty` / `--stat`) should pass through to real git, not return manifest JSON


mikelane:
## What happens

The `git` binary resolves to the git-prism shim (`~/.local/share/git-prism/bin/git`, ahead of `/usr/bin/git` on PATH). For scripted format-string invocations it returns a change-manifest JSON instead of the text the flags asked for.

    $ /usr/bin/git show -s --format=%ct HEAD
    1780329644

    $ git show -s --format=%ct HEAD          # git-prism shim
    {"metadata": {"base_ref": "HEAD^", "head_ref": "HEAD"}, "files": [], "token_estimate": 0}

`-s --format=%ct` is a caller asking for exactly one field. The shim ignores the format and hands back a manifest.

## Why it bites

Any script that parses scripted git output breaks, and it breaks silently. Concrete case from the cage:

    # wants a bare epoch like "1780329644"
    base_ct = run(["git", "show", "-s", "--format=%ct", ref]).stdout.strip()
    base_time = float(base_ct)   # float("{ ...json... }") raises ValueError

The ValueError gets swallowed, the function returns "no tickets resolved," and the PR-draft feature quietly emits the wrong list. No crash, just wrong output. The same failure hits anything parsing `git log --pretty=...` or `git diff --stat`.

## Fix

Make the shim a strict superset of git. Never change the output of a standard subcommand a script relies on.

- Pass through verbatim whenever the call carries an explicit `--format=`, `--pretty=`, `-s`, `--porcelain`, `--stat`, or `-z`.
- Emit the manifest JSON only for the bare agent-facing forms (`git show <ref>`, `git diff <a>..<b>`), or behind an explicit opt-in: a `git prism ...` subcommand or a `GIT_PRISM=1` env flag.

If a human or a script typed a format string, they want that format. The structured JSON is for callers who opted into git-prism, not for everyone who runs `git`.

## Nice-to-have: put commit data in the manifest

Today it's `{metadata, files: [], token_estimate}` with no commit-level fields. For `git show <commit>`, a structured consumer wants:

- `commit`: `sha` + `short_sha`, `parents` (merge-aware), `author` / `committer` with `name`, `email`, and both `date_iso` and `date_epoch` (the epoch is the `%ct` scalar), `subject` and `body` split apart.
- `files[]`: `path`, `change_type` (A/M/D/R/C), `old_path` for renames, `additions`, `deletions`, `is_binary`.
[...]

PR #339, opened 2026-06-01
fix(git): peel annotated tags to commits in ref resolution (#337)

### LINKED ISSUE #337: Annotated tags not peeled to commits: get_change_manifest / shim error on tag..tag ranges
## Summary 

`get_change_manifest` and the PATH shim's `git diff`/`git log` interception fail when `base_ref`/`head_ref` is an annotated tag. Ref resolution reads the tag object but never peels it to the underlying commit, so the diff/history pipeline gets a tag where it expects a commit. This breaks diffing two releases (e.g. `v0.8.0..v0.9.0`), which means git-prism can't analyze its own tagged releases.

## Repro

In a clone with annotated release tags:

```shell
# MCP tool
get_change_manifest(repo_path=".", base_ref="v0.8.0", head_ref="v0.9.0")

# PATH shim, as a detected agent
git diff v0.8.0..v0.9.0
git log  v0.8.0..v0.9.0
```

## Actual

```shell
git error: failed to read object: Object named 6ab58c91ee3f994a1a5fafd4b2185390c3b61a83
was supposed to be of kind commit, but was kind tag.
```

Separately, `git show <annotated-tag>` is intercepted but returns `"files": []` (silently wrong) instead of erroring.

## Expected 

The same manifest that real `git` and commit/branch refs produce. Annotated tags should be transparently peeled to their target commit.

## Workaround (confirmed)

Peeling to a commit works:

```shell
get_change_manifest(base_ref="v0.8.0^{commit}", head_ref="v0.9.0^{commit}")   # -> full 52-file manifest
git diff 'v0.8.0^{}..v0.9.0^{}'                                               # -> ok
```

So it's specifically un-peeled annotated tag objects. Lightweight refs, branches, and raw SHAs are fine.

## Likely location 

Ref resolution, probably `src/git/refs.rs`. Whatever resolves a ref to an object should peel tag objects (follow `tag.target` / `peel_to_commit`) before handing to the diff/history pipeline. The git show path needs the same peel (currently yields empty files).

## Environment 

git-prism 0.9.0 (Homebrew, macOS arm64). Reproduces on both surfaces (MCP tool + PATH shim), which share the ref-resolution path.

## Suggested fix

Peel annotated tags to commits during ref resolution, and add a regression test covering a `tag..tag` manifest built from an annotated tag (e.g. under `tests/`).

mikelane:
## Summary

Fixes #337. `get_change_manifest`, `get_commit_history`, `get_file_snapshots`, and the PATH shim's `git diff`/`git log`/`git show` interception all failed when a ref was an **annotated tag** — so git-prism couldn't analyze its own tagged releases (e.g. `v0.8.0..v0.9.0`).

## Root cause

`RepoReader::peel_to_commit` (`src/git/reader.rs`) resolved a ref and then called `object.try_into_commit()`. For an annotated tag, `rev_parse_single` resolves to the **tag object**, and `try_into_commit()` errors:

```
git error: failed to read object: <oid> was supposed to be of kind commit, but was kind tag.
```

Lightweight tags, branches, and raw SHAs were unaffected (they resolve directly to commits).

## Fix

One-line root fix: `try_into_commit()` → `peel_to_commit()`, which follows the tag → … → commit chain before converting. It's a no-op for objects that are already commits, so branches/SHAs/lightweight tags behave exactly as before. The `GitError::ReadObject` error contract is preserved verbatim for genuinely unresolvable refs (e.g. a tag pointing at a tree/blob still errors cleanly — no panic, no silent empty result).

Because both the MCP tools and the PATH shim funnel through `peel_to_commit`, this single change fixes every surface.

## Tests

- `it_peels_annotated_tag_to_commit` (RED→GREEN) + triangulation (lightweight tag, tag/branch SHA equivalence).
- Adversarial-QA probes: nested annotated-tag chains (`tag -a v2 v1`), tag→tree and tag→blob clean-error contracts, raw-SHA regression guard, annotated `tag..tag` range equals the commit..commit range and is non-empty, mixed tag..branch range.
- A manifest-path test exercising peel-through-the-shim with non-empty output.

## Out of scope / follow-up

The shim's bare `git show <ref>` returns an empty `files` array (it passes an empty `paths` slice to `build_snapshots`) — this predates this PR and affects all ref types equally, so it is **not** a peel regression. It is fixed in the parallel PR for #338 (which rewrites `git show` to use the manifest pipeline).

## Verification

```
cargo fmt --check        # clean
cargo clippy --all-targets -- -D warnings   # 0 errors
cargo test               # 939 passed
```

### DIFF
--- a/src/shim/handlers.rs
+++ b/src/shim/handlers.rs
@@ -365,4 +365,34 @@ mod tests {
             "expected 'functions' key in function_context output, got: {json}"
         );
     }
+
+    #[test]
+    fn it_handles_show_snapshot_for_annotated_tag_without_error() {
+        // Before the peel fix, `git show <annotated-tag>` would panic or exit
+        // non-zero because peel_to_commit failed with "was kind tag". This test
+        // confirms the handler completes successfully (exit SUCCESS, valid JSON
+        // with a "files" array) after the fix.
+        let (_dir, path) = init_repo_with_two_commits();
+
+        Command::new("git")
+            .args(["tag", "-a", "v1.0", "-m", "release v1.0"])
+            .current_dir(&path)
+            .output()
+            .unwrap();
+
+        let classification = Classification::ShowSnapshot { sha: "v1.0" };
+        let mut out = Vec::new();
+        let code = handle(&classification, &path, &mut out);
+        assert_eq!(
+            code,
+            ExitCode::SUCCESS,
+            "handle should exit SUCCESS for annotated tag, got non-zero"
+        );
+
+        let json: serde_json::Value = serde_json::from_slice(&out).unwrap();
+        assert!(
+            json.get("files").and_then(|f| f.as_array()).is_some(),
+            "expected 'files' array in show output for annotated tag, got: {json}"
+        );
+    }
 }
--- a/src/shim/handlers.rs
+++ b/src/shim/handlers.rs
@@ -455,4 +455,60 @@ mod tests {
             "annotated-tag show must produce the same files as target-sha show"
         );
     }
+
+    /// The `Classification::Manifest` path peels BOTH ends of a range.  This
+    /// test creates two annotated tags on consecutive commits and asserts that
+    /// the manifest over `v0.1..v0.2` returns a non-empty `files` array that
+    /// contains the file added in the second commit.  A "was kind tag" peel
+    /// failure would surface here as either an error exit or an empty file list.
+    #[test]
+    fn qa_manifest_over_annotated_tag_range_returns_changed_files() {
+        let (dir, path) = init_repo_with_two_commits();
+
+        let git = |args: &[&str]| {
+            Command::new("git")
+                .args(args)
+                .current_dir(&path)
+                .output()
+                .unwrap()
+        };
+
+        // v0.1 tags the first commit (HEAD~1 after init_repo_with_two_commits)
+        git(&["tag", "-a", "v0.1", "-m", "v0.1 release", "HEAD~1"]);
+        // v0.2 tags the current HEAD (second commit, which added world.txt)
+        git(&["tag", "-a", "v0.2", "-m", "v0.2 release", "HEAD"]);
+
+        let classification = Classification::Manifest {
+            range: "v0.1..v0.2",
+        };
+        let mut out = Vec::new();
+        let code = handle(&classification, &path, &mut out);
+        assert_eq!(
+            code,
+            ExitCode::SUCCESS,
+            "manifest over annotated tag range must exit SUCCESS"
+        );
+
+        let json: serde_json::Value = serde_json::from_slice(&out).unwrap();
+        let files = json
+            .get("files")
+            .and_then(|f| f.as_array())
+            .expect("expected 'files' array in manifest output");
+
+        assert!(
+            !files.is_empty(),
+            "manifest over v0.1..v0.2 must be non-empty (expected world.txt)"
+        );
+
+        let paths: Vec<&str> = files
+            .iter()
+            .filter_map(|f| f.get("path").and_then(|p| p.as_str()))
+            .collect();
+        assert!(
+            paths.contains(&"world.txt"),
+            "expected world.txt in manifest files, got: {paths:?}"
+        );
+
+        drop(dir);
+    }
 }
--- a/src/shim/handlers.rs
+++ b/src/shim/handlers.rs
@@ -395,4 +395,64 @@ mod tests {
             "expected 'files' array in show output for annotated tag, got: {json}"
         );
     }
+
+    // ===== ADVERSARIAL QA PROBES (issue #337 pen-test) =====
+
+    // SUSPICIOUS (pre-existing, NOT a #337 regression): `git show <ref>` via the
+    // shim always returns files: [] because handle_show_snapshot calls
+    // build_snapshots with an empty `paths` slice (&[]). build_snapshots only
+    // processes files explicitly listed in `paths`; it never enumerates the
+    // changed files of the range. This is true for annotated tags, lightweight
+    // tags, AND raw SHAs identically — the empty `&[]` predates this PR
+    // (origin/main:src/shim/handlers.rs:134). The peel change is therefore
+    // correct and unaffected here; the dev's
+    // it_handles_show_snapshot_for_annotated_tag_without_error test passes only
+    // because the snapshot is empty for everything, which masks whether the
+    // peel actually surfaces the right commit's content. Reported as SUSPICIOUS,
+    // not blocked, because it is out of scope for the #337 peel fix.
+
+    /// Regression guard within #337 scope: `git show <annotated-tag>` and
+    /// `git show <target-sha>` produce IDENTICAL output (both empty today, but
+    /// must not diverge — peeling must map the tag to the same range the SHA
+    /// produces). This stays green and protects the equivalence the peel change
+    /// is supposed to guarantee.
+    #[test]
+    fn qa_show_annotated_tag_output_equals_target_sha_output() {
+        let (_dir, path) = init_repo_with_two_commits();
+        Command::new("git")
+            .args(["tag", "-a", "v1.0", "-m", "release v1.0"])
+            .current_dir(&path)
+            .output()
+            .unwrap();
+        let target_sha = head_sha(&path);
+
+        let mut out_tag = Vec::new();
+        assert_eq!(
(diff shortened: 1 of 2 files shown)


mikelane (commented):
## PR Review — fix(git): peel annotated tags to commits in ref resolution (#337)

**Verdict: APPROVE** — 0 CRITICAL, 0 HIGH, 1 MEDIUM, 1 LOW

---

No bugs confirmed through test execution. All 948 tests pass. The production change is mechanically correct: `peel_to_commit()` on a gix object follows the tag→commit chain before conversion and is a documented no-op for objects already of kind commit. The adversarial QA suite exercises the exact failure modes (tag→tree, tag→blob, nested chains) and each surfaces a clean `GitError::ReadObject` — no panics, no silent empties.

---

### Phase 1 — Quality Tools

- `cargo fmt --check`: clean
- `cargo clippy --all-targets -- -D warnings`: 0 errors (3 pre-existing warnings, none in changed files)
- `cargo test`: 948 passed, 0 failed

---

No changes to `.pre-commit-config.yaml`, `lefthook.yml`, `pyproject.toml`, or any lint/mypy config. No suppressions added.

---

### Phase 3 — Code Standards

The production fix in `src/git/reader.rs` is idiomatic. The UFCS→method refactor (`Self::resolve_ref_with_fallback(self, ..) → self.resolve_ref_with_fallback(..)`) is the correct Rust style. Inline format captures (`format!("{file_path}" ..)`) are the modern Rust 2021 idiom. No bare `unwrap()` in production code paths. All public APIs retain their type signatures and error contracts.

---

The comment block in `src/shim/handlers.rs` starting "SUSPICIOUS (pre-existing, NOT a #337 regression)" is an unusually long in-code analysis block explaining WHY the `files: []` behavior exists and is out of scope. This crosses from "explains why" into "explains what and defends a decision." It belongs in the PR body (where it already lives) or a follow-up issue, not inline. This is a LOW finding — the comment is accurate and harmless, but it will confuse future readers who encounter it without the PR context.

**LOW** — `src/shim/handlers.rs` ~line 430: The multi-line SUSPICIOUS comment block documents a pre-existing limitation correctly but exceeds what belongs in source. Consider replacing with a one-line reference: `// NOTE: empty paths slice → empty files array; pre-existing, tracked in #338.`

---

### Phase 5 — Test Quality

Excellent test suite. Specific praise:

- **All tests go through the public API** (`resolve_commit`, `diff_commits`, `handle`). No `._method()` access.
- **Naming is precise**: `it_peels_annotated_tag_to_commit`, `qa_nested_annotated_tag_chain_resolves_to_commit`, etc. Not a single "should" in sight.
- **Assertions are real**: every test asserts specific field values, error variant matches, or non-empty collection membership — zero vacuous `assert!(true)`.
- **Triangulation is present**: the three happy-path tests (annotated tag, lightweight tag, SHA equivalence) deliberately force generalization, not just the one red→green case.
- **Error variant assertions** (`matches!(result.unwrap_err(), GitError::ReadObject(_))`) correctly pin the error contract without over-specifying the message string.

One MEDIUM finding: `it_handles_show_snapshot_for_annotated_tag_without_error` asserts `files` array presence but not that it is non-empty. Given the acknowledged pre-existing behavior (empty paths slice → empty files), the test currently passes vacuously for the same reason a SHA would pass — it cannot distinguish "peel succeeded and returned data" from "peel succeeded and returned nothing". The comment in `qa_show_annotated_tag_output_equals_target_sha_output` correctly documents this. The stronger test is `qa_manifest_over_annotated_tag_range_returns_changed_files` (which does assert non-empty and specific file names) — that test correctly guards the regression. The weaker test is not harmful but does not add signal the stronger test does not already provide.

**MEDIUM** — `src/shim/handlers.rs` ~line 406: `it_handles_show_snapshot_for_annotated_tag_without_error` cannot distinguish a successful peel with content from a successful peel with empty content. The `qa_manifest_over_annotated_tag_range_returns_changed_files` test provides the stronger guard. Consider adding a comment to the weaker test noting it is a smoke test for exit code / JSON shape, not content correctness.

---

### Phase 6 — Architecture

The single-point-of-fix design is correct. `peel_to_commit` is the sole entry point that all callers (10 production callers across `diff.rs` and `reader.rs`) flow through. The fix is in exactly the right place — not scattered across call sites.

No SOLID violations. No speculative abstraction. No god class growth (reader.rs grows by test-only lines only in the test module).

---

No security concerns. No shell-out to git in production code paths — all new production code uses `gix` exclusively. Test fixtures use `Command::new("git")` which is explicitly permitted by CLAUDE.md. No hardcoded secrets, no new serialization surfaces, no injection vectors.

The `peel_to_commit()` call has the same O(chain-depth) complexity as `try_into_commit()` for a commit object (one object read), plus one additional object fetch per tag in the chain for annotated tags. For real-world tag chains (depth ≤ 2), this is negligible.

---

Divergence point: `12448d28605792ecb3df491e32d487bd8199074b` (same as `origin/main` HEAD). The branch was cut from the exact current main — zero sibling PRs landed between branch creation and this review. No regression risk.

Deletion volume: 6 lines removed (all from the production function body being refactored). 387 lines added (test-only). Net additions >> deletions on a bugfix PR. No regression signal.

---

### Praise

The commit structure is exemplary: fix commit → adversarial QA probes commit → church nits + shim regression guard commit. Each commit is self-contained and the message explains the why. The PR body is accurate, concise, and correctly scopes what is and is not fixed. The "out of scope / follow-up" section explicitly calls out the pre-existing `git show` empty-files behavior rather than silently working around it — exactly the right transparency.

The `qa_manifest_over_annotated_tag_range_returns_changed_files` test is the star of the suite: it exercises `Classification::Manifest` end-to-end through the shim handler with two annotated tags on real commits, asserts non-empty output, and names the specific expected file. This is what a regression guard looks like.

---

**Decision: APPROVE.** The one-line production fix is correct, minimal, and well-understood. The test suite is adversarially thorough. Quality tools are clean. No blockers.

src/shim/handlers.rs
```diff
@@ -365,4 +365,150 @@ mod tests {
...
+                &mut out_tag
+            ),
+            ExitCode::SUCCESS
+        );
+        let mut out_sha = Vec::new();
+        assert_eq!(
+            handle(
+                &Classification::ShowSnapshot { sha: &target_sha },
+                &path,
+                &mut out_sha
```
mikelane, line 443: **LOW** — This multi-line SUSPICIOUS comment block is accurate and well-intentioned, but it exceeds what belongs in source code. It explains a pre-existing limitation that is already documented in the PR body and will be fixed in the parallel #338 PR. A future reader arriving at this code without PR context will find a 15-line inline analysis jarring next to the tests it annotates.

Consider replacing with a single reference line:
```rust
// NOTE: empty paths slice → empty files array; pre-existing behavior, tracked in #338.
```

Not a blocker — the comment is correct — but it will become stale once #338 lands.

src/shim/handlers.rs
```diff
@@ -365,4 +365,150 @@ mod tests {
...
+    }
+
+    // ===== ADVERSARIAL QA PROBES (issue #337 pen-test) =====
+
+    // SUSPICIOUS (pre-existing, NOT a #337 regression): `git show <ref>` via the
+    // shim always returns files: [] because handle_show_snapshot calls
+    // build_snapshots with an empty `paths` slice (&[]). build_snapshots only
+    // processes files explicitly listed in `paths`; it never enumerates the
+    // changed files of the range. This is true for annotated tags, lightweight
+    // tags, AND raw SHAs identically — the empty `&[]` predates this PR
```
mikelane, line 406: **MEDIUM** — This test asserts exit SUCCESS and a `files` key in the JSON, but cannot distinguish a successful peel-with-content from a successful peel-with-empty-content. Because `handle_show_snapshot` passes an empty `&[]` paths slice to `build_snapshots`, the `files` array is empty for annotated tags and raw SHAs alike — so this assertion would pass even if peeling were completely broken (as long as any valid JSON was returned).

The actual regression guard lives in `qa_manifest_over_annotated_tag_range_returns_changed_files` (further below), which asserts non-empty output with a specific expected filename through the manifest path.

This test is not harmful as a smoke test for exit code and JSON shape. Suggest adding a comment making the scope explicit:
```rust
// Smoke test: verifies exit SUCCESS and valid JSON shape only.
// Content correctness is guarded by qa_manifest_over_annotated_tag_range_returns_changed_files.
```

Not a blocker given the stronger test already exists.

PR #340, opened 2026-06-01
fix(shim): passthrough scripted-output git + enrich git show <ref> (#338)

### LINKED ISSUE #338: Scripted git output (`--format` / `--pretty` / `--stat`) should pass through to real git, not return manifest JSON

mikelane:
## Summary

Fixes #338. Two related fixes to the PATH shim, restoring its "strict superset of git" contract and making `git show` useful to structured consumers.

### 1. Scripted-output passthrough (the bug)

The shim returned change-manifest JSON for scripted format-string invocations — e.g. `git show -s --format=%ct HEAD` returned a JSON object instead of the bare epoch, silently breaking any script that parses git output (`git log --pretty=…`, `git diff --stat`, etc.).

The classifier now passes through to real git whenever a scripted-output flag is present: `--format`/`--pretty` (with or without value), `-s`/`--no-patch`, `--porcelain`, `--stat`, `-z`. The guard applies across `show`, `log`, `diff`, **and `blame`** (`git blame --porcelain` is scripted output a script parses, the same bug class). Plain agent-facing forms (`git show <ref>`, `git diff a..b`, `git blame <path>`) still intercept.

### 2. `git show <ref>` enrichment

Bare `git show <ref>` previously returned `{files: [], token_estimate: 0}` for every commit (it passed an empty paths slice to the snapshot builder). It now returns a real manifest via the change-manifest pipeline:

- `commit`: `sha`, `short_sha`, `parents` (merge-aware), `author`/`committer` (`name`, `email`, `date_iso`, `date_epoch`), `subject`, `body`.
- `files[]`: `path`, `old_path` (renames), `change_type` (A/M/D/R/C), `additions`, `deletions`, `is_binary`.
- top-level `diffstat`: `files_changed`, `insertions`, `deletions`.
- a computed `token_estimate` (no longer hardcoded).

Parentless **root commits** are handled via a new `diff_root_commit` (empty-tree base). Commit-time extraction is **fail-loud**: a malformed commit time propagates an error rather than silently emitting a `1970` timestamp.

### Builds on #337

This depends on the annotated-tag peel fix (#337, merged). Together, `git show <annotated-tag>` resolves the tag to its commit (#337) and returns the enriched manifest (#338) — verified byte-identical to `git show <target-sha>`. The parent ref is built from the peeled `commit.sha` so caret-parent resolution works for tag inputs.

## Tests

996 tests. Classifier passthrough coverage for every scripted flag across show/log/diff/blame; `split_commit_message` unit tests; an end-to-end multi-paragraph-body test; type-serialization unit tests (guard the JSON contract on Windows, where the unix-only e2e tests are skipped); and adversarial-QA regression suites covering root/merge/rename/binary/copy commits and the annotated-tag synergy.

No MCP tool surface changed — `ShowManifestResponse` is a shim-only response shape, not a registered MCP tool.

## Verification

```
cargo fmt --check        # clean
cargo clippy --all-targets -- -D warnings   # 0 errors
cargo test               # 996 passed
```

### DIFF
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -124,19 +124,8 @@ fn classify_show<'a>(rest: &[&'a str]) -> Classification<'a> {
 }
 
 /// Return `true` when the argv slice contains any flag that requests
-/// scripted/machine-readable text output from git.  When these flags are
-/// present the caller wants the raw git output — not a JSON manifest.
-///
-/// Flags covered:
-/// - `--format=<val>` and `--format` (pretty-print format string)
-/// - `--pretty=<val>` and `--pretty` (pretty-print preset or format string)
-/// - `-s` / `--no-patch` (suppress diff, often paired with `--format`)
-/// - `--porcelain` (machine-readable output for `status`, `blame`, etc.)
-/// - `--stat` (diffstat text output)
-/// - `-z` (NUL-terminated output)
-/// Return `true` when the argv slice contains any flag that requests
-/// scripted/machine-readable text output from git.  When these flags are
-/// present the caller wants the raw git output — not a JSON manifest.
+/// scripted/machine-readable text output from git (the caller wants raw git
+/// output, not a JSON manifest).
 ///
 /// Flags covered:
 /// - `--format=<val>` and `--format` (pretty-print format string)
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -134,6 +134,28 @@ fn classify_show<'a>(rest: &[&'a str]) -> Classification<'a> {
 /// - `--porcelain` (machine-readable output for `status`, `blame`, etc.)
 /// - `--stat` (diffstat text output)
 /// - `-z` (NUL-terminated output)
+/// Return `true` when the argv slice contains any flag that requests
+/// scripted/machine-readable text output from git.  When these flags are
+/// present the caller wants the raw git output — not a JSON manifest.
+///
+/// Flags covered:
+/// - `--format=<val>` and `--format` (pretty-print format string)
+/// - `--pretty=<val>` and `--pretty` (pretty-print preset or format string)
+/// - `-s` / `--no-patch` (suppress diff, often paired with `--format`)
+/// - `--porcelain` (machine-readable output for `status`, `blame`, etc.)
+/// - `--stat` (diffstat text output)
+/// - `-z` (NUL-terminated output)
+///
+/// # SAFETY: `--` separator not honoured
+///
+/// git treats `--` as the end of flags and the start of pathspecs, so
+/// `git diff a..b -- --stat` passes `--stat` as a filename, not a flag.
+/// This scanner does not track `--` and would incorrectly treat that
+/// `--stat` as a scripted-output flag, causing passthrough when interception
+/// was correct.  This is a theoretical input (no real caller names a file
+/// `--stat`) and adding a runtime guard would complicate the hot path for
+/// no practical benefit.  If it becomes real, scan for `--` first and stop
+/// checking beyond it.
 fn has_scripted_output_flag(tokens: &[&str]) -> bool {
     tokens.iter().any(|tok| {
         tok.starts_with("--format=")
--- a/src/shim/classify.rs
+++ b/src/shim/classify.rs
@@ -160,6 +160,10 @@ fn has_scripted_output_flag(tokens: &[&str]) -> bool {
 }
 
 fn classify_blame<'a>(rest: &[&'a str]) -> Classification<'a> {
+    // Scripted-output flags (e.g. --porcelain) mean the caller wants raw text.
+    if has_scripted_output_flag(rest) {
+        return Classification::Passthrough;
+    }
     // First non-flag argument is the path.
     if let Some(path) = rest.iter().copied().find(|t| !t.starts_with('-')) {
         return Classification::BlameSnapshot { path };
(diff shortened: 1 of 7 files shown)


src/shim/classify.rs
```diff
@@ -97,21 +101,69 @@ fn classify_log<'a>(rest: &[&'a str]) -> Classification<'a> {
...
+/// # SAFETY: `--` separator not honoured
+///
+/// git treats `--` as the end of flags and the start of pathspecs, so
+/// `git diff a..b -- --stat` passes `--stat` as a filename, not a flag.
+/// This scanner does not track `--` and would incorrectly treat that
+/// `--stat` as a scripted-output flag, causing passthrough when interception
+/// was correct.  This is a theoretical input (no real caller names a file
+/// `--stat`) and adding a runtime guard would complicate the hot path for
+/// no practical benefit.  If it becomes real, scan for `--` first and stop
+/// checking beyond it.
```
mikelane, line 147: **LOW — Intentional non-passthrough flags deserve a callout in the SAFETY comment**

`has_scripted_output_flag` correctly omits `--name-only` and `--name-status` (those still benefit from enrichment, as confirmed by the test at line 516). The existing `# SAFETY` doc comment only notes the `--` separator gap. Worth adding a one-liner like:

```
// Deliberately excluded: --name-only, --name-status — callers using these
// still want structured JSON, not raw text.
```

This prevents a future contributor from adding them by reflex. Non-blocking — correct as-is.

src/shim/handlers.rs
```diff
@@ -117,25 +122,173 @@ fn handle_function_context<W: Write>(
...
+/// Extract structured commit metadata for `sha` using gix (no shell out).
+fn build_show_commit_detail(repo_path: &Path, sha: &str) -> anyhow::Result<ShowCommitDetail> {
+    let reader =
+        RepoReader::open(repo_path).map_err(|e| anyhow::anyhow!("failed to open repo: {e}"))?;
+    let commit = reader
+        .peel_to_commit(sha)
+        .map_err(|e| anyhow::anyhow!("failed to resolve commit {sha}: {e}"))?;
+
+    let full_sha = commit.id().to_string();
+    let short_sha = full_sha.chars().take(8).collect::<String>();
```
mikelane, line 209: **LOW — `short_sha` length is hardcoded to 8, diverging from git's dynamic abbrev**

Real `git show` uses `core.abbrev` (default 7, auto-scales to avoid ambiguity in large repos). `take(8)` is a consistent, documented choice for this API surface, but callers relying on `short_sha` for display parity with `git log --oneline` will see 8 chars instead of the repo-adaptive length. Non-blocking for this PR — just worth a comment or a CLAUDE.md note that the value is intentionally fixed.

mikelane (commented):
## PR #340 Review — fix(shim): scripted-output passthrough + git show enrichment

### Summary

This is a well-executed, two-concern bugfix PR. The classifier passthrough (Phase 1) is correct and complete — all four intercept paths (show/log/diff/blame) guard with `has_scripted_output_flag` before any classification, the flag set is appropriate, and the tests cover every documented form including the exact bug-report case (`git show -s --format=%ct HEAD`). The show enrichment (Phase 2) correctly uses the manifest pipeline for diffstat (real per-file insertions/deletions, not net delta), handles root commits with `diff_root_commit` + `walk_tree_as_additions`, and uses the peeled SHA from `build_show_commit_detail` as the base ref — integrating the #337 annotated-tag fix correctly. `signature_to_commit_signature` is fallible rather than silently substituting epoch 0, which is the right call.

**Regression check (Phase 8 — mandatory):** Divergence point is `1749d505663d55eabb5fa7c3781a1ec61180e976` = current `main` tip. `src/git/reader.rs` was not modified by this PR. The `peel_to_commit` implementation and all annotated-tag QA tests from #337 are present unchanged in the branch. No regression.

**Layer hygiene:** `src/git/diff.rs` imports only gix types. `src/shim/handlers.rs` imports `crate::git::reader::RepoReader` (correct direction). `src/tools/types.rs` adds new serialisable types in the right layer. No violations.

**gix-only in production:** `build_show_commit_detail`, `diff_root_commit`, `walk_tree_as_additions` all use gix exclusively. The only `Command::new("git")` calls are in integration tests (`#[cfg(test)]` / separate test files), which is explicitly permitted by the conventions in CLAUDE.md.

### Blockers

None.

### Findings

**LOW (2)** — non-blocking, filed as tracking issues:

1. `src/shim/classify.rs:147` — The `# SAFETY` doc on `has_scripted_output_flag` notes the `--` separator gap but does not call out that `--name-only` / `--name-status` are *deliberately excluded* (they still benefit from enrichment). A future contributor may add them by reflex. A one-line exclusion rationale in the comment would prevent the ambiguity. (Inline comment posted.)

2. `src/shim/handlers.rs:209` — `short_sha` is hardcoded to 8 characters via `take(8)`, whereas real git uses `core.abbrev` (default 7, auto-scales for large repos). This is a documented intentional deviation, but the deviation is not called out in a comment or CLAUDE.md. Non-blocking — consistent choice, just undocumented. (Inline comment posted.)

### Praise

- The two-pass `token_estimate` construction in `handle_show_snapshot` (build with 0, then compute and write back) is an established pattern in this codebase and applied correctly here.
- `signature_to_commit_signature` propagating a real error instead of silently returning epoch 0 is the right design. The error message includes the email address for debuggability.
- Test decomposition in the pentest files is exemplary: one behavior per test, exact-value assertions (not shape checks), fixture repos built from scratch, no shared mutable state.
- The `file_change_to_show_entry` adapter cleanly isolates the `FileChange → ShowFileEntry` mapping at the layer boundary — a small function that makes the intent obvious.
- The peeled-SHA parent ref (`format!("{}^", commit.sha)` not `format!("{}^", sha)`) is the correct fix for the annotated-tag parent resolution problem, and the comment in the source explains exactly why.

### Decision

APPROVED. No blockers. 0 CRITICAL, 0 HIGH, 0 MEDIUM, 2 LOW. The two LOW findings are tracking items only.

PR #344, opened 2026-06-01
fix(shim): use stable path for shim target so it survives brew upgrade (#343)

### LINKED ISSUE #343: shim install bakes version-pinned Homebrew Cellar path; breaks on every brew upgrade
## Summary

`git-prism shim install` bakes the **fully-resolved, version-pinned** binary path into the shim symlink. For Homebrew installs this is the `…/Cellar/git-prism/<version>/bin/git-prism` path, which a `brew upgrade` invalidates — the shim then points at a stale (and eventually garbage-collected) Cellar directory, silently running the old version or dangling entirely.

## Repro

1. `brew install mikelane/tap/git-prism` (say, v0.9.0)
2. `git-prism shim install` → symlink `~/.local/share/git-prism/bin/git` → `/opt/homebrew/Cellar/git-prism/0.9.0/bin/git-prism`
3. `brew upgrade git-prism` (to v0.9.1)
4. The shim still points at `…/Cellar/git-prism/0.9.0/bin/git-prism`. Brew removes the 0.9.0 Cellar dir on cleanup → the shim is stale/dangling.

## Root cause

`src/hooks.rs`, `install_path_shim` (~line 211):

```rust
let current_exe = std::env::current_exe()?
    .canonicalize()   // resolves ALL symlinks
    .context("failed to canonicalize current executable path")?;
```

`canonicalize()` resolves the entire symlink chain. Homebrew installs `git-prism` as `/opt/homebrew/bin/git-prism` → `../Cellar/git-prism/<version>/bin/git-prism`. Canonicalizing yields the **version-pinned** Cellar path, which is exactly what must NOT be persisted, because it changes on every upgrade.

## Impact

- Homebrew users: the shim breaks on every `brew upgrade` until `shim install` is re-run. Worst case (old Cellar GC'd) the shim dangles and `git` resolution fails for any process that hits it.
- `cargo install` (`~/.cargo/bin/git-prism`) and from-source builds are unaffected — those paths are stable across upgrades.

## Proposed fix

Resolve the shim target to a **stable** path rather than the canonicalized Cellar path:

- If the resolved `current_exe` lives under a Homebrew Cellar (`…/Cellar/<name>/<version>/…`), point the symlink at the stable opt path instead — `$(brew --prefix)/bin/git-prism` (e.g. `/opt/homebrew/bin/git-prism`), which Homebrew keeps pointing at the current version. Detect via the path pattern or `brew --prefix`.
- Otherwise, use `current_exe()` **without** fully canonicalizing (make it absolute, but don't resolve the final symlink) so a stable invocation path is preserved.

Also:
- **`git-prism shim status`** should detect when the symlink target is dangling OR points into a versioned `Cellar/<name>/<version>/` directory, and warn the user to re-run `shim install` (and, post-fix, that the target is now stable).
- Consider a regression test that simulates the Homebrew layout (a `opt/bin/git-prism` symlink → `Cellar/x/1.0.0/bin/git-prism`) and asserts the shim targets the stable `opt` path, surviving a simulated upgrade (repoint the opt symlink to `Cellar/x/1.1.0/...`).

## Notes

- Found right after the v0.9.1 release. The fix ships in v0.9.2.
- The shim's runtime dispatch (`argv[0]` ends with `git`) is unaffected; this is purely about which path the *install* persists.

mikelane:
## Summary

Fixes #343. `git-prism shim install` baked the **version-pinned Homebrew Cellar path** into the shim symlink, so every `brew upgrade` left the shim pointing at a stale (eventually GC'd) Cellar directory.

## Root cause

`install_path_shim` used `std::env::current_exe()?.canonicalize()?`, and `canonicalize()` resolves the full symlink chain — turning `/opt/homebrew/bin/git-prism` into `…/Cellar/git-prism/<version>/bin/git-prism`. That version-pinned path was persisted as the shim target.

## Fix

New `stable_shim_target(canonical_exe)`: when the path matches the Homebrew Cellar shape — validated **positionally** (`Cellar` exactly 4 components from the end, `bin` second-to-last) — it rewrites to the stable `<prefix>/bin/<bin>` (which Homebrew keeps current across upgrades), and only if that target exists. Anything else (cargo installs, source builds, stray directories literally named `Cellar`) is returned unchanged. Derivation is pure `Path::components()` — no shelling out to `brew`.

`shim status` (and, for parity, the legacy `hooks status`) now warns when the stored shim target is version-pinned to a Cellar path or dangles, advising the user to re-run `git-prism shim install`.

## Tests

10 integration probes + boundary unit tests pinning the `n-5`/`n-2`/`n<6` positional arithmetic (with hand-built absolute paths, since `canonicalize()` would otherwise hide the boundary), an upgrade-survival test that resolves the *install-time* target after a simulated `brew upgrade`, filename-preservation, missing-stable-bin fallback, and status-advisory parity. 1015 tests pass; clippy/fmt clean.

## Scope

Source-only (`std::path`); no dependency or `Cargo.toml` changes. Ships in v0.9.2.

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -403,7 +403,7 @@ fn path_contains_cellar_component(path: &Path) -> bool {
 }
 
 /// The advisory message shown when a shim target is a version-pinned Cellar path.
-const CELLAR_STALENESS_ADVISORY: &str = "shim points at a version-pinned Homebrew Cellar path; \
+pub(crate) const CELLAR_STALENESS_ADVISORY: &str = "shim points at a version-pinned Homebrew Cellar path; \
      run `git-prism shim install` to update to a stable path \
      that survives `brew upgrade`";
 
@@ -738,10 +738,18 @@ mod tests {
         std::os::unix::fs::symlink(&cellar_exe, &link).unwrap();
 
         let status = path_shim_status(home);
-        assert!(
-            matches!(status, PathShimStatus::Installed { ref staleness_warning, .. } if staleness_warning.is_some()),
-            "expected Installed with a staleness warning for Cellar target; got {status:?}"
-        );
+        match status {
+            PathShimStatus::Installed {
+                staleness_warning: Some(ref w),
+                ..
+            } => {
+                assert_eq!(
+                    w, CELLAR_STALENESS_ADVISORY,
+                    "staleness warning must equal the canonical advisory string"
+                );
+            }
+            other => panic!("expected Installed with exact staleness warning; got {other:?}"),
+        }
     }
 
     #[test]
@@ -833,12 +841,16 @@ mod tests {
         std::fs::remove_file(&stable_exe).unwrap();
         std::os::unix::fs::symlink(&cellar_exe_110, &stable_exe).unwrap();
 
-        // The stable path still exists and now resolves to 1.1.0.
-        // Canonicalize both sides: on macOS /tmp is a symlink to /private/tmp,
-        // so TempDir paths need normalizing before comparison.
-        let resolved = std::fs::canonicalize(&stable_exe).unwrap();
+        // The install-time target (`target_before`) must re-resolve to the
+        // post-upgrade Cellar exe, proving a stable-path shim automatically
+        // follows `brew upgrade` without reinstalling.
+        // Canonicalize both sides: on macOS /tmp is a symlink to /private/tmp.
+        let resolved = std::fs::canonicalize(&target_before).unwrap();
         let expected = std::fs::canonicalize(&cellar_exe_110).unwrap();
-        assert_eq!(resolved, expected, "stable path should follow the upgrade");
+        assert_eq!(
+            resolved, expected,
+            "shim target chosen at install time must re-resolve to the post-upgrade version"
+        );
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


src/hooks.rs
```diff
@@ -603,6 +713,244 @@ mod tests {
...
+
+        let result = stable_shim_target(&cellar_exe);
+        assert_eq!(
+            result, cellar_exe,
+            "sbin (not bin) at n-2 must not be rewritten; got {result:?}"
+        );
+    }
+
+    #[test]
+    fn path_with_fewer_than_six_components_is_not_rewritten() {
```
mikelane, line 942: **LOW — misleading test comment / wrong boundary exercised**

The path `/a/Cellar/x/bin/git-prism` used here has **6** components (`/`, `a`, `Cellar`, `x`, `bin`, `git-prism`), not 5. The `n < 6` guard is never reached — the path is correctly returned unchanged because `Cellar` sits at index 1 (`n-5`), which is `a`, not `"Cellar"`. So the test passes for the right behaviour but exercises the **wrong guard** (positional Cellar check, not the length check).

The comment "5 components total" and the docstring "hits the n<6 guard" are inaccurate. A path that actually exercises the length guard would be something like `/Cellar/x/bin/git-prism` (5 components: `/`, `Cellar`, `x`, `bin`, `git-prism`).

The behaviour is correct either way (both guards prevent rewriting, and the existence check is the final safety net), so this is not a bug — just an inaccurate test name/comment that could mislead future readers relying on it to understand the boundary logic.

mikelane (commented):
## PR #344 — fix(shim): stable path for shim target so it survives brew upgrade

### Summary

This is a focused, well-engineered fix for a genuine footgun. `install_path_shim` previously called `canonicalize()` on the current exe, which on Homebrew resolves the symlink chain all the way to the version-pinned Cellar path. That path is GC'd by `brew upgrade`, silently breaking the shim. The fix is pure `std::path` arithmetic: strict positional validation of the Cellar layout (`Cellar` at n-5, `bin` at n-2, n≥6) with a mandatory existence guard before rewriting.

---

### Phase 0 — Bug investigation

One subtlety found on deep inspection: the test `path_with_fewer_than_six_components_is_not_rewritten` uses the path `/a/Cellar/x/bin/git-prism`, which has **6** components (`/`, `a`, `Cellar`, `x`, `bin`, `git-prism`) — not 5. The `n < 6` guard is never exercised; the path is returned unchanged because `parts[n-5]` = `"a"` ≠ `"Cellar"` (the positional check fires first). The comment says "5 components total" and "hits the n<6 guard" — both inaccurate. The behaviour is correct, but the `n < 6` branch is untested by any test in the suite. Not a bug (the existence guard is the final safety net, and in practice an absolute Unix path has at minimum 2 components), but the test comment is misleading. Filed as LOW.

---

### Phase 1 — Quality tools

- `cargo fmt --check`: **clean**
- `cargo clippy --all-targets -- -D warnings`: **clean** (3 pre-existing warnings from vendored `tree-sitter-kotlin` C scanner, not introduced by this PR)
- `cargo test`: **1015 passed, 0 failures**
- No `#[allow(...)]` suppressions on new code, no `// TODO` without tickets, no commented-out code in new files

---

`lefthook.yml` runs `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` on pre-push with no excludes or disabled checks. Config is clean.

---

### Phase 3 — Code standards

`stable_shim_target` is `pub(crate)` — correctly scoped, not leaking into the public API. Pure `std::path::Path::components()` — no shelling to `brew`, no `std::process::Command`, no external dependency. `path_contains_cellar_component` is a private helper used only by the status-warning path. `CELLAR_STALENESS_ADVISORY` is `pub(crate)` const so tests can assert on the exact string. No bare `unwrap()` in production code.

---

Comments throughout explain WHY (Homebrew Cellar GC behaviour, positional validation rationale, existence guard semantics). The `stable_shim_target` doc block is excellent: it explains the Cellar layout, the strictness of the validation, and what falls back unchanged. No WHAT-comments in production code.

---

### Phase 5 — Test quality

The test suite is genuinely adversarial. Highlights:

- **Integration (black-box) tests** in `tests/shim_stable_path_pentest.rs` copy the actual compiled binary to a synthetic on-disk layout and invoke `shim install` as a subprocess — the only correct way to exercise the `current_exe()` branch without mocking the OS.
- PENTEST A1 (stray `Cellar` dir in non-Homebrew path) and A2 (degenerate Cellar shape) would have caught any naive "first `Cellar` component" implementation.
- The upgrade-survival test correctly uses a symlink for the stable `<prefix>/bin/<bin>` (as Homebrew actually does) and resolves both sides with `canonicalize()` to handle the macOS `/tmp → /private/tmp` indirection.
- All tests are hermetic (TempDir), deterministic, and test through public API / CLI only.
- **LOW finding**: `path_with_fewer_than_six_components_is_not_rewritten` in `hooks.rs` has an inaccurate comment claiming to exercise the `n < 6` guard. See inline comment.

---

### Phase 6 — Architecture

- `stable_shim_target` is pure path arithmetic with no I/O side effects except the existence check — easily testable and composable.
- `install_path_shim` delegates target resolution to `stable_shim_target` in a single call.
- The two Cellar-detection concerns (install-time rewriting vs. status-time advisory) use appropriately different logic: `stable_shim_target` uses strict positional validation; `path_contains_cellar_component` uses a simpler component scan for the advisory (where false positives just add a warning, not a wrong rewrite).
- No speculative abstractions, no YAGNI violations, no SRP violations.

---

### Phase 7 — Security

- No shell injection: all Cellar detection is pure `Path::components()` arithmetic.
- No symlink following in the detection logic (only an existence check on the proposed stable path — safe; it either exists or falls back).
- `uninstall_path_shim` guards against deleting a regular file and only removes symlinks pointing at the current exe — correct ownership check.
- No hardcoded secrets, no SQL, no deserialization.

---

### Phase 8 — Regression vs. main

Branch diverges at exactly `4eff752` (v0.9.1 = current HEAD of main). Main has not moved since divergence — zero sibling-PR regression risk. Deletions: 7 lines removed (small match-arm updates in `main.rs` and `shim_cmd.rs` to destructure the new `staleness_warning` field on `PathShimStatus::Installed`). These are correct structural updates.

---

### Findings

| Severity | Count | Description |
|----------|-------|-------------|
| CRITICAL | 0 | — |
| HIGH | 0 | — |
| MEDIUM | 0 | — |
| LOW | 1 | `path_with_fewer_than_six_components_is_not_rewritten`: comment claims to test `n < 6` guard; actually exercises the positional guard. Misleading, not wrong. |

**Blockers:** none.

---

### Praise

- The positional validation strategy (checking exact component offsets from the end) is precisely what prevents the stray-`Cellar`-directory footgun. A naive "find first Cellar component" approach would have been exploitable.
- The existence guard is correct defence-in-depth: even if the positional check had a flaw, the shim cannot point at a phantom path.
- `CELLAR_STALENESS_ADVISORY` as a `pub(crate)` const shared between production and test code is a clean detail — tests cannot drift from what the user actually sees.
- The `BrokenLink` variant correctly embeds the Cellar reinstall advisory so the dangling-target case (post-upgrade GC) surfaces the remedy, not just the broken path.

---

### Decision: APPROVED

The fix is correct, the positional validation is strict, the tests are genuinely adversarial, and the code is clean. The one LOW finding (misleading test comment) is worth fixing in a follow-up but does not block merge.

Issue #346, opened 2026-06-03
Shim path emits no telemetry from reviews (v0.9.2): gh symlink not installed, shim never inits OTel, pr-diff needs local objects


mikelane:
## Summary
The shim is the intended integration path — agents run normal `git` / `gh`, git-prism intercepts transparently, and they never call the MCP tools. Today that path delivers **no telemetry** (and can't transparently serve `gh pr diff`) for three independent reasons. The MCP `serve` path is healthy, so the OTLP pipeline itself is fine.

## Blockers (all in git-prism, none in the agents/skills)

- [ ] **1 — `git-prism shim install` never creates the `gh` symlink.** gh interception shipped in #333 (v0.9.0), but install only makes `git` (`hooks.rs: PATH_SHIM_LINK_NAME = "git"`). So `gh pr diff` hits the real `gh` unless you `ln -s` a `gh` symlink by hand (the v0.9.0 notes say to). Fix: `shim install`/`status`/`uninstall` should create/manage `gh` too.
- [ ] **2 — shim mode never initializes or flushes telemetry.** In `src/main.rs`, the `basename == "git" || "gh"` branch returns `run_shim(...)` and never calls `telemetry::init()`; that runs only on the `serve` path (`server.rs:869`), which also holds the `TelemetryGuard` that flushes on drop. So in shim mode the global meter is a no-op, the `record_shim_*` counters (#303) write nothing, nothing flushes, and `GIT_PRISM_OTLP_ENDPOINT` is never read. Fix: in shim mode call `telemetry::init()` when the endpoint is set AND `force_flush()` before exit — the `PeriodicReader` won't fire in a sub-second process, so `Drop` alone isn't enough.
- [ ] **3 — `gh pr diff <N>` interception requires the PR commits locally.** The handler resolves base/head SHAs via `gh pr view`, then builds the manifest from local git objects, erroring if the head SHA isn't present. Real `gh pr diff` streams from the API and needs nothing local. Review skills run `gh pr diff <N>` cold (no checkout), so on any PR whose head isn't fetched it errors instead of returning data. Fix: fetch the missing SHAs in the handler, or document `gh pr checkout` first.

## Evidence

Shim emits no telemetry (installed 0.9.2, git shim first on PATH, endpoint set):

    git -C <repo> diff HEAD~1 HEAD            # normal diff, exit 0
    GIT_PRISM_SHADOW_SAMPLE_PCT=100 git ...   # still no metrics
[...]

PR #351, opened 2026-06-03
fix(shim): emit telemetry from shim mode — init + bounded force-flush (#348)

### LINKED ISSUE #348: fix(shim): initialize and force-flush telemetry in shim mode; default service version to crate version
## Summary

Part of #346 (blocker 2 + version-label note). Shim mode never initializes or flushes telemetry. In `src/main.rs` the `basename == "git" || "gh"` branch returns `run_shim(...)` (`main.rs:272-280`) and never calls `telemetry::init()` — that runs only on the `serve` path, which also holds the `TelemetryGuard` that flushes on drop. So in shim mode the global meter is a no-op, the `record_shim_*` counters (#303) write nothing, and `GIT_PRISM_OTLP_ENDPOINT` is never read.

Additionally: a sub-second shim process never triggers the `PeriodicReader` (`telemetry.rs:191`), so `Drop` alone won't flush. A `force_flush()` before exit is required — and `TelemetryGuard` has no such method today.

## Scope

- Shim dispatch calls `telemetry::init()` when `GIT_PRISM_OTLP_ENDPOINT` is set; holds the guard for the process lifetime.
- Add `TelemetryGuard::force_flush()` and call it before the shim exits (covering both the passthrough/exec path where reachable and the structured-output path).
- Default `GIT_PRISM_SERVICE_VERSION` to `env!("CARGO_PKG_VERSION")` when unset — fixes the 0.9.0-vs-0.9.2 mislabel. Applies to **both** serve and shim paths.

## Fix locations

- `src/main.rs` — shim dispatch: init + hold guard + force_flush before exit.
- `src/telemetry.rs` — add `force_flush` method; default service version when env unset.

## Acceptance criteria

- [ ] shim path initializes telemetry when endpoint is set
- [ ] no-op (zero cost, no stderr) when endpoint unset
- [ ] `force_flush` invoked before shim exit; metrics reach the exporter in a sub-second process
- [ ] service version defaults to crate version when `GIT_PRISM_SERVICE_VERSION` is unset (serve + shim)
- [ ] TDD: unit tests for init gating, force_flush, and version default

## Notes

- TDD mandatory. The exec-replacement passthrough path never returns — ensure flush happens before `exec`/exit on every path that emitted metrics.
- Disjoint files from the other two #346 PRs.

mikelane:
## Summary

Closes #348 (blocker 2 of #346, plus the version-label note).

Shim mode emitted **no telemetry**: the `git || gh` dispatch branch in `main.rs` returned `run_shim(...)` and never called `telemetry::init()` (which runs only on the `serve` path). So the global meter was a no-op, the `record_shim_*` counters wrote nothing, and `GIT_PRISM_OTLP_ENDPOINT` was never read. A sub-second shim process also never triggers the `PeriodicReader`, so `Drop` alone can't flush — an explicit flush is required.

## What changed

- Shim dispatch calls `telemetry::init()` (via a banner-suppressed `init_quiet()`) when the endpoint is set, holds the guard, and flushes before exit.
- New `TelemetryGuard::force_flush_bounded(Duration)` — flushes on a background thread with an `mpsc` `recv_timeout`. The passthrough paths flush with a **300 ms** cap **before `execvp`** replaces the process, so a recorded metric is never dropped, and a down/unreachable collector can't add the full 5 s `EXPORT_TIMEOUT` to every intercepted git command.
- Service version / name default to `env!("CARGO_PKG_VERSION")` / the built-in name when the env var is unset or empty.

## Note for the reporter

The "telemetry mislabeled `0.9.0`" symptom in #346 is **not** a code bug — `telemetry.rs` already defaulted the version to the crate version. It comes from the MCP client registration passing `env: { GIT_PRISM_SERVICE_VERSION: "0.9.0" }` to `git-prism serve`. Drop that override (or set it to the current version) in the MCP config.

## Gauntlet

- **Adversarial QA (2 passes)**: found the passthrough metric-drop bug (recorded then `execvp` before flush) — fixed with a paired OTLP-capture test; verified stdout JSON contract stays clean and metric attributes are bounded enums (no PII / cardinality leak).
- **Church**: observability (added `init_quiet` so the success banner doesn't print on every git command; made the cwd-failure branch use the bounded flush like its siblings), test (version-default + double-flush + bounded-deadline coverage).
- 1033 tests pass; clippy `-D warnings` clean; fmt clean.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -282,7 +282,12 @@ async fn main() -> std::process::ExitCode {
         // structured-output path run_shim returns an ExitCode, and we
         // force_flush before returning so metrics reach the exporter even
         // though the PeriodicReader hasn't ticked yet.
-        let mut telemetry_guard = telemetry::init();
+        //
+        // Use `init_quiet`: the shim runs once per intercepted git/gh command,
+        // so the per-invocation "telemetry initialized" banner would flood the
+        // developer's shell with one stderr line per git call. Failure paths
+        // still log, so a misconfigured endpoint stays visible.
+        let mut telemetry_guard = telemetry::init_quiet();
         // run_shim either exec-replaces the process (passthrough, never returns)
         // or returns ExitCode after printing structured JSON or an error.
         // On passthrough paths force_flush is called inside run_shim before execvp.
--- a/src/main.rs
+++ b/src/main.rs
@@ -275,9 +275,19 @@ async fn main() -> std::process::ExitCode {
             env: &agent_detection::StdEnvSource,
             argv0: args.first().copied().unwrap_or("git"),
         };
+        // Initialize telemetry when an OTLP endpoint is configured. The guard
+        // must live until after run_shim returns — on the exec-replace
+        // (passthrough) path the process is replaced and Rust's drop glue
+        // never runs anyway, so no flush is needed there. On the
+        // structured-output path run_shim returns an ExitCode, and we
+        // force_flush before returning so metrics reach the exporter even
+        // though the PeriodicReader hasn't ticked yet.
+        let mut telemetry_guard = telemetry::init();
         // run_shim either exec-replaces the process (passthrough, never returns)
         // or returns ExitCode after printing structured JSON or an error.
-        return shim::run_shim(&args, &agent_detection::StdEnvSource, &exec);
+        let exit_code = shim::run_shim(&args, &agent_detection::StdEnvSource, &exec);
+        telemetry_guard.force_flush();
+        return exit_code;
     }
 
     if let Err(e) = run().await {
--- a/src/main.rs
+++ b/src/main.rs
@@ -285,7 +285,14 @@ async fn main() -> std::process::ExitCode {
         let mut telemetry_guard = telemetry::init();
         // run_shim either exec-replaces the process (passthrough, never returns)
         // or returns ExitCode after printing structured JSON or an error.
-        let exit_code = shim::run_shim(&args, &agent_detection::StdEnvSource, &exec);
+        // On passthrough paths force_flush is called inside run_shim before execvp.
+        // On the structured-output path the guard is flushed here after run_shim returns.
+        let exit_code = shim::run_shim(
+            &args,
+            &agent_detection::StdEnvSource,
+            &exec,
+            &mut telemetry_guard,
+        );
         telemetry_guard.force_flush();
         return exit_code;
     }
--- a/src/shim/mod.rs
+++ b/src/shim/mod.rs
@@ -12,11 +12,22 @@ pub(crate) mod shadow;
 
 use std::path::PathBuf;
 use std::process::ExitCode;
+use std::time::Duration;
 
 use crate::agent_detection::EnvSource;
 use crate::metrics::{ShimOutcome, ShimSubcommand};
 use crate::shim::classify::{Classification, classify};
 use crate::shim::real_git::RealGitExec;
+use crate::telemetry::TelemetryGuard;
+
+/// Flush deadline used on passthrough paths (before execvp).
+///
+/// A DOWN OTLP collector makes `force_flush()` block up to `EXPORT_TIMEOUT`
+/// (5 s).  That would stall the shell on every intercepted git command even
+/// when the collector is unreachable.  This cap bounds the wait; the flush
+/// thread is abandoned if it hasn't completed in time and will be torn down
+/// when the process image is replaced by execvp.
+const PASSTHROUGH_FLUSH_TIMEOUT: Duration = Duration::from_millis(300);
 
 /// Main entry point for shim mode.
 ///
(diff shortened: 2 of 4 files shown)


mikelane (commented):
## Review Summary

**VERDICT: APPROVE** — 0 critical, 0 high, 0 medium, 2 low. This is a tight, well-reasoned fix for a real metric-drop bug. The adversarial-QA tests are first-class citizens, the flush design is sound, and the stdout JSON contract is demonstrably clean.

---

## Phase 0 — Bug Identification

Wrote mental failing tests for each suspicion below. None produced a real failure.

**Suspicion 1: double-flush + Drop causes panic or double-free.**
`force_flush_bounded` *clones* the `Arc`-backed provider handles into the background thread — the originals remain in `self.meter_provider` / `self.tracer_provider`. `Drop` then calls `shutdown()` via `take()`, which moves the `Option` out. A second `Drop` on an already-`take()`n guard is a no-op. The `it_tolerates_double_force_flush_then_drop_on_active_guard` test covers exactly this sequence. Not a bug.

**Suspicion 2: flush thread outlives execvp / is abandoned with a live tx.**
When `recv_timeout` fires and the caller returns, `rx` is dropped. On the passthrough path `execvp` replaces the process image; the flush thread and its `tx` are silently torn down with the process. On normal exit the thread joins via OS process teardown. `tx.send(())` returns `Err(SendError)` if `rx` is already dropped — the `let _ =` discard is correct. Not a bug.

**Suspicion 3: mpsc channel never signals — deadlock.**
`tx` is captured by move into the thread. The thread calls `send(())` unconditionally after both flush attempts (success or logged error), then returns. If the flush thread panics before `send`, the thread stack unwinds, `tx` is dropped, and `recv_timeout` receives `Err(RecvTimeoutError::Disconnected)` — which the `let _ =` also discards. No deadlock path exists.

**Suspicion 4: stdout polluted by telemetry diagnostics.**
All `eprintln!` calls in `init_with_attacher` and `force_flush*` go to stderr. The structured-output path writes to `out_buf` then `stdout().write_all()`. The success banner is suppressed on the shim path via `init_quiet`. Verified by the `it_flushes_shim_metric_on_structured_path_control` test which asserts `out.status.success()` and checks stderr separately. Not a bug.

**Suspicion 5: flush genuinely happens on every passthrough branch.**
`src/shim/mod.rs` lines 67, 74, 83, 92 — every branch that calls `exec.passthrough()` is preceded by `telemetry_guard.force_flush_bounded(PASSTHROUGH_FLUSH_TIMEOUT)`. The `it_flushes_shim_invocation_metric_on_passthrough_before_process_replacement` pentest drives the loop-break path end-to-end and asserts an OTLP export arrives before the process is replaced. Not a bug.

**Suspicion 6: PII / unbounded cardinality in metric attributes.**
`record_shim_invocation` uses `ShimOutcome` (bounded enum), `record_shim_classification` uses `ShimSubcommand` (bounded enum), `record_shim_response_bytes` is a counter value (not an attribute). No file paths, ref names, repo identifiers, or free-form strings appear in metric attributes. Not a bug.

---

## Phase 1 — Quality Tools

- `cargo fmt --check`: **clean**
- `cargo clippy --all-targets -- -D warnings`: **clean** (only a pre-existing vendored C scanner warning in `vendor/tree-sitter-kotlin/src/scanner.c`, outside this PR's scope)
- `cargo test`: **1033 tests pass, 0 failed** — including both new pentest tests in `shim_passthrough_telemetry_pentest.rs`

---

## Phase 5 — Test Quality

**Strengths:**
- `it_flushes_shim_invocation_metric_on_passthrough_before_process_replacement` is a real end-to-end behavioral test. It drives the compiled binary via subprocess, binds a real TCP capture server, and asserts an OTLP export arrives. This is exactly the right level of test for a flush-before-exec property.
- `it_flushes_shim_metric_on_structured_path_control` is the correct control: it validates the harness works before attributing any failure to the production code path under test.
- `it_tolerates_double_force_flush_then_drop_on_active_guard` directly covers the shim's actual lifecycle (flush → exec → Drop unreachable, or flush → return → Drop). The assertion `guard.is_active()` after `force_flush` proves the borrow-not-move semantics are intentional and preserved.
- `it_force_flushes_bounded_active_guard_within_deadline` uses RFC 5737 TEST-NET-1 (`192.0.2.1`) — an unroutable address that causes connections to hang rather than immediately refuse — to exercise the bounded-wait path. This is the right choice over `localhost` (which would fast-fail and never exercise the timeout path).
- `ENV_MUTEX` pattern for serializing `set_var`/`remove_var` across tests is correct and necessary.

**LOW finding:** `spawn_capture_server` spawns a background thread with an `incoming()` iterator that runs until the process exits. The `TcpListener` is not guarded by a stop signal or `JoinHandle`. This is not a correctness issue for the current test structure (each test calls it once per process invocation), but is a potential flakiness seed if tests are ever parallelized with `--test-threads=N` and two tests race for ephemeral ports from a shared listener. Consider adding a `shutdown_tx: mpsc::Sender<()>` to signal the listener thread to stop, or at minimum a comment noting the intentional process-scoped lifetime.

---

## Phase 7 — Security and Performance

No injection surfaces. No hardcoded secrets. The OTLP endpoint URL is developer-controlled environment configuration, not network input. The 300 ms flush cap prevents a DOWN collector from becoming a latency DoS on every intercepted git command — exactly the right trade-off.

**LOW finding:** The 300 ms `PASSTHROUGH_FLUSH_TIMEOUT` is a module-level constant in `src/shim/mod.rs` that cannot be overridden without recompilation. For the current use case this is fine. If the shim is ever deployed in environments with higher-latency collectors (cross-region, VPN-tunneled), operators have no escape hatch short of rebuilding. A `GIT_PRISM_FLUSH_TIMEOUT_MS` env var fallback would be a nice-to-have, not a blocker.

---

## Phase 8 — Regression vs Main

- Divergence point: `0334c9d` = current `main` HEAD (v0.9.2 release). The branch is exactly 4 commits ahead of main with zero file overlap — no files were modified on main since divergence.
- 17 deletions on a PR adding 701 lines. The deletions are the old `run_shim` signature and minor refactors. No regression signal.
- **Zero regression risk.**

---

## Praise

The adversarial test design in `shim_passthrough_telemetry_pentest.rs` is excellent. Writing the test in RED (as a failing pentest documenting the exact bug) before writing the fix is the correct TDD sequence, and the comments in the test file preserve that history for future readers. Using RFC 5737 TEST-NET-1 for the bounded-deadline test is a precise choice — `localhost:4318` would fast-fail and never exercise the actual timeout path. The `ENV_MUTEX` pattern for env-var test isolation is the right solution for a process-global problem.

---

*Reviewed by pr-reviewer (adversarial, independent). Tool results: fmt clean, clippy -D warnings clean, 1033/1033 tests pass.*

PR #357, opened 2026-06-04
fix(shim): classify git <global-options> <subcommand> so gh/git -C diff isn't bypassed (#356)

### LINKED ISSUE #356: fix(shim): 'git -C <path> diff/log/show' bypasses interception and falls through to raw git
## Summary

The shim intercepts `git diff` / `git log` / `git show` / `git blame` for a ref range and returns a JSON manifest to agent callers. But when git's global `-C <path>` option precedes the subcommand (`git -C <path> diff <range>`), the shim falls through to **raw git** instead of producing the manifest.

## Repro (Homebrew git-prism 0.9.3, agent detected as ClaudeCode)

With `-C <path>` before the subcommand (from outside the repo):

```
~/.local/share/git-prism/bin/git -C /Users/mikelane/dev/git-prism diff HEAD~1..HEAD
# -> raw unified diff (passthrough to real git)
```

Same range, same repo, no `-C` (run from inside the repo):

```
cd /Users/mikelane/dev/git-prism
~/.local/share/git-prism/bin/git diff HEAD~1..HEAD
# -> { "metadata": { ..., "version": "0.9.3" }, "summary": { ... } }   (JSON manifest)
```

The only difference is the `-C <path>` prefix.

## Likely cause

The shim's argv classifier appears to look for the subcommand at a fixed position (e.g. `argv[1]`). With `-C <path>` (two tokens) ahead of `diff`, the subcommand isn't where it's expected, so the call is classified as passthrough. The same would likely affect other pre-subcommand global options that the agent/tooling commonly use: `-c <name=value>`, `--git-dir <path>`, `--work-tree <path>`, `--namespace <name>`.

## Impact

`git -C <path> <subcommand>` is extremely common in agent and tooling contexts (operating on a repo without `cd`-ing into it). Those calls silently return raw diffs instead of structured manifests — losing the token savings the shim exists to provide, and (if telemetry classification keys on the subcommand token) potentially mis-recording or under-recording the invocation.

## Proposed fix

Parse git's pre-subcommand global options before locating the subcommand: handle the value-taking ones (`-C <path>`, `-c <name=value>`, `--git-dir <path>`, `--work-tree <path>`, `--namespace <name>`) plus git's other documented global flags, then classify on the first non-option token. When `-C <path>` is present, honor it as the repo path used to build the manifest.

## Environment

git-prism 0.9.3 (Homebrew), macOS arm64, zsh 5.9.

mikelane:
## Summary

Closes #356.

Through the shim, `git -C <path> diff/log/show/blame <range>` fell through to **raw git** instead of returning the JSON manifest. The classifier took the subcommand as `argv[1]`, so with `-C <path>` ahead of it, `argv[1]` was `-C` (not the subcommand) and the call classified as `Passthrough`. `git -C <path>` is extremely common in agent/tooling contexts (operating on a repo without `cd`), so those calls silently lost the structured output the shim exists to provide.

## What changed

- **`src/shim/classify.rs`** — `skip_git_global_options` skips git's leading global options before locating the subcommand: value-taking (`-C`, `-c`, `--git-dir`, `--work-tree`, `--namespace`, `--super-prefix`, `--config-env`, `--exec-path`, in both `--opt value` and `--opt=value` forms) and valueless (`-p`/`--paginate`/`--no-pager`/`--bare`/…). An unrecognized leading `-` token falls through to `Passthrough` (safe default). `classify()` now returns `ClassifyResult { classification, repo_dir_overrides }`, where `repo_dir_overrides` carries all `-C` values in argv order.
- **`src/shim/mod.rs`** — `resolve_repo_path` folds the `-C` values with git's `cd`-semantics (an absolute value resets the base, a relative value joins onto it; `git -C a -C b` ⇒ repo `a/b`). `GIT_PRISM_REPO` stays the highest-precedence override.

## Gauntlet

- **Adversarial QA (2 passes):** pass 1 found that relative *stacked* `-C a -C b` kept only the last value and opened `./b` — silently returning a manifest for the **wrong repo** (exit 0, no warning); fixed with the cd-semantics fold. Pass 2 confirmed the original bug is dead, no shell is invoked (`-C '…; touch PWNED'` is an inert literal path), path traversal errors cleanly, and a nonexistent/non-repo `-C` exits 1 with an actionable message. 1081 tests.
- **Church:** naming (`dash_c` → `repo_dir_overrides`), +14 global-option edge tests, CHANGELOG copy.


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **`git -C <path> diff/log/show/blame` now routes through the shim correctly.** Previously the argv classifier looked for the subcommand at a fixed position (`argv[1]`), so leading git global options like `-C <path>`, `-c <name=value>`, `--git-dir <path>` caused the call to be classified as `Passthrough` and raw git output was returned instead of a JSON manifest. The classifier now skips all recognised git global options (both `--opt value` and `--opt=value` forms, and valueless flags like `--no-pager`) before locating the subcommand. When `-C <path>` is present, the path is used as the repository root for manifest building, matching git's own semantics (last `-C` wins). Unrecognised leading `-` tokens fall through to passthrough — the documented safe default. (#356)
```
mikelane: **MEDIUM — CHANGELOG description contradicts the implementation**

The entry reads: "When `-C <path>` is present, the path is used as the repository root for manifest building, matching git's own semantics (**last `-C` wins**)."

"Last `-C` wins" is the description of the **pre-fix** behaviour — it is precisely the bug this PR fixes. The actual implementation in `resolve_repo_path` folds every `-C` value in order with `cd`-semantics (absolute resets the base, relative joins), so `git -C a -C b` resolves to `<cwd>/a/b`, not `<cwd>/b`.

The regression guard test `relative_stacked_dash_c_must_match_git_cd_semantics` explicitly asserts this is NOT last-wins.

Suggested replacement: _"When one or more `-C <path>` values are present they are folded in order using git's `cd`-semantics: an absolute path resets the base; a relative path joins onto it. So `git -C a -C b` resolves the repo at `<cwd>/a/b`, matching real git exactly."_

This is a documentation-only issue — the code is correct.

mikelane (commented):
## Review: fix(shim): classify `git <global-options> <subcommand>` so `gh/git -C diff` isn't bypassed (#356)

### Summary

Solid fix for a real bypass bug. The implementation is correct, the classifier stays pure, resolution lives in the right layer, injection is impossible, and the adversarial regression guard (`relative_stacked_dash_c_must_match_git_cd_semantics`) definitively proves the pre-fix "last-`-C`-wins" silent-wrong-repo behaviour cannot return. All 1081 tests pass; `cargo clippy --all-targets -- -D warnings` is clean on Rust code (3 pre-existing C warnings from the vendored Kotlin tree-sitter scanner, not introduced here); `cargo fmt --check` passes.

### Phase 0 — Bug identification (adversarial probing)

Every edge case from the review brief was traced and tested:

- **`-C` as last token (no value):** `tokens.get(i + 1)` returns `None`, push is skipped, `i += 2` overshoots, loop exits → `(None, &[], vec![])` → Passthrough. No panic. Covered by `it_passes_through_when_dash_c_is_last_token_with_no_value`.
- **`--opt=value` vs `--opt value`:** Both forms handled for all 5 long-taking options. Single-token form advances by 1; two-token form by 2. Separate tests for each form.
- **Interleaved globals:** Sequential `i`-advancement processes all interleaved tokens correctly. Covered by `it_handles_interleaved_globals_around_dash_c` and `it_interleaves_multiple_dash_c_among_other_globals_preserving_order`.
- **Unknown leading flag → Passthrough with no stale `-C` leak:** Returns `(None, &[], vec![])` — the empty `vec![]` is deliberate; accumulated `-C` values are discarded so they cannot leak into a Passthrough result and mis-route a downstream handler. Covered by `it_passes_through_unknown_flag_after_valid_globals_discarding_overrides`.
- **Subcommand-lookalike values:** `-C log diff A..B` correctly treats `log` as the `-C` path value and `diff` as the subcommand. `-c diff log A..B` correctly treats `diff` as the `-c` config value and `log` as the subcommand.
- **`--exec-path` bare form:** Treated as valueless (advances by 1 only), not consuming the next token as a value. Correct.
- **`cd`-semantics fold in `resolve_repo_path`:** Absolute path resets the base; relative joins. `git -C a -C b` → `cwd/a/b`. `git -C /abs -C rel` → `/abs/rel`. `git -C /abs` → `/abs`. Correct.
- **`GIT_PRISM_REPO` precedence:** Checked first before any `-C` resolution. Correct.
- **Bad/nonexistent `-C` path:** No validation at the resolution layer — intentional. Path validity is gix's responsibility; an invalid path causes gix to fail cleanly, not a silent wrong-repo or a panic. Correct design.
- **No shell invocation, no injection:** `classify()` is pure (`#[cfg(no_io)]`-equivalent — no I/O calls). `resolve_repo_path` uses only `std::env::current_dir()` and `PathBuf::from()`. The `-C` value is an inert literal — no shell expansion, no command substitution.
- **Regression vs main:** Branch diverges directly from `a8090ac` (latest main). Zero deletions from main's file set. No regression risk.

### Phase 1 — Quality tools

- `cargo clippy --all-targets -- -D warnings`: **PASS** (0 Rust errors/warnings; 3 pre-existing C warnings from vendored `vendor/tree-sitter-kotlin/src/scanner.c`, not introduced by this PR).
- `cargo fmt --check`: **PASS**.
- `cargo test`: **PASS** — 1081 tests across 11 suites.

Comments are excellent throughout. The `SAFETY:` comment in `has_scripted_output_flag` explaining the `--` separator limitation is exactly the right level of WHY documentation. The pentest module doc comment precisely describes the three-repo fixture layout and which failure each assertion guards against.

### Phase 5 — Test quality

- All unit tests exercise `classify()` via public API only (through the `classification()` helper). No private method access.
- Integration pentest treats the binary as a black box — the only way to test end-to-end `-C` routing through a symlink.
- 88 unit tests for the classifier; 1 end-to-end regression guard for the stacked `-C` scenario.
- The three-way fixture (repos `a`, `a/b`, `b`) cleanly distinguishes all three failure modes: last-wins, first-only, and correct fold.

### Blockers

None. The single finding below is documentation-only and does not affect correctness.

### Findings

**MEDIUM** (`CHANGELOG.md:12`) — The CHANGELOG entry says "last `-C` wins" which is the description of the **pre-fix bug**, not the new behaviour. The implementation correctly folds all `-C` values with `cd`-semantics. See inline comment for suggested text. Does not block merge — the code is correct and the test suite proves it.

### Praise

- The design to return `vec![]` (not accumulated `repo_dir_overrides`) on unknown-flag Passthrough is subtle and correct. A stale `-C` must not leak into a Passthrough result. This was explicitly tested.
- Choosing to *not* validate path existence in `resolve_repo_path` is the right call — path-doesn't-exist is gix's error domain, not the classifier's.
- The three-repo fixture in `shim_git_c_bypass_pentest.rs` is a textbook regression guard. It fails for the original bug, the intermediate regression (first-only fold), and the correct fix each produce distinguishable output.
- `skip_git_global_options` staying private keeps the classifier's internal mechanics behind the stable `classify()` API.

### Decision

**APPROVED** — 0 CRITICAL, 0 HIGH, 1 MEDIUM (CHANGELOG text only, code is correct), 0 LOW. Solo-owner adversarial review complete.

mikelane (commented):
Re-review at 7ed9ccf (comment-only amendment to prior APPROVE at 1e72f48).

**Diff verification:** The delta from 1e72f48 to 7ed9ccf is exactly four comment lines added above the `if argv.len() < 2` guard in `src/shim/classify.rs`. No executable code changed.

**Skip justification soundness:** The annotation is correct. `classify_gh` requires `rest.get(1)` to be a numeric PR number — meaning argv length >= 4 (`["gh", "pr", "diff", "<N>"]`). A length-2 argv produces `rest = &[]`, and `rest.first() == None`, so the only reachable outcome is `Passthrough`. For git, `skip_git_global_options` returns `(None, &[], vec![])` when no subcommand token exists, which also routes to `Passthrough`. The mutant `<= 2` is genuinely unkillable — this is a domain invariant, not a coverage gap.

**Quality gates (all pass):**
- `cargo test`: 1048 tests, all passing
- `cargo clippy --all-targets -- -D warnings`: clean (vendored Kotlin C scanner warning is pre-existing, out of scope)
- `cargo fmt --check`: exit 0

**Prior APPROVE findings:** All previously reviewed concerns (0 critical, 0 high, 0 blockers) remain unaffected — no production code touched.

**Decision: APPROVED.** The skip annotation is accurate, well-justified, and uses the correct `cargo-mutants: skip` pragma with a human-readable explanation of the equivalence class. This is the right way to handle provably-unkillable mutants.

mikelane:
**Status: APPROVED and ready, held on an infrastructure issue (not code).**

At head `668dc31` (branch merged up to current main):
- **All required checks pass:** Test (macOS/Ubuntu/Windows), Clippy, Format, BDD Acceptance, Integrity, Check.
- pr-reviewer **APPROVE** (0 findings); full suite green locally (1100 tests).
- The **only** red is the advisory/non-required `Mutants (incremental)` check, which is killed by a **GitHub Actions runner shutdown at ~14 min on 4 consecutive runs** (`##[error]The runner has received a shutdown signal`). This smells like a GH Actions minutes/spending cap cancelling in-progress jobs — worth checking repo/org billing, since it will keep killing the mutants job on every PR.
- The one mutant the job managed to flag before being killed (`classify.rs` `argv.len() < 2` → `<= 2`) is **provably equivalent** (no git/gh argv of length 2 ever classifies to non-Passthrough), so it is unkillable by any test.

**To merge later:** either (a) re-run the mutants job once the runner/quota issue clears, then merge; or (b) merge ignoring the advisory check (admin or a fine-grained bypass), since every required gate is green.

**Note on the mutant skip:** the `// cargo-mutants: skip` comment added here did **not** suppress the mutant (it still reported MISSED) — cargo-mutants honours the `#[mutants::skip]` attribute or `exclude_re` in `.cargo/mutants.toml`, not line comments. If you want the equivalent mutant genuinely excluded, use one of those instead. (Moot while the runner keeps getting reclaimed.)

PR #358, opened 2026-06-04
fix(shim): write PATH block to .zshenv + .zshrc so the shim wins for agent shells (#355)

### LINKED ISSUE #355: fix(shim): shim install only writes .zprofile — agent shells (.zshenv) and .zshrc PATH re-ordering leave the shim un-prioritized
## Summary

`git-prism shim install` creates the `git`/`gh` symlinks and (on consent) appends `export PATH="$HOME/.local/share/git-prism/bin:$PATH"` to `~/.zprofile`. But `~/.zprofile` is read only by **login** shells. In a common real-world setup the shim never wins on PATH, so interception silently doesn't happen — even though `install` reports success and `shim status` shows the symlinks present.

This reintroduces the exact symptom #346 set out to fix (no interception / no telemetry from agent git/gh), but at the install layer: a user installs the shim, removes the legacy redirect hook per the v0.9.0 migration, and ends up with nothing intercepting.

## Two compounding gaps

1. **Agent shells read `.zshenv`, not `.zprofile`.** Claude Code's Bash tool runs a non-login, non-interactive `zsh -c`, which sources **only** `~/.zshenv`. The installer never writes `.zshenv`, so the shim dir is absent from the agent's PATH.
2. **`.zshrc` re-prepends Homebrew *after* `.zshenv`.** Even with the shim in `.zshenv` (sourced first), a typical `~/.zshrc` runs `eval "$(brew shellenv)"` / `export PATH="/opt/homebrew/bin:$PATH"` / cargo / nvm / etc. — all prepending **after** `.zshenv`, demoting the shim behind real `git`/`gh`. So any shell that sources `.zshrc` (interactive, and Claude Code's inherited env) resolves real git first.

## Repro (macOS arm64, zsh 5.9, Homebrew git-prism 0.9.3)

Immediately after `git-prism shim install` (which wrote only `~/.zprofile`):

```
zsh -lc  'which git'  ->  ~/.local/share/git-prism/bin/git   # login: .zprofile line works
zsh -c   'which git'  ->  /opt/homebrew/bin/git              # agent shell: .zshenv never written -> real git
zsh -ic  'which git'  ->  /opt/homebrew/bin/git              # interactive: .zshrc re-prepends brew -> real git
```

`git-prism agent-detect` correctly returns `{"agent":"ClaudeCode","signal":"AiAgent"}` — but the shim is never reached because it isn't first on PATH.

Manual fix that worked: add the prepend to `~/.zshenv` **and** append it to the end of `~/.zshrc` (after `brew shellenv` and friends). After that, all four modes (`-c`/`-ic`/`-lc`/`-lic`) resolve `git`/`gh` -> the shim, and `git diff <range>` returns the JSON manifest.

## Proposed fix

- Write the prepend to a file agent shells actually source. `~/.zshenv` covers every zsh invocation (login/non-login/interactive/non-interactive); for bash, `~/.bash_profile` + `~/.bashrc`.
- Because interactive rc files re-prepend other dirs afterward, also re-assert the shim **last** in the interactive rc (end of `~/.zshrc`/`~/.bashrc`), guarded to avoid duplicates:
  ```sh
  case ":$PATH:" in
    *":$HOME/.local/share/git-prism/bin:"*) ;;
    *) export PATH="$HOME/.local/share/git-prism/bin:$PATH" ;;
  esac
  ```
- At minimum, make `shim status` / the install output verify the shim dir is **first** on PATH for a non-login shell (e.g. probe `zsh -c 'echo $PATH'`) and warn explicitly when it isn't — rather than only reflecting the current (likely login) shell.

## Environment

git-prism 0.9.3 (Homebrew, `/opt/homebrew/bin/git-prism`). macOS arm64, zsh 5.9. Claude Code agent shell = non-login `zsh -c`.

mikelane:
## Summary

Closes #355.

`git-prism shim install` only wrote the PATH prepend to `~/.zprofile` (login shells only), so the primary target — agent shells (Claude Code's non-login `zsh -c`, which read only `~/.zshenv`) — never got the shim on PATH. And even with `.zshenv`, a typical `~/.zshrc` re-prepends Homebrew (`brew shellenv`) / cargo / nvm *after* it, demoting the shim behind real git/gh. Net: install reported success but nothing intercepted.

## What changed

- Install writes a **marker-delimited, idempotent** block (`# >>> git-prism shim >>>` … `# <<< git-prism shim <<<`) to **two** files per shell — `~/.zshenv` + the end of `~/.zshrc` (zsh), `~/.bash_profile` + `~/.bashrc` (bash). `.zshenv` covers agent shells; the `.zshrc` write re-asserts priority after the brew/cargo prepends.
- **Forces-first** PATH semantics: the block strips any demoted occurrence of the shim dir (in front/middle/last position) and prepends exactly one — so it wins even when later rc lines prepend other dirs.
- **Idempotent + safe:** re-running install strips *all* prior managed blocks (and orphaned/reordered markers) and writes exactly one. Marker detection uses full-line equality, so a user rc line that merely *mentions* the marker text is never mistaken for a block boundary.
- `git-prism shim status` now warns when the shim dir isn't first on PATH for a non-login shell.

## Gauntlet

- **Adversarial QA (3 passes):** pass 1 found 2 marker idempotency bugs (duplicate blocks never collapsed; END-before-START caused unbounded growth); pass 2 found a **HIGH silent-data-loss** bug (substring marker match deleted user rc lines to EOF, including pasted setup instructions); pass 3 confirmed all fixed. 1040 tests, behave 18/18.
- **Church:** rust-purist caught a real bug (bash users were told the wrong rc files on decline) + helper extraction; test-purist replaced a fraudulent "any-position" assertion with one that *executes* the generated block under zsh; observability error context; naming/python cleanup.

Deferred (non-blocking, tracked): `SHIM_PATH_BLOCK_BODY` hardcodes the shim path literal rather than deriving from `hooks::PATH_SHIM_REL_DIR` — a drift risk that needs a const→runtime shape change.

### DIFF
--- a/bdd/features/shim_parity.feature
+++ b/bdd/features/shim_parity.feature
@@ -184,49 +184,55 @@ Feature: Shim parity with the redirect hook and first-class shim subcommand
   # decline, and already-in-PATH (no prompt needed).
   # ===========================================================================
 
-  Rule: User consents to auto-PATH setup — rc file is updated idempotently
+  Rule: User consents to auto-PATH setup — rc files are updated idempotently
 
     @ISSUE-325
-    Scenario: Accepting PATH setup appends the export line to the shell rc
+    Scenario: Accepting PATH setup writes the managed block to .zshenv for agent shells
       Given an isolated HOME directory
       And the shim directory is not in PATH
-      And a shell rc file exists at ".zprofile" under HOME
       When I run "git-prism shim install" and consent to PATH setup with the isolated HOME
       Then the exit code is 0
-      And the rc file ".zprofile" under HOME contains the shim directory export line
+      And the rc file ".zshenv" under HOME contains the shim directory export line
       And the output contains "restart"
 
     @ISSUE-325
-    Scenario: Re-running install does not append the export line a second time
+    Scenario: Accepting PATH setup writes the managed block to .zshrc to win over brew prepends
       Given an isolated HOME directory
       And the shim directory is not in PATH
-      And a shell rc file exists at ".zprofile" under HOME
+      When I run "git-prism shim install" and consent to PATH setup with the isolated HOME
+      Then the exit code is 0
+      And the rc file ".zshrc" under HOME contains the shim directory export line
+
+    @ISSUE-325
+    Scenario: Re-running install does not duplicate the managed block
+      Given an isolated HOME directory
+      And the shim directory is not in PATH
       When I run "git-prism shim install" and consent to PATH setup with the isolated HOME
       And I run "git-prism shim install" and consent to PATH setup with the isolated HOME again
-      Then the rc file ".zprofile" under HOME contains the shim export line exactly once
+      Then the rc file ".zshenv" under HOME contains the shim export line exactly once
 
-  Rule: User declines auto-PATH setup — rc file is not modified
+  Rule: User declines auto-PATH setup — rc files are not modified
 
     @ISSUE-325
-    Scenario: Declining PATH setup prints manual instructions and leaves rc unchanged
+    Scenario: Declining PATH setup prints manual instructions and leaves rc files unchanged
       Given an isolated HOME directory
       And the shim directory is not in PATH
-      And a shell rc file exists at ".zprofile" under HOME
+      And a shell rc file exists at ".zshenv" under HOME
       When I run "git-prism shim install" and decline PATH setup with the isolated HOME
       Then the exit code is 0
-      And the rc file ".zprofile" under HOME is unchanged
+      And the rc file ".zshenv" under HOME is unchanged
       And the output contains ".local/share/git-prism/bin"
 
-  Rule: Shim directory already in PATH — no prompt, no rc modification
+  Rule: Shim directory already first in PATH — no prompt, no rc modification
 
     @ISSUE-325
-    Scenario: No PATH prompt when shim directory is already in PATH
+    Scenario: No PATH prompt when shim directory is already first in PATH
       Given an isolated HOME directory
       And the shim directory is already in PATH
-      And a shell rc file exists at ".zprofile" under HOME
+      And a shell rc file exists at ".zshenv" under HOME
       When I run "git-prism shim install" with the isolated HOME
       Then the exit code is 0
-      And the rc file ".zprofile" under HOME is unchanged
+      And the rc file ".zshenv" under HOME is unchanged
 
   # ===========================================================================
   # @ISSUE-326 — redirect hook removal (hard removal, no deprecation phase)
--- a/bdd/steps/shim_parity_steps.py
+++ b/bdd/steps/shim_parity_steps.py
@@ -651,17 +651,25 @@ def step_rc_contains_export(context: Context, rel_path: str) -> None:
     )
 
 
+_SHIM_BLOCK_START_MARKER = "# >>> git-prism shim >>>"
+
+
 @then("the rc file \"{rel_path}\" under HOME contains the shim export line exactly once")
 def step_rc_contains_export_exactly_once(context: Context, rel_path: str) -> None:
-    """Assert the shim export line appears exactly once (idempotency)."""
+    """Assert the managed PATH block appears exactly once (idempotency).
+
+    The new marker-delimited block contains the shim path fragment twice (once
+    in each branch of the case statement), so we count block start-markers
+    rather than raw fragment occurrences.
+    """
     rc_path = context.isolated_home / rel_path
     assert rc_path.exists(), f"RC file not found: {rc_path}"
     content = rc_path.read_text()
-    occurrences = content.count(_SHIM_EXPORT_FRAGMENT)
+    occurrences = content.count(_SHIM_BLOCK_START_MARKER)
     assert occurrences == 1, (
-        f"Expected the shim export fragment to appear exactly once in {rc_path}, "
-        f"but found {occurrences} occurrences.\n"
-        f"Fragment: {_SHIM_EXPORT_FRAGMENT!r}\n"
+        f"Expected the managed PATH block to appear exactly once in {rc_path}, "
+        f"but found {occurrences} occurrences of the start marker.\n"
+        f"Marker: {_SHIM_BLOCK_START_MARKER!r}\n"
         f"File contents:\n{content}"
     )
 
--- a/bdd/steps/shim_parity_steps.py
+++ b/bdd/steps/shim_parity_steps.py
@@ -21,6 +21,7 @@ Platform note for @ISSUE-322:
 
 from __future__ import annotations
 
+import json
 import os
 import re
 import shutil
@@ -141,10 +142,9 @@ def step_shim_dir_already_in_path(context: Context) -> None:
 
     Stores the modified PATH in context so the When step can inject it.
     """
-    shim_dir = str(_shim_dir_path(context))
     shim_dir_path = _shim_dir_path(context)
     shim_dir_path.mkdir(parents=True, exist_ok=True)
-    context.injected_path = f"{shim_dir}:{os.environ.get('PATH', '')}"
+    context.injected_path = f"{shim_dir_path}:{os.environ.get('PATH', '')}"
 
 
 @given("the shim is installed with a \"gh\" symlink")
@@ -479,8 +479,9 @@ def step_shim_install_consent(context: Context) -> None:
     """Run git-prism shim install, answering 'y' to the PATH setup prompt."""
     binary = _find_binary(context)
     env = _isolated_env(context)
-    # Pin SHELL to zsh so detect_rc_file always picks .zprofile, regardless of
-    # the CI runner's ambient $SHELL (which is /bin/bash on GitHub Actions).
+    # Pin SHELL to zsh so the install always writes the zsh rc files
+    # (.zshenv + .zshrc), regardless of the CI runner's ambient $SHELL
+    # (which is /bin/bash on GitHub Actions).
     env["SHELL"] = "/bin/zsh"
     # If the shim dir was injected into PATH by a prior Given step, use that.
     if hasattr(context, "injected_path"):
@@ -505,8 +506,9 @@ def step_shim_install_consent_again(context: Context) -> None:
     """Run git-prism shim install a second time with 'y' consent (idempotency check)."""
     binary = _find_binary(context)
     env = _isolated_env(context)
-    # Pin SHELL to zsh so detect_rc_file always picks .zprofile, regardless of
-    # the CI runner's ambient $SHELL (which is /bin/bash on GitHub Actions).
+    # Pin SHELL to zsh so the install always writes the zsh rc files
+    # (.zshenv + .zshrc), regardless of the CI runner's ambient $SHELL
+    # (which is /bin/bash on GitHub Actions).
     env["SHELL"] = "/bin/zsh"
     result = subprocess.run(  # noqa: S603
         [str(binary), "shim", "install"],
@@ -717,11 +719,10 @@ def step_files_entries_have_change_scope(context: Context) -> None:
     binary returns plain text or its own JSON, neither of which has a
     change_scope field on file entries.
     """
-    import json as _json
     stdout = (context.result.stdout or "").strip()
     try:
-        data = _json.loads(stdout)
-    except _json.JSONDecodeError as exc:
+        data = json.loads(stdout)
+    except json.JSONDecodeError as exc:
         raise AssertionError(
             f"Output is not valid JSON: {exc}\n"
             f"stdout: {stdout[:500]}\n"
--- a/src/shim_cmd.rs
+++ b/src/shim_cmd.rs
@@ -227,17 +227,20 @@ fn write_marker_block(rc_path: &std::path::Path) -> Result<()> {
 ///   - Inverted block: END appears before START
 ///   - Orphaned END markers with no preceding START
 ///
-/// The invariant: after this call, the returned string contains neither
-/// `BLOCK_START_MARKER` nor `BLOCK_END_MARKER`.
+/// Marker detection uses full-line equality (`line.trim() == MARKER`), not
+/// substring matching.  A user line that merely *mentions* the marker text
+/// (e.g. `echo "# >>> git-prism shim >>>"`) is preserved unchanged.  Only a
+/// line whose entire trimmed content is the marker string is treated as a
+/// block boundary.
 fn strip_all_marker_blocks(content: &str) -> String {
     let mut result = String::with_capacity(content.len());
     let mut inside_block = false;
 
     for line in content.lines() {
-        if line.contains(BLOCK_START_MARKER) {
+        if line.trim() == BLOCK_START_MARKER {
             // Enter managed block; discard this line.
             inside_block = true;
-        } else if line.contains(BLOCK_END_MARKER) {
+        } else if line.trim() == BLOCK_END_MARKER {
             // Exit managed block; discard this line regardless of whether we
             // saw a matching START (handles orphaned END markers).
             inside_block = false;
@@ -510,6 +513,39 @@ mod tests {
             after.matches(BLOCK_START_MARKER).count(),
             1,
             "exactly one managed block must remain; got:\n{after}"
+        );
+    }
+
+    /// ADVERSARIAL QA (issue #355, gate-3 pass 2): a user line that merely
+    /// *contains* the start-marker text as a substring (e.g. an `echo` or a
+    /// comment that references the marker, or instructions the tool itself
+    /// printed and the user pasted as a note) must NOT cause the writer to treat
+    /// everything after it as a managed block and silently delete it.
+    ///
+    /// `strip_all_marker_blocks` keys on `line.contains(BLOCK_START_MARKER)`, so
+    /// the substring flips `inside_block = true` and every following user line is
+    /// dropped until an END marker appears — and if none does, deletion runs to
+    /// EOF. This is real data loss of user-owned rc content.
+    #[test]
+    fn write_rc_block_does_not_eat_user_lines_that_mention_the_marker_substring() {
+        let dir = TempDir::new().unwrap();
+        let home = dir.path();
+        // No real managed block here — just a user comment/echo that happens to
+        // quote the marker text, followed by genuine user config.
+        let initial = format!(
+            "export EDITOR=vim\necho \"paste the {BLOCK_START_MARKER} line into your rc\"\nexport SECRET_TOKEN=keepme\nalias gp='git-prism'\n"
+        );
+        std::fs::write(home.join(".zshenv"), &initial).unwrap();
+        write_rc_block_for_shell(home, "zsh").unwrap();
+        let after = std::fs::read_to_string(home.join(".zshenv")).unwrap();
+        assert!(
+            after.contains("export SECRET_TOKEN=keepme"),
+            "user config after a line that only mentions the marker substring must \
+             be preserved, not deleted; got:\n{after}"
+        );
+        assert!(
+            after.contains("alias gp='git-prism'"),
+            "all trailing user config must survive; got:\n{after}"
         );
     }
 
--- a/src/shim_cmd.rs
+++ b/src/shim_cmd.rs
@@ -11,9 +11,6 @@ use anyhow::Result;
 
 use crate::hooks;
 
-/// The full export line written to shell rc files.
-const SHIM_EXPORT_LINE: &str = r#"export PATH="$HOME/.local/share/git-prism/bin:$PATH""#;
-
 /// Install the PATH shim symlink and print the result to stdout.
 ///
 /// Idempotent: if the symlink already exists and points at the current binary,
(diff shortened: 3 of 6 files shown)


mikelane (commented):
## PR #358 Review — fix(shim): write PATH block to .zshenv + .zshrc so the shim wins for agent shells

**Verdict: APPROVE** — 0 CRITICAL, 0 HIGH, 1 MEDIUM, 2 LOW. No blockers.

---

### Quality Tools

| Tool | Result |
|------|--------|
| `cargo fmt --check` | PASS |
| `cargo clippy --all-targets -- -D warnings` | PASS (3 warnings all in vendored Kotlin C grammar, pre-existing) |
| `cargo test` | PASS — 1077 tests, 0 failures |
| `behave features/shim_parity.feature --tags='~@not_implemented'` | PASS — 18 scenarios, 95 steps, 0 failures |

---

No bugs confirmed via failing tests. The two pre-existing bugs that were already found+fixed (marker idempotency accumulation on truncated blocks; substring marker matching causing silent data loss) are documented by the adversarial test corpus — every probed case passes.

---

No linter exclusions, no ignore directives, no suppressed checks added in this PR.

---

### Phase 3 — Marker-Block Idempotency

**Verified correct.** `strip_all_marker_blocks` iterates every line and uses a state machine (`inside_block: bool`). Handles all four corruption forms:
- Well-formed START..END pairs (any count) — all stripped
- Truncated block (START, no END) — handled: `inside_block` stays `true`, all subsequent lines dropped, then a single canonical block is appended
- Inverted (END before START) — orphaned END sets `inside_block = false` (no-op since already false), then START enters the block; both markers stripped
- Orphaned END alone — stripped without error

The pentest in `shim_marker_pentest.rs` runs the binary end-to-end against all four corruptions and chains two installs to verify convergence. All pass.

---

### Phase 3 — Full-Line Equality Data-Loss Fix (HIGH finding, confirmed fixed)

The pre-fix code used `line.contains(BLOCK_START_MARKER)` — a substring check. A user line like `echo "# >>> git-prism shim >>>"` would trigger `inside_block = true` and silently delete all subsequent user config until an END marker appeared (or EOF — deleting everything). That is real, irreversible data loss.

The fix at `src/shim_cmd.rs:240-243` uses `line.trim() == BLOCK_START_MARKER` — full-line equality. A line that merely *mentions* the marker text as a substring is never treated as a block boundary. The test `write_rc_block_does_not_eat_user_lines_that_mention_the_marker_substring` and the BDD-level pentest `consent_writes_block_even_when_fragment_appears_in_unrelated_line` both confirm this.

No other location in the production code uses `contains(BLOCK_START_MARKER)` for boundary detection — the grep confirms all production uses of the marker constants are either: the equality comparison in `strip_all_marker_blocks`, the `format!` that assembles the new block, or `matches().count()` used only in tests for counting assertions.

(The docstring at line 525 that says "keys on `line.contains(...)`" is describing the *old* buggy behavior being tested against — accurate description of what the test is proving was fixed.)

---

### Phase 3 — Forces-First PATH Semantics

**Verified correct.** The `SHIM_PATH_BLOCK_BODY` `case` statement:
1. Checks whether `:/Users/mikelane/.opencode/bin:/Users/mikelane/google-cloud-sdk/bin:/Users/mikelane/.local/share/solana/install/active_release/bin:/usr/local/sbin:/Users/mikelane/.cargo/bin:/opt/homebrew/bin:/Users/mikelane/.rbenv/shims:/Users/mikelane/.rbenv/bin:/Users/mikelane/.pyenv/shims:/Users/mikelane/.pyenv/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/pkg/env/global/bin:/Library/Apple/usr/bin:/Library/TeX/texbin:/opt/homebrew/sbin:/Users/mikelane/.nvm/versions/node/v24.13.1/bin:/Users/mikelane/.krew/bin:/Users/mikelane/.opencode/bin:/Users/mikelane/google-cloud-sdk/bin:/Users/mikelane/.local/share/solana/install/active_release/bin:/usr/local/sbin:/Users/mikelane/.cargo/bin:/Users/mikelane/Library/pnpm:/opt/homebrew/opt/fzf/bin:/Users/mikelane/.local/bin:/Users/mikelane/bin:/Users/mikelane/Library/Application Support/JetBrains/Toolbox/scripts:/Users/mikelane/.orbstack/bin:/Users/mikelane/.lmstudio/bin:/Applications/Warp.app/Contents/Resources/bin:/Users/mikelane/bin:/Users/mikelane/Library/Application Support/JetBrains/Toolbox/scripts:/Users/mikelane/.orbstack/bin:/Users/mikelane/.lmstudio/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/agent-sdk-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/amazon-location-service/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/atomic-agents/324399402b9b/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/aws-serverless/1.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/chrome-devtools-mcp/1.1.1/bin:/Users/mikelane/.claude/plugins/cache/church/church/2.3.1/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/circleback/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/claude-code-setup/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/claude-md-management/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/code-review/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/code-simplifier/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/commit-commands/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/data-engineering/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/data/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/deploy-on-aws/1.2.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/discord/0.0.4/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/feature-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/frontend-design/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/github/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/gopls-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/greptile/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/hookify/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/imessage/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/jdtls-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/kotlin-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/mcp-server-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/mintlify/acd6d2e0128c/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/neon/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/notion/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/playground/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/playwright/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/plugin-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/postman/1.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pr-review-toolkit/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pyright-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/rust-analyzer-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/security-guidance/2.0.3/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/skill-creator/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/slack/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/superpowers/5.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/swift-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/typescript-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pydantic-ai/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-code-warp/warp/2.1.0/bin:` already *starts with* `:/Users/mikelane/.local/share/git-prism/bin:` — if so, skip (shim is already first)
2. Otherwise: wraps PATH in colons, applies global `` substitution to strip ALL existing occurrences (colon-anchored, so `git-prism/bin-extra` superstrings are not stripped), strips leading/trailing colons, then prepends the shim dir

The behavioral pentest `generated_block_puts_shim_first_exactly_once_from_any_path_position` *executes* the generated block under real zsh with the shim in front/middle/last/absent positions and asserts the result in each case. This is the correct way to test shell codegen — not structural substring assertions. All four cases pass.

---

### Phase 3 — Shell Routing (install/decline/status)

**Verified correct.** `rc_files_for_shell` uses `Path::new(shell).file_name()` — basename extraction, not substring matching. `/usr/local/bin/newbash` routes to zsh files (basename `newbash` ≠ `bash`); `/bin/bash` routes to bash files. Empty/unknown shell defaults to zsh. Tests `shell_routing_does_not_misroute_shell_path_containing_bash_as_substring` and `shell_routing_defaults_to_zsh_files_for_unknown_shell` confirm.

The BDD decline scenario pins `SHELL=/bin/zsh` in the consent When step (line 484 of `shim_parity_steps.py`) so CI on GitHub Actions (which runs bash by default) produces deterministic assertions against `.zshenv`/`.zshrc`. The decline step does NOT pin SHELL — it just runs with ambient env and checks the pre-created `.zshenv` is unchanged, which is correct since a decline never writes anything regardless of shell.

Status uses `run_status_with_path` which probes only the injected PATH string — no rc file interaction, no shell routing needed. Correct.

---

All comments explain WHY (e.g., the rationale for `case ":/Users/mikelane/.opencode/bin:/Users/mikelane/google-cloud-sdk/bin:/Users/mikelane/.local/share/solana/install/active_release/bin:/usr/local/sbin:/Users/mikelane/.cargo/bin:/opt/homebrew/bin:/Users/mikelane/.rbenv/shims:/Users/mikelane/.rbenv/bin:/Users/mikelane/.pyenv/shims:/Users/mikelane/.pyenv/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/pkg/env/global/bin:/Library/Apple/usr/bin:/Library/TeX/texbin:/opt/homebrew/sbin:/Users/mikelane/.nvm/versions/node/v24.13.1/bin:/Users/mikelane/.krew/bin:/Users/mikelane/.opencode/bin:/Users/mikelane/google-cloud-sdk/bin:/Users/mikelane/.local/share/solana/install/active_release/bin:/usr/local/sbin:/Users/mikelane/.cargo/bin:/Users/mikelane/Library/pnpm:/opt/homebrew/opt/fzf/bin:/Users/mikelane/.local/bin:/Users/mikelane/bin:/Users/mikelane/Library/Application Support/JetBrains/Toolbox/scripts:/Users/mikelane/.orbstack/bin:/Users/mikelane/.lmstudio/bin:/Applications/Warp.app/Contents/Resources/bin:/Users/mikelane/bin:/Users/mikelane/Library/Application Support/JetBrains/Toolbox/scripts:/Users/mikelane/.orbstack/bin:/Users/mikelane/.lmstudio/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/agent-sdk-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/amazon-location-service/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/atomic-agents/324399402b9b/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/aws-serverless/1.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/chrome-devtools-mcp/1.1.1/bin:/Users/mikelane/.claude/plugins/cache/church/church/2.3.1/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/circleback/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/claude-code-setup/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/claude-md-management/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/code-review/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/code-simplifier/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/commit-commands/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/data-engineering/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/data/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/deploy-on-aws/1.2.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/discord/0.0.4/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/feature-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/frontend-design/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/github/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/gopls-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/greptile/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/hookify/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/imessage/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/jdtls-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/kotlin-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/mcp-server-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/mintlify/acd6d2e0128c/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/neon/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/notion/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/playground/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/playwright/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/plugin-dev/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/postman/1.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pr-review-toolkit/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pyright-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/rust-analyzer-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/security-guidance/2.0.3/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/skill-creator/unknown/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/slack/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/superpowers/5.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/swift-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/typescript-lsp/1.0.0/bin:/Users/mikelane/.claude/plugins/cache/claude-plugins-official/pydantic-ai/0.1.0/bin:/Users/mikelane/.claude/plugins/cache/claude-code-warp/warp/2.1.0/bin:"` pattern, the colon-anchoring note about superstring siblings, the separator logic). No what-comments found. Section-header comments are absent. The adversarial test docstrings explaining the pre-fix bug behavior are particularly clear.

---

### Phase 5 — Test Quality

Tests access only public API (`write_rc_block_for_shell`, `run_install_with_io`, `run_status_with_path_for_test`). No private method access. The pentest suite uses `env!("CARGO_BIN_EXE_git-prism")` for black-box binary invocation. One-behavior-per-test. Parametrize not used where it could be (the four PATH-position cases are handled by a loop in one test) — this is a MEDIUM finding, noted below.

---

### Phase 7 — Security

The rc block body (`SHIM_PATH_BLOCK_BODY`) is a compile-time constant. The `write_marker_block` function at line 201 assembles: `format!("{BLOCK_START_MARKER}\n{SHIM_PATH_BLOCK_BODY}\n{BLOCK_END_MARKER}\n")` — all three components are `&'static str` constants. No user-controlled value is ever interpolated into the written block. No shell injection vector exists in the rc content.

The rc file path is `home.join(env_file)` where `env_file` is one of four literal strings (`.zshenv`, `.zshrc`, `.bash_profile`, `.bashrc`). No path traversal possible.

`shim_dir_is_first_in_path` reads `std::env::var("PATH")` and performs only string comparisons — no execution, no interpolation.

No symlink-following hazard: `std::fs::write` creates/overwrites the file at the resolved path. If HOME itself is a symlink, that's the user's own configuration. No TOCTOU: the read-then-write in `write_marker_block` is a normal rc-file operation, not a security boundary.

---

Branch diverged from `a8090ac` which is the exact current tip of main. `git log a8090ac..origin/main` is empty — zero commits landed on main since divergence. No regression risk. No overlap analysis needed.

---

### Non-Blocking Findings

**MEDIUM — Loop in test instead of parametrize** (`tests/shim_marker_pentest.rs:188-209`): `generated_block_puts_shim_first_exactly_once_from_any_path_position` iterates `cases` with a `for` loop. A failing case produces a useful error message (label is included), but the loop stops at first failure so subsequent cases are not reported. Using separate test functions or a macro-generated parametrize would give per-case isolation. Not a blocker — the behavioral verification here is genuinely valuable and the error messages are diagnostic.

**LOW — `run_status_with_io` and `run_status_with_path_for_test` are `#[cfg(test)]` but `pub`** (`src/shim_cmd.rs:286-299`): These are only reachable from tests in the same crate, so the `pub` visibility is wider than necessary. `pub(crate)` or `pub(super)` would be more precise. No correctness impact.

**LOW — Comment in `shim_marker_pentest.rs` docstring references old implementation detail** (line 13): "The block writer keys idempotency on the START marker (`contains(START)`), then `replace_marker_block` reassembles..." — this describes the pre-fix implementation that `strip_all_marker_blocks` replaced. It's accurate as historical context for why those tests exist, but a reader unfamiliar with the history might expect a function named `replace_marker_block` to still exist. Consider a one-line note that this describes the prior approach the tests were written to probe.

---

### Praise

The behavioral pentest approach in `generated_block_puts_shim_first_exactly_once_from_any_path_position` is exactly right: extracting the real generated block from a real binary run and executing it under zsh tests the actual contract, not a fragile substring of the source. This pattern should be the template for any future shell-codegen tests in this codebase.

The `strip_all_marker_blocks` state machine is clean and handles all four corruption forms in one pass without lookahead. The trailing-newline normalization logic (`trim_end_matches('\n')` + conditional re-append) correctly preserves the invariant that the cleaned content ends with exactly one newline when non-empty.

The `rc_files_for_shell` basename-extraction approach correctly closes the shell-path substring bug (a shell at `/usr/local/newbash` is not bash) without over-engineering — a two-arm match on basename is exactly the right shape for this problem.

---

**VERDICT: APPROVE** — 0 blockers. Ready to merge.

mikelane (commented):
## Re-review at f7c8c16 — APPROVE

The single change since the prior APPROVE (2d9f6e6) is one test-only commit: a `zsh_available()` probe added to `tests/shim_marker_pentest.rs` that early-returns from `generated_block_puts_shim_first_exactly_once_from_any_path_position` when zsh is not found on the host, rather than panicking with NotFound.

### Skip-guard correctness

The guard is sound on all three dimensions:

**When zsh IS present** (macOS, dev machines): `zsh_available()` returns `true`, the early return is not taken, and the full codegen-under-zsh assertion — execute the generated PATH block through real zsh across the front/middle/last/absent position matrix — runs exactly as before.

**When zsh is absent** (ubuntu-latest, windows-latest CI runners): the early return is taken and an `eprintln!` notes which coverage layer still holds. The marker detection, idempotency, forces-first logic, and data-loss guard are all exercised by the hermetic unit tests in `src/shim_cmd.rs` regardless of whether zsh is installed. Only the "does the generated shell code actually execute correctly under zsh" assertion is deferred; that is the right trade-off for a CI runner that has no zsh.

**The probe itself** (`Command::new("zsh").arg("-c").arg("exit 0").output().map(|o| o.status.success()).unwrap_or(false)`) correctly distinguishes "zsh not found" (returns false via `unwrap_or`) from "zsh found but exits non-zero" (also false — conservative). No false positives.

### No other tests weakened

Scanned all three test files that contained `return;` or `skip` patterns:
- `shim_marker_pentest.rs` — only the one zsh-execution pentest is guarded; all other 14+ tests in this file are unconditional.
- `shim_auto_path_pentest.rs` — unchanged; no new skips.
- `test_gh_classify.rs` — the only conditional is `#[cfg(unix)]` (compile-time) and a vacuous-if-gh-absent comment inside `it_passes_through_gh_non_diff_subcommands`, neither of which is new.

### Quality gates

| Check | Result |
|---|---|
| `cargo fmt --check` | PASS (exit 0) |
| `cargo clippy --all-targets -- -D warnings` | PASS (0 Rust warnings; vendored Kotlin C scanner warning pre-existing) |
| `cargo test` | PASS — 1077 tests, 0 failures, 11 suites |
| BDD `shim_parity.feature` | PASS — 18/18 scenarios, 0 skipped, 0 failed |

### Regression check

Branch diverges from `a8090ac` (v0.9.3 release commit). `git log a8090ac..origin/main` is empty — main has not moved since divergence. Zero regression risk.

All findings from the prior APPROVE still hold. No production code changed in this commit.

Issue #364, opened 2026-06-05
fix(cli): replace 4× current_dir().expect() panics with ? in structured subcommand handlers


mikelane:
## Summary

The four structured CLI subcommand handlers (`manifest`, `history`, `snapshot`, `context`) resolve the default repo path with an unhandled `.expect()`:

```rust
let repo_path = repo.map(PathBuf::from).unwrap_or_else(|| {
    std::env::current_dir().expect("cannot determine current directory")
});
```

`src/main.rs:336` (Manifest), `:364` (History), `:389` (Snapshot), `:416` (Context).

`std::env::current_dir()` returns `Err` when the cwd has been deleted or is inaccessible (a real case for the shim, which runs inside agent-driven throwaway tmp repos that may be removed out from under a long-ish call). On that error the binary **panics** with a backtrace instead of returning a clean `anyhow` error. `main()` already returns `anyhow::Result<()>`, so the idiomatic fix is `?`, not panic.

## Why it matters now

This is on the structured shim path (`git diff`/`log`/`show`/`blame`) — exactly the path agents hit constantly. A panic there prints a Rust backtrace into the agent's tool output instead of a tidy error, and (pre-#361) could interleave with the telemetry teardown. Found during the #361/#362 church pass (rust-purist flagged it CRITICAL); pre-existing on `main`, so filed separately rather than expanding that PR's scope.

## Proposed fix

Extract a small helper and use `?`:

```rust
fn resolve_cli_repo_path(repo: Option<String>) -> anyhow::Result<PathBuf> {
    match repo {
        Some(r) => Ok(PathBuf::from(r)),
        None => std::env::current_dir()
            .context("cannot determine current working directory"),
    }
}
```

Replace all four `unwrap_or_else(|| current_dir().expect(...))` sites with `resolve_cli_repo_path(repo)?`. The `serve` and other subcommands are unaffected.

## Acceptance criteria

- [ ] None of the `manifest`/`history`/`snapshot`/`context` handlers call `.expect()` on `current_dir()`.
- [ ] A deleted/inaccessible cwd produces a clean `anyhow` error (non-zero exit, human-readable message) instead of a panic + backtrace.
- [ ] Test: invoking one handler with a removed cwd returns an `Err`, not a panic. (TDD — the failing test comes first.)

## Out of scope

- The shim passthrough path (doesn't resolve a repo path this way).
- Any other `.expect()`/`.unwrap()` audit beyond these four sites.

Issue #363, opened 2026-06-05
serve: self-terminate orphaned MCP daemons + document shim-vs-MCP


mikelane:
Part of #360 (Symptom 3 — `serve` daemons accumulate one-per-session).

## Summary

`git-prism serve` (the MCP stdio server) accumulates roughly **one process per Claude Code session, never cleaned up** — 5+ observed, uptimes spanning from ~2 h to >1 day. These are the MCP server launched by the client's `~/.claude.json` registration; their lifecycle is owned by the MCP client, but git-prism can defend against orphaning.

## Clarifications (from #360 investigation)

- The **shim does not talk to these daemons** — it's a self-contained one-shot process (`gix::open` in-process; the only socket code in the repo is `telemetry.rs`). So daemon↔shim contention is **not** the per-call latency source.
- However, the orphaned daemons **do** load the OTLP collector (each exports telemetry), which contributes to the structured-flush stalls tracked in the sibling flush-bounding issue.
- If a user is fully on the **shim model** and their agents never call the git-prism MCP tools (`get_change_manifest`, etc.), **the MCP `serve` registration is redundant** and can be removed — which eliminates the accumulation entirely.

## Scope

- `src/server.rs` (serve lifecycle): make `serve` self-terminate when its client/parent goes away — e.g. exit on stdin EOF (stdio transport closed) and/or parent-process death — so a dead Claude Code session can't leave an indefinite orphan.
- Docs: clearly distinguish the **shim** path (transparent git/gh interception, no server) from the **MCP server** path (the tools), and state that shim-only setups don't need the `serve` registration.
- Out of scope: Claude Code's own MCP-server reaping (upstream).

## Acceptance criteria

- [ ] A `serve` process exits promptly when its stdio client disconnects / parent dies (no indefinite orphans across sessions).
- [ ] Docs explain shim-vs-MCP and that shim-only users can drop the MCP registration.
- [ ] (Optional) an idle-timeout self-exit as a backstop.

## Note

This is partly mitigated for the reporter today: if they're shim-only, removing the `git-prism serve` entry from `~/.claude.json` stops the accumulation immediately (and reduces collector load, helping the stalls).

Issue #360, opened 2026-06-05
Shim adds ~0.3-0.5s to every git call and intermittently stalls for minutes; `serve` daemons accumulate one-per-session


mikelane:
## Summary

On an AI-agent-driven session, the `git` shim adds steady per-call latency and, intermittently, blocks for *minutes* on a single call. Meanwhile `git-prism serve` daemons pile up (looks like one per session, never cleaned up), and I suspect contention between them is the stall source. Surfaced while running a Python test suite that shells `git` heavily inside throwaway tmp repos.

## Environment

- git-prism 0.9.3
- macOS 26.4.1, arm64
- git 2.50.1 (Apple Git-155)
- shim: `~/.local/share/git-prism/bin/git` -> `/opt/homebrew/bin/git-prism`, ahead of `/usr/bin/git` on PATH
- session driven by an AI coding agent (Claude Code), so `agent-detect` is presumably active

## Symptom 1: per-call overhead (steady)

`git rev-parse HEAD`, shim vs real binary:

    shim  (~/.local/share/git-prism/bin/git):  0.33-0.60s
    real  (/usr/bin/git):                       0.07-0.31s

So ~0.3-0.5s on every call, including read-only ones. Across a suite that makes hundreds of git calls in tmp repos that measured as 187s (shim) vs 133s (shim bypassed), about 54s of pure tax.

## Symptom 2: intermittent multi-minute stalls

Same suite, same machine, ranged from ~40s to ~3min run-to-run. During a slow run I sampled the process tree: the parent (pytest) sat at ~1.5% CPU for 2m37s (blocked, not computing) with a live `git-prism` child. The call eventually returned and the suite passed, so it is a stall, not a hang. Non-deterministic.

## Symptom 3: `serve` daemons accumulate

5 `git-prism serve` processes running, by uptime:

    1d 03h
    23h 33m
    23h 33m
    02h 12m
    01h 45m

Looks like roughly one per session with no cleanup. If they share a backing store or lock, contention is a plausible cause of Symptom 2.

## Impact

Agent-driven workflows that shell git in ephemeral repos (test suites, tight CI-like loops) pay a large and unpredictable tax. Recording throwaway tmp repos is probably not desirable for the data model either.

## What would help (questions / asks)

1. Is the multi-daemon accumulation expected? Should `serve` be a singleton, or GC stale daemons? Could several contend and cause the stalls?
[...]

PR #365, opened 2026-06-05
fix(shim): bound structured-path telemetry teardown (#361) + per-invocation opt-out (#362)

### LINKED ISSUE #361: fix(shim): bound the structured-path telemetry flush to stop multi-minute stalls
Part of #360 (Symptom 2 — multi-minute stalls).

## Summary

The shim's **structured-output** path (`git diff` / `log` / `show` / `blame`, which build and print a JSON manifest) flushes telemetry with an **unbounded** `force_flush()` at `src/main.rs:301`, while the **passthrough** path correctly uses `force_flush_bounded(PASSTHROUGH_FLUSH_TIMEOUT)` (300 ms). So a structured call blocks for as long as the OTLP export takes. When the collector at `GIT_PRISM_OTLP_ENDPOINT` is saturated or slow, a single `git diff` can stall for **minutes**.

Introduced by #348 (telemetry-in-shim), shipped in v0.9.3. Before that the shim never flushed at all.

## Why it stalls

`telemetry::init_quiet()` + a metric record + flush now run on every shim call. On the structured path the final flush (`main.rs:301`) is uncapped. With several orphaned `git-prism serve` daemons (see the sibling daemon-lifecycle issue) plus heavy shim traffic all exporting to the same `localhost:4318` collector, the collector backs up and the uncapped flush blocks the git call. Reported live: a `git` call sat blocked (~1.5% CPU) for 2m37s with a live `git-prism` child, then returned.

## Scope

- `src/main.rs` shim dispatch: bound the structured-path flush the same way the passthrough path is bounded (`force_flush_bounded`). Consider exposing the bound via an env var (e.g. `GIT_PRISM_FLUSH_TIMEOUT_MS`), since the deferred-findings issue #359 already flagged that the 300 ms passthrough constant is compile-time only.
- Out of scope: the passthrough path (already bounded); the broader per-invocation opt-out (separate issue); the daemon accumulation (separate issue).

## Acceptance criteria

- [ ] The structured-output path's telemetry flush is time-bounded; a saturated or unreachable collector cannot block a `git diff`/`log`/`show`/`blame` call for more than the bound.
- [ ] Metrics still flush when the collector is responsive (no data loss in the common case).
- [ ] A test proves the structured path returns within the bound even when the exporter endpoint is a black hole.
- [ ] Repro check: with `GIT_PRISM_OTLP_ENDPOINT` pointed at an unreachable/slow endpoint, a structured shim call returns promptly instead of blocking.

## Verify-the-cause first

Confirm the diagnosis by reproducing #360's stall with `GIT_PRISM_OTLP_ENDPOINT` **unset** — if the multi-minute stalls disappear, the unbounded structured flush is the cause.

### LINKED ISSUE #362: feat(shim): per-invocation passthrough/opt-out so ephemeral repos skip processing + telemetry
Part of #360 (Symptom 1 — steady per-call overhead; ask #2).

## Summary

There is no first-class way to tell the shim to **skip its processing and telemetry** for a given call or repo. Agent-driven workflows that shell `git` heavily in throwaway tmp repos (test suites, tight loops) pay ~0.3–0.5 s/call of git-prism cold-start + (since v0.9.3) telemetry init/flush — measured as ~54 s of tax across a suite (187 s shimmed vs 133 s bypassed) — and recording ephemeral repos isn't desirable for the data model anyway. The only workaround today is putting `/usr/bin` ahead of the shim on `PATH`, which is all-or-nothing.

## What "skip" should mean

When the opt-out is active the shim should **`exec` the real `git`/`gh` immediately**, before `telemetry::init_quiet()` and any classification — i.e. ~zero added latency, identical behavior to the real binary, and **no** recording.

## Design options (pick one or more)

1. **Env var (primary):** `GIT_PRISM_PASSTHROUGH=1` (or `GIT_PRISM_DISABLE=1`) → unconditional immediate exec. Simplest; covers the "wrap a test-suite command" case. Note the reporter already tried `GIT_PRISM_DISABLE` / `_OFF` / `_BYPASS` / `_PASSTHROUGH` expecting this — none short-circuit today.
2. **Marker file:** honor a `.git-prism-ignore` (or similar) anywhere up the cwd's parents → skip for that repo tree. Good for "this checkout is ephemeral."
3. **cwd ignore-glob:** a configurable set of path globs (e.g. tmp dirs) the shim skips.

## Scope

- `src/main.rs` shim dispatch / `src/shim/`: short-circuit to `exec` real git/gh **before** telemetry init when the opt-out is active.
- Decide the opt-out surface (env var is the minimal high-value piece; marker-file/glob are stretch).
- Docs: document the opt-out and recommend it over the `PATH` hack.

## Acceptance criteria

- [ ] An env var makes the shim exec the real binary with no measurable added latency vs `/usr/bin/git` (no telemetry init, no classification).
- [ ] No telemetry/recording happens for opted-out calls.
- [ ] Documented in README / shim docs.
- [ ] Test: opted-out invocation produces byte-identical output to real git and emits no metrics.
- [ ] (Stretch) marker-file and/or cwd-glob opt-out.

mikelane:
## Summary

Two fixes from the #360 perf investigation, both in the shim/telemetry path:

- **Closes #361** — bound the structured-output path's telemetry teardown. The shim's structured path (`git diff`/`log`/`show`/`blame` → JSON manifest) flushed telemetry with an **unbounded** `force_flush()`, so a saturated or unreachable OTLP collector could stall a `git diff` for minutes (#360 symptom 2). The flush is now `force_flush_bounded(STRUCTURED_FLUSH_TIMEOUT)` (500 ms), the unbounded `TelemetryGuard::force_flush()` method is gone, **and** the `TelemetryGuard::drop` shutdown — which `main.rs` runs immediately after the flush on `return` — is now bounded too (`DROP_SHUTDOWN_TIMEOUT` 500 ms). The second half mattered: bounding only the explicit flush merely relocated the stall into the guard's `Drop`, where `tp.shutdown()` + `mp.shutdown()` each block up to 5 s on a dead collector. Caught by adversarial QA driving the **built binary** end-to-end (a `#[cfg(test)]` unit test can't reproduce it — the real provider wiring is `#[cfg(not(test))]`-gated).

- **Closes #362** — per-invocation opt-out. Set `GIT_PRISM_PASSTHROUGH=1` (or the alias `GIT_PRISM_DISABLE=1`) and the shim execs the real `git`/`gh` **immediately** — before `telemetry::init_quiet()`, before classification — with near-zero added latency and no recording. This replaces the all-or-nothing `PATH` hack for agent workflows that shell `git` heavily in throwaway tmp repos.

## What changed

- `src/main.rs` — opt-out checked before telemetry init on the shim branch; structured-path flush now bounded.
- `src/shim/mod.rs` — `passthrough_opt_out_requested()` (truthy = `1`/`true`, case-insensitive); `STRUCTURED_FLUSH_TIMEOUT` (500 ms); opt-out short-circuit at step 0 of `run_shim`.
- `src/telemetry.rs` — removed unbounded `force_flush()`; bounded `Drop` shutdown via the same spawn + `recv_timeout` pattern as `force_flush_bounded`.
- `README.md` — "Per-invocation opt-out" section + env-var table rows.
- `tests/shim_passthrough_optout_pentest.rs` — opt-out e2e: recursion-safe, no telemetry, `DISABLE` alias.
- `tests/shim_structured_teardown_pentest.rs` — drives the built binary through the structured path against an unroutable TEST-NET-1 OTLP endpoint and asserts the whole invocation returns under 4 s (was ~10.5 s before the Drop bound).

## Testing

- Full suite green (`cargo test`), including both new e2e pentests.
- `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean.
- Opt-out verified for `1`/`true`/`TRUE`/`GIT_PRISM_DISABLE`; inactive for `0`/`false`/empty/unset/`2`/`yes`/`on`/whitespace.
- Structured `git diff` against a black-hole collector: ~10.5 s → 3.16 s.

## Out of scope

- #363 (serve daemon lifecycle) — separate PR.
- Pre-existing `current_dir().expect()` panics in the CLI subcommand handlers — filed as #364.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Per-invocation shim opt-out via `GIT_PRISM_PASSTHROUGH` / `GIT_PRISM_DISABLE`.** Setting either env var to a truthy value (`1` or case-insensitive `true`) makes the shim `exec` the real `git` / `gh` immediately — before `telemetry::init_quiet()` and before command classification — with near-zero added latency and no recording. This replaces the all-or-nothing `PATH` hack for agent workflows that shell `git` heavily in throwaway tmp repos, where per-call cold-start and telemetry overhead is unwanted and recording ephemeral repos is undesirable anyway. The opt-out exec is recursion-safe (it resolves and runs the real binary, not the shim symlink, even when the shim is first on `PATH`). (#362)
+
+### Fixed
+
+- **The shim's structured-output path (`git diff` / `log` / `show` / `blame`) no longer stalls on an unreachable OpenTelemetry collector.** The structured path flushed telemetry with an unbounded `force_flush()`, and `main` then dropped the `TelemetryGuard`, whose `Drop` ran unbounded `shutdown()` calls — so a single `git diff` could block for minutes (≈10 s per call was reproduced against a black-hole endpoint: 5 s tracer shutdown + 5 s meter shutdown). Both the explicit flush and the `Drop`-path shutdown are now time-bounded (500 ms each), mirroring the already-bounded passthrough path; a saturated or unreachable collector can no longer stall an intercepted git command. Telemetry still delivers normally when the collector is responsive (proven by an end-to-end data-loss guard against a live capture server). (#361)
+
 ## [0.9.3] — 2026-06-03
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -166,6 +166,27 @@ yourself to force vanilla git for one command:
 GIT_PRISM_INSIDE_SHIM=1 git diff main..HEAD   # bypasses the shim
 ```
 
+### Per-invocation opt-out (`GIT_PRISM_PASSTHROUGH`)
+
+Set `GIT_PRISM_PASSTHROUGH=1` (or the alias `GIT_PRISM_DISABLE=1`) to make the
+shim exec the real git/gh **immediately** — before telemetry init, before
+classification, with zero added latency:
+
+```bash
+GIT_PRISM_PASSTHROUGH=1 git diff main..HEAD   # real git, no shim overhead
+```
+
+This is the first-class alternative to putting `/usr/bin` ahead of the shim on
+`PATH`. Use it to wrap a test suite or tight loop where shim processing and
+telemetry are not desirable:
+
+```bash
+GIT_PRISM_PASSTHROUGH=1 cargo test   # git calls inside tests skip the shim
+```
+
+Accepted truthy values: `1`, `true` (case-insensitive). Any other value (or
+unset) leaves the shim active.
+
 ### Debugging
 
 Set `GIT_PRISM_DEBUG_RESOLVER=1` to print the resolved real-git path to stderr,
@@ -739,6 +760,8 @@ Optional OpenTelemetry instrumentation, disabled by default and opt-in via envir
 | `GIT_PRISM_OTLP_HEADERS` | Planned, not yet wired ([#43](https://github.com/mikelane/git-prism/issues/43)). Setting this variable has no effect today; managed OTLP backends that require auth headers need a local collector proxy. | unset |
 | `GIT_PRISM_SERVICE_NAME` | Service name reported to the backend. | `git-prism` |
 | `GIT_PRISM_SERVICE_VERSION` | Service version reported to the backend. | crate version |
+| `GIT_PRISM_PASSTHROUGH` | Set to `1` or `true` to skip shim processing and telemetry entirely for a single invocation. See [Per-invocation opt-out](#per-invocation-opt-out-git_prism_passthrough). | unset |
+| `GIT_PRISM_DISABLE` | Alias for `GIT_PRISM_PASSTHROUGH`. | unset |
 
 Quick start with Jaeger (any OTLP-compatible backend works):
 ```bash
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ GIT_PRISM_INSIDE_SHIM=1 git diff main..HEAD   # bypasses the shim
 
 Set `GIT_PRISM_PASSTHROUGH=1` (or the alias `GIT_PRISM_DISABLE=1`) to make the
 shim exec the real git/gh **immediately** — before telemetry init, before
-classification, with zero added latency:
+classification, with near-zero added latency:
 
 ```bash
 GIT_PRISM_PASSTHROUGH=1 git diff main..HEAD   # real git, no shim overhead
--- a/src/main.rs
+++ b/src/main.rs
@@ -298,7 +298,7 @@ async fn main() -> std::process::ExitCode {
             &exec,
             &mut telemetry_guard,
         );
-        telemetry_guard.force_flush();
+        telemetry_guard.force_flush_bounded(shim::STRUCTURED_FLUSH_TIMEOUT);
         return exit_code;
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,8 +5,6 @@ pub(crate) mod metrics;
 pub(crate) mod pagination;
 pub(crate) mod privacy;
 mod server;
-// The shim module is complete but not yet wired to argv[0] dispatch (#287).
-#[allow(dead_code)]
 mod shim;
 mod shim_cmd;
 mod telemetry;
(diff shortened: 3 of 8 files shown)


mikelane (commented):
## Code Review — PR #365: fix(shim): bound structured-path telemetry teardown (#361) + per-invocation opt-out (#362)

**VERDICT: APPROVE** — 0 critical, 0 high, 2 medium, 1 low. No blockers.

---

### Summary

Both fixes are sound and the implementation is correct end-to-end. The bounded-Drop pattern is the right design: spawn one thread, move ownership of both providers into it, abandon after a single `recv_timeout`. No shared mutable state escapes to the main thread after timeout, double-shutdown is impossible (move semantics + `take()` empties the `Option`s), and a healthy collector still receives telemetry (proven by the data-loss pentest). The opt-out predicate is minimal, correct, and recursion-safe.

---

No bugs found. The adversarial QA artifact records the run-2 HIGH finding (Drop path still unbounded after force_flush was bounded) and confirms it was fixed in commit 27bc943. The three pentests provide concrete regression guards.

**Soundness analysis of the bounded-Drop pattern (`telemetry.rs:104-128`):**
- **No UB:** providers are *moved* into the spawned thread; the struct's `Option` fields are `None` the instant `take()` returns. No raw pointers, no `unsafe`, no shared mutable state across the thread boundary.
- **No data loss on healthy collector:** `force_flush_bounded` runs *before* Drop (the `main.rs` sequence is explicit: flush → return → Drop). For a healthy collector the flush completes within its 500 ms window and the export is already delivered by the time Drop's shutdown thread starts. The data-loss pentest guards this.
- **No double-shutdown:** a `TelemetryGuard` is dropped exactly once (Rust move semantics). The `if tp.is_none() && mp.is_none() { return; }` early-exit makes even a hypothetical re-entry a no-op.

---

### Phase 1 — Quality tools

- **`cargo fmt --check`**: clean (exit 0).
- **`cargo clippy --all-targets -- -D warnings`**: exit 0. The 3 warnings are pre-existing vendored tree-sitter-kotlin C build warnings — identical on `main`, not caused by this PR.

---

### Phase 5 — Test quality

**`shim_structured_teardown_pentest.rs`** — Non-vacuous: asserts `!out.stdout.is_empty()` (proves structured path, not passthrough). 4 s timing bound is calibrated correctly: generous for slow CI, tight enough to catch a regression to ~10.5 s.

**`shim_passthrough_optout_pentest.rs`** — Three separate behaviors in three separate test functions. The marker-git pattern for recursion safety is elegant; the additional `!stdout.starts_with('{')` assertion kills the mutation where structured JSON leaks through. The no-telemetry test correctly asserts absence of telemetry using a live capture server.

**`shim_bounded_drop_dataloss_pentest.rs`** — Guards the dual case: healthy-collector delivery under bounded teardown. `recv_timeout(5s)` for arrival is appropriately generous vs the 500 ms Drop budget.

**Unit test `qa_structured_path_total_teardown_is_bounded_on_black_hole_endpoint`** — Correctly documents its own limitation (providers are no-ops in test builds; the unit-level test cannot reproduce the real stall). This self-limitation comment is exemplary — future readers won't waste time wondering why the e2e pentest "duplicates" what looks like unit coverage.

---

### Phase 8 — Regression check (mandatory)

- Divergence point: `cbb693e893c61add1e8ee47c68d27d022b7af40b`
- Files modified on `main` since divergence: **none** — branch was cut from current `main` HEAD, no sibling PRs landed after divergence.
- Deletion lines in PR patch: **0**
- Regression risk: none.

---

### Findings

| MEDIUM | `tests/shim_passthrough_optout_pentest.rs:150` | `recv_timeout` maps `Err(Disconnected)` to `false` via `unwrap_or` — if the capture-server thread panicked, the test passes as a false negative. Low probability but worth a comment explaining the known limitation. |
| MEDIUM | `tests/shim_passthrough_optout_pentest.rs` + `tests/shim_bounded_drop_dataloss_pentest.rs` | `spawn_capture_server` is byte-for-byte duplicated. Consider extracting to `tests/common/capture_server.rs` in a follow-up. Tolerable in test code but unnecessary duplication. |
| LOW | `src/telemetry.rs:75` | Comment says "Clone the Arc-backed provider handles so they can be moved into the thread" (WHAT). Missing the WHY: clone-not-take is intentional so the guard stays `is_active()` and Drop's shutdown thread still runs. A one-liner would prevent future readers from "fixing" this to `take()`. |

---

### Praise

The adversarial QA loop that caught "fix merely relocated the stall" (run 2) and then the dual data-loss check (run 3) is exactly the rigor these fixes require. The CHANGELOG entry frames the problem precisely: "merely relocated from `force_flush` to `Drop`" is the correct diagnosis. The `it_tolerates_double_force_flush_then_drop_on_active_guard` test is an excellent guard for the clone-not-take design decision — it documents the intent as an executable contract.

The two MEDIUM findings are post-merge improvements. Neither affects correctness.

Issue #366, opened 2026-06-05
serve orphan watchdog misses PID-namespace reparenting (getppid()==0)


mikelane:
Follow-up from the #363 church/QA review.

## Summary

The serve orphan watchdog's predicate is `should_exit_orphaned(ppid) = (ppid == 1)` (`src/server.rs`). It catches the common case — when the launching Claude Code session dies, the daemon is reparented to init/launchd (pid 1) and exits. This solves the reported macOS accumulation (#360 symptom 3) correctly.

**Gap:** in a Linux **PID namespace**, a process whose parent lived outside the namespace can observe `getppid() == 0` (the parent is no longer mapped into the namespace) rather than `1`. In that case `should_exit_orphaned(0)` returns `false` and the watchdog **never fires** — the exact orphan scenario it exists to prevent, missed. Reachable in containerized Claude Code / CI setups; not reachable on macOS (launchd reparents to 1).

Current behavior is pinned by a test (`it_returns_false_when_ppid_is_zero`) added in #363 so the gap is documented, not silent.

## Decision needed

- **Option A — broaden the predicate to `ppid <= 1`.** Treat `getppid() == 0` (parent invisible) as orphaned too. Simple, closes the gap. Risk: could a *healthy* serve process ever legitimately see `getppid() == 0`? (Only if it is itself pid 1 of a namespace whose creator lives outside — unusual for `git-prism serve`.) Needs confirmation that 0 never occurs for a non-orphan in target environments.
- **Option B — document the limitation.** Keep `ppid == 1`; state in the watchdog doc that PID-namespace reparenting to 0 is out of scope (rely on stdin-EOF / idle-timeout backstops instead).

## Acceptance criteria

- [ ] A decision is recorded (A or B) with rationale about whether `getppid() == 0` is reachable for a healthy `git-prism serve` in supported environments.
- [ ] If A: predicate becomes `ppid <= 1`, the pinning test flips to assert `true` for `0`, and a watchdog comment explains the namespace case.
- [ ] If B: the watchdog doc comment states the PID-namespace limitation explicitly.

## Context

The watchdog uses a 5 s poll of `libc::getppid()` and `std::process::exit(0)` on orphan. This issue is only about the *predicate's* namespace edge — not the polling mechanism.

PR #370, opened 2026-06-05
feat(serve): self-terminate orphaned MCP daemons on parent death (#363)

### LINKED ISSUE #363: serve: self-terminate orphaned MCP daemons + document shim-vs-MCP

mikelane:
## Summary

**Closes #363** (part of #360 — `serve` daemons accumulate one-per-session).

The MCP `serve` daemon had no self-termination path: when the Claude Code session that launched it exited without closing stdin, the process lingered indefinitely. In shim-heavy setups this accumulated multiple orphaned `git-prism serve` processes (observed: 5+, uptimes from ~2 hours to >1 day), each still loading the OTLP collector.

`serve` now spawns a watchdog that polls its parent pid every 5 s and exits when it has been reparented to init/launchd (`ppid == 1`) — the signal that the launching session died. The README also documents the shim-vs-MCP distinction and notes that shim-only users can drop the `serve` registration entirely (`claude mcp remove git-prism`).

## What changed

- `src/server.rs` — `should_exit_orphaned(ppid)` (pure, injected pid for hermetic testing), `spawn_orphan_watchdog()` (tokio task: poll `libc::getppid()` every `ORPHAN_POLL_INTERVAL`, `std::process::exit(0)` on orphan with a pid-tagged stderr line), wired into `run_server`. Unit tests for the predicate including the `ppid == 0` boundary.
- `Cargo.toml` — `libc = "0.2"` (already in the build graph transitively; now declared) for `getppid()`.
- `README.md` — "Shim vs. MCP server — which do you need?" section + escape hatch.
- `CHANGELOG.md` — `[Unreleased]` entry.

## Design notes

- The hard `process::exit(0)` from the watchdog intentionally skips the telemetry-flush-on-drop: a reparented process's launching session is gone and its collector is likely gone too; lingering to flush is worse than dropping a few spans.
- The predicate keys on `ppid == 1` (init/launchd reparenting), which covers the reported macOS accumulation. The Linux PID-namespace edge where `getppid() == 0` is **out of scope** here — pinned by a test (`it_returns_false_when_ppid_is_zero`) and tracked for a decision in #366.

## Testing

- Full suite green (`cargo test`); `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean.
- Adversarial QA (gate 3) passed at the final HEAD; the watchdog predicate has true/false-polarity triangulation (ppid 1 → exit; 0/2/12345 → stay).

## Out of scope (filed separately)

- `getppid() == 0` PID-namespace handling — #366.
- The MCP client's own server reaping (upstream Claude Code).

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -13,6 +13,43 @@ use crate::tools::{
     build_manifest, build_review_change, build_snapshots, build_worktree_manifest,
 };
 
+/// How often the orphan watchdog polls the parent pid.
+const ORPHAN_POLL_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);
+
+/// Returns `true` when the process has been reparented to init/launchd (ppid == 1),
+/// which means the Claude Code session that launched `git-prism serve` has exited
+/// without closing stdin. Exiting on this condition prevents indefinite orphan daemons.
+///
+/// The argument is injected for testability — callers pass `libc::getppid() as u32`
+/// in production and a fixed value in unit tests.
+pub fn should_exit_orphaned(ppid: u32) -> bool {
+    ppid == 1
+}
+
+/// Spawn a background task that polls the parent pid every [`ORPHAN_POLL_INTERVAL`]
+/// and hard-exits the process when [`should_exit_orphaned`] returns `true`.
+///
+/// This is a fire-and-forget guard: `tokio::select!` in `run_server` races the
+/// serve future against a never-resolving placeholder, while this task hard-exits
+/// on orphan. The hard-exit is intentional — a reparented process has no meaningful
+/// cleanup to do (telemetry flush is best-effort; the OTLP collector is gone too).
+fn spawn_orphan_watchdog() {
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(ORPHAN_POLL_INTERVAL).await;
+            // SAFETY: getppid() is always safe to call; it never fails.
+            let ppid = unsafe { libc::getppid() } as u32;
+            if should_exit_orphaned(ppid) {
+                eprintln!(
+                    "git-prism: parent process exited (ppid == 1); \
+                     exiting orphaned serve process"
+                );
+                std::process::exit(0);
+            }
+        }
+    });
+}
+
 /// Convert a `ChangeScope` variant to a static metric label string.
 fn change_scope_label(scope: ChangeScope) -> &'static str {
     match scope {
@@ -882,6 +919,11 @@ pub async fn run_server() -> anyhow::Result<()> {
         );
     }
     crate::metrics::get().record_session_started();
+    // Spawn the orphan watchdog before starting the serve loop. It runs
+    // concurrently and hard-exits the process if the parent pid becomes 1
+    // (reparented to init/launchd), which happens when the Claude Code session
+    // that launched `git-prism serve` dies without closing stdin.
+    spawn_orphan_watchdog();
     let server = GitPrismServer::new();
     let transport = tokio::io::join(tokio::io::stdin(), tokio::io::stdout());
     server.serve(transport).await?.waiting().await?;
@@ -1079,6 +1121,32 @@ mod tests {
         assert!(
             server_info.capabilities.prompts.is_none(),
             "prompts capability must not be advertised — this server does not implement prompts"
+        );
+    }
+
+    #[test]
+    fn it_exits_when_reparented_to_init() {
+        assert!(
+            should_exit_orphaned(1),
+            "ppid == 1 means the process was reparented to init/launchd \
+             (the Claude Code session died); serve must exit"
+        );
+    }
+
+    #[test]
+    fn it_does_not_exit_when_parent_is_alive() {
+        assert!(
+            !should_exit_orphaned(2),
+            "ppid == 2 means a live parent; serve must not exit"
+        );
+    }
+
+    #[test]
+    fn it_does_not_exit_for_typical_shell_parent_pid() {
+        // Realistic shell pids are in the thousands; must not trigger exit.
+        assert!(
+            !should_exit_orphaned(12345),
+            "ppid == 12345 is a normal shell pid; serve must not exit"
         );
     }
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -14,6 +14,7 @@ use crate::tools::{
 };
 
 /// How often the orphan watchdog polls the parent pid.
+#[cfg(unix)]
 const ORPHAN_POLL_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);
 
 /// Returns `true` when the process has been reparented to init/launchd (ppid == 1),
@@ -22,7 +23,8 @@ const ORPHAN_POLL_INTERVAL: std::time::Duration = std::time::Duration::from_secs
 ///
 /// The argument is injected for testability — callers pass `libc::getppid() as u32`
 /// in production and a fixed value in unit tests.
-pub fn should_exit_orphaned(ppid: u32) -> bool {
+#[cfg(unix)]
+pub(crate) fn should_exit_orphaned(ppid: u32) -> bool {
     ppid == 1
 }
 
@@ -35,6 +37,7 @@ pub fn should_exit_orphaned(ppid: u32) -> bool {
 /// serve future entirely. The hard-exit is intentional — a reparented process has
 /// no meaningful cleanup to do (telemetry flush is best-effort; the OTLP collector
 /// is likely gone too).
+#[cfg(unix)]
 fn spawn_orphan_watchdog() {
     tokio::spawn(async move {
         loop {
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -13,6 +13,52 @@ use crate::tools::{
...
+/// is likely gone too).
+fn spawn_orphan_watchdog() {
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(ORPHAN_POLL_INTERVAL).await;
+            // SAFETY: `getppid` is a POSIX system call with no preconditions —
+            // it takes no arguments, dereferences no pointers, and touches no
+            // shared state, so there is nothing the caller can violate. It is
+            // also infallible (it cannot return an error).
+            let raw_ppid: libc::pid_t = unsafe { libc::getppid() };
```
mikelane, line 49: **CRITICAL — Windows build broken (CI: Test (windows-latest) FAILED)**

`libc::pid_t` and `libc::getppid()` are POSIX-only; they are not defined in the `libc` crate on Windows targets. The compiler emits:

```
error[E0425]: cannot find type `pid_t` in crate `libc`
error[E0425]: cannot find function `getppid` in crate `libc`
```

The fix is to gate the entire watchdog on `#[cfg(unix)]`:

```rust
#[cfg(unix)]
const ORPHAN_POLL_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);

#[cfg(unix)]
fn spawn_orphan_watchdog() {
    tokio::spawn(async move {
        loop {
            tokio::time::sleep(ORPHAN_POLL_INTERVAL).await;
            let raw_ppid: libc::pid_t = unsafe { libc::getppid() };
            let ppid = raw_ppid as u32;
            if should_exit_orphaned(ppid) {
                eprintln!(
                    "git-prism: serve pid {} reparented to init (ppid == 1);                      parent Claude Code session exited — shutting down orphaned daemon",
                    std::process::id()
                );
                std::process::exit(0);
            }
        }
    });
}

#[cfg(not(unix))]
fn spawn_orphan_watchdog() {
    // Orphan detection via getppid() is POSIX-only; no-op on Windows.
}
```

The call-site in `run_server()` needs no change — it compiles on all platforms once both cfg variants exist. The `ORPHAN_POLL_INTERVAL` constant must also be gated (or moved inside the unix impl) to avoid an `unused item` warning on Windows.

Reference: the `libc` crate only exposes `pid_t` and `getppid` on unix targets.

mikelane (commented):
## PR Review — feat(serve): self-terminate orphaned MCP daemons on parent death (#363)

### Summary

This is a well-designed, focused feature: an orphan-watchdog background task that polls `getppid()` and hard-exits when reparented to init/launchd. The design rationale is clearly documented (why hard-exit vs. graceful shutdown, why 5s poll, why `ppid == 1`). The injectable predicate (`should_exit_orphaned(ppid: u32)`) cleanly separates testable logic from the syscall. The README additions are clear and accurate.

**One CRITICAL blocker prevents merge — Windows build is broken.**

---

### Blockers

#### CRITICAL — `src/server.rs:46` — Windows build fails to compile

`libc::pid_t` and `libc::getppid()` are POSIX-only and do not exist in the `libc` crate on Windows targets. CI confirms: `Test (windows-latest)` fails with:

```
error[E0425]: cannot find type `pid_t` in crate `libc`
error[E0425]: cannot find function `getppid` in crate `libc`
```

The watchdog function and the `ORPHAN_POLL_INTERVAL` constant must be wrapped in `#[cfg(unix)]` / `#[cfg(not(unix))]` guards. Idiomatic fix:

```rust
#[cfg(unix)]
const ORPHAN_POLL_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);

#[cfg(unix)]
fn spawn_orphan_watchdog() {
    tokio::spawn(async move {
        loop {
            tokio::time::sleep(ORPHAN_POLL_INTERVAL).await;
            let raw_ppid: libc::pid_t = unsafe { libc::getppid() };
            let ppid = raw_ppid as u32;
            if should_exit_orphaned(ppid) {
                eprintln!(
                    "git-prism: serve pid {} reparented to init (ppid == 1);                      parent Claude Code session exited — shutting down orphaned daemon",
                    std::process::id()
                );
                std::process::exit(0);
            }
        }
    });
}

#[cfg(not(unix))]
fn spawn_orphan_watchdog() {
    // Orphan detection via getppid() is POSIX-only; no-op on Windows.
}
```

The call-site in `run_server()` needs no change.

---

### Other Findings

**MEDIUM** — `src/server.rs:25` — `should_exit_orphaned` is `pub` but all callers are inside this crate. The injectable-argument design makes it fully testable at `pub(crate)` or even private visibility. Recommend `pub(crate)` to tighten the API surface.

**LOW** — `README.md` — Two occurrences of ASCII `--` in the shim-vs-MCP table body where the surrounding prose uses em-dashes. Cosmetic inconsistency.

**LOW** — `src/server.rs:39` — `tokio::spawn` returns a `JoinHandle` that is silently dropped (fire-and-forget, as documented). This is intentional and correct; flagging only so a future reader with a cancellation requirement knows where to look.

---

### Praise

- The `// SAFETY:` comment on the `unsafe` block is exemplary — it addresses all three conditions (no preconditions, no pointer dereferences, no shared state, infallible).
- The `ppid == 0` test pins current behavior with a reference to #366. This is the correct pattern for "known gap, tracked separately".
- Triangulation across four test cases (0, 1, 2, 12345) is thorough and all test names are behaviorally descriptive.
- Pure-addition diff (132 lines added, 0 deleted from existing code) — zero regression risk confirmed against main.
- Hard-exit justification is documented at the declaration site, not just in the PR description.

---

### Verdict: REQUEST_CHANGES (1 critical blocker)

Add `#[cfg(unix)]` / `#[cfg(not(unix))]` guards to restore the Windows build. All other aspects are merge-ready once this is resolved.

src/server.rs
```diff
@@ -13,6 +13,61 @@ use crate::tools::{
...
+            if should_exit_orphaned(ppid) {
+                eprintln!(
+                    "git-prism: serve pid {} reparented to init (ppid == 1); \
+                     parent Claude Code session exited — shutting down orphaned daemon",
+                    std::process::id()
+                );
+                std::process::exit(0);
+            }
+        }
+    });
```
mikelane, line 62: **Fix verified at 6fd54de.**

`#[cfg(unix)]` correctly gates `ORPHAN_POLL_INTERVAL`, `should_exit_orphaned`, the real `spawn_orphan_watchdog`, and all four predicate tests. The `#[cfg(not(unix))]` no-op stub allows `run_server()` to compile unchanged on all targets. Unix behavior is identical to the reviewed version. `pub(crate)` applied to `should_exit_orphaned` (MEDIUM finding addressed). CI fully green including `Test (windows-latest)`. **Blocker resolved — APPROVED.**

mikelane (commented):
## Follow-up Review — fix verified at 6fd54de

**APPROVED.**

The critical blocker (Windows build broken by `libc::getppid()`/`pid_t`) is fully resolved. The cfg-gating is correct and complete:

- `#[cfg(unix)]` on `ORPHAN_POLL_INTERVAL`, `should_exit_orphaned`, and the real `spawn_orphan_watchdog` — no Unix behavior change.
- `#[cfg(not(unix))]` no-op `fn spawn_orphan_watchdog() {}` — `run_server()`'s call-site compiles unchanged on all targets.
- All four predicate tests gated `#[cfg(unix)]` — they reference the now-unix-only `should_exit_orphaned`.
- `should_exit_orphaned` tightened to `pub(crate)` (MEDIUM finding addressed).

The LOW finding on `libc` being declared unconditionally in `Cargo.toml` rather than under `[target.'cfg(unix)'.dependencies]` is harmless — Rust does not warn on unused crate dependencies, and Windows CI is green. Tracked in #371; not a blocker.

CI fully green at this SHA: Format / Check / Clippy / Integrity / Test (ubuntu) / Test (macos) / Test (windows) / BDD Acceptance / Mutants — all pass. `mergeStateStatus: CLEAN`.

PR #372, opened 2026-06-06
feat(serve): watchdog testability seam + lifecycle logs; document orphan predicate (#366, #371)

### LINKED ISSUE #366: serve orphan watchdog misses PID-namespace reparenting (getppid()==0)

### LINKED ISSUE #371: chore(serve): watchdog testability seam + serve lifecycle log lines (#363 deferrals)
Low-priority polish deferred from the #363 church/QA review (PR #370). Non-blocking; batched.

## 1. Watchdog loop testability seam (test-purist)

`spawn_orphan_watchdog` (`src/server.rs`) is only partially tested — `should_exit_orphaned` is unit-tested, but the loop's *wiring* (does it actually call `should_exit_orphaned(getppid())` rather than a hardcoded value?) and the poll-interval constant aren't. A mutation like `from_secs(5) → from_secs(0)` survives. Extract a thin injectable seam so the getppid→predicate wiring is testable without spawning a process:

```rust
fn watchdog_should_exit(getppid: impl Fn() -> u32) -> bool {
    should_exit_orphaned(getppid())
}
```
`process::exit(0)` and the literal sleep duration remain legitimately hard to unit-test and are fine to leave to a BDD/integration check.

## 2. Lifecycle log lines (observability-purist)

The watchdog logs on orphan-exit (with pid), but the surrounding `serve` lifecycle is otherwise quiet:
- **No startup "now serving" line.** `run_server` has no stderr marker that the serve loop began, so there's no "started at T" to pair against the watchdog's "exiting" line. Suggested: `eprintln!("git-prism: serve pid {} started; orphan-watchdog polling every {}s", std::process::id(), ORPHAN_POLL_INTERVAL.as_secs())` after the watchdog spawn.
- **Clean shutdown is silent.** When `waiting()` returns normally (stdin closed), `run_server` falls through to `Ok(())` with no line, so a vanished daemon can't be distinguished from a crash. Suggested: a single line before the telemetry drop.

## Acceptance criteria

- [ ] The getppid→predicate wiring has a unit test via an injectable seam (kills the hardcoded-ppid and zero-interval mutants).
- [ ] `serve` emits a startup line and a clean-shutdown line so the full daemon lifetime is reconstructable from stderr.

These are debuggability/maintainability improvements, not correctness fixes.

mikelane:
## Summary

Hardens the `serve` orphan watchdog. Two issues:

- **Closes #366** — resolved by **keeping `should_exit_orphaned` at `ppid == 1`** and documenting why `getppid() == 0` is deliberately *not* treated as orphaned. The investigation: orphaned processes always reparent to PID 1 (init/launchd, or the PID namespace's own init), so an orphan observes `getppid() == 1` — already caught. `getppid() == 0` instead means the parent is in an ancestor PID namespace, i.e. the process is itself a namespace/container **init** (e.g. a container entrypoint). Broadening to `<= 1` would false-kill such a process ~5s after startup while catching zero additional real orphans, so it was rejected. The decision is now pinned by `it_returns_false_when_ppid_is_zero` and explained in the predicate's doc comment.
- **Closes #371** — three deferred-cleanup items:
  - **Watchdog testability seam**: extracted `watchdog_should_exit(getppid: impl Fn() -> u32)` so the `getppid → predicate` wiring is unit-tested (injected fakes: `||1`→true, `||0`→false, `||4242`→false), killing the "loop ignores getppid / hardcodes the pid" mutant. The `loop`/`sleep`/`exit(0)` shell stays the accepted untestable surface.
  - **Lifecycle logging**: `serve pid N started` (cross-platform), `orphan-watchdog active; polling every 5s` (Unix), and `serve pid N stdin closed; shutting down cleanly` (cross-platform, before the telemetry drop) — so an operator can reconstruct the full daemon lifetime: started → [watchdog active] → {orphaned-exit | clean stdin-close}.
  - **Dependency hygiene**: moved `libc = "0.2"` to `[target.'cfg(unix)'.dependencies]` (it's only used for the Unix-gated `getppid()`).

All watchdog code remains `#[cfg(unix)]`-gated with the `#[cfg(not(unix))]` no-op; Windows compiles unchanged. No behavior change to the predicate vs. what shipped in #363 — this PR adds the testability seam, the lifecycle logs, the dependency-hygiene fix, and the documented `getppid()==0` decision.

## Testing

- Full suite green; `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean.
- Predicate pinned by `{0→false, 1→true, 2→false, 3→false}`; seam wiring pinned by injected getppid fakes.
- Adversarial QA (gate 3) + church (rust/dep/test/observability/secret) — clean. (QA's container-PID-1 probe is what drove the `== 1` decision.)


src/server.rs
```diff
@@ -39,21 +61,26 @@ pub(crate) fn should_exit_orphaned(ppid: u32) -> bool {
...
-            if should_exit_orphaned(ppid) {
+            if watchdog_should_exit(|| {
+                // SAFETY: `getppid` is a POSIX system call with no preconditions —
+                // it takes no arguments, dereferences no pointers, and touches no
+                // shared state, so there is nothing the caller can violate. It is
+                // also infallible (it cannot return an error).
+                let raw: libc::pid_t = unsafe { libc::getppid() };
+                // `pid_t` is signed but a real ppid is always >= 1; the cast is a
+                // plain widening reinterpretation used only for the `<= 1` compare.
+                raw as u32
```
mikelane, line 79: **LOW — Stale comment text: `<= 1` should be `== 1`**

```rust
// `pid_t` is signed but a real ppid is always >= 1; the cast is a
// plain widening reinterpretation used only for the `<= 1` compare.
```

The inline comment says `<= 1` but the predicate at line 41 is `ppid == 1`. This is a residue from the reverted `ppid <= 1` attempt at commit 6829277. The comparison that actually runs is `== 1`, so the comment misleads future readers into thinking there is a range check here.

Suggested fix:
```rust
// `pid_t` is signed; real ppids are positive. Cast to u32 is a plain
// widening reinterpretation — the `== 1` compare is the only use.
```

Comments must describe what the code actually does, not a discarded alternative.

mikelane (commented):
## PR #372 Review — watchdog testability seam + lifecycle logs; document orphan predicate

### Summary

This PR delivers two tightly scoped changes: (a) a closure-injection seam on the watchdog so the getppid→predicate wiring is unit-testable without touching a real process table, plus three new unit tests pinning that seam; and (b) three stderr lifecycle log lines on `run_server` giving operators a complete daemon timeline from stderr alone. The doc comment on `should_exit_orphaned` is substantially improved, correctly documenting the `ppid == 0` / namespace-init rationale and confirming the final predicate is `== 1` (not the reverted `<= 1`).

No bugs found. The predicate is correctly `ppid == 1`. The seam composes correctly (`watchdog_should_exit(getppid) → should_exit_orphaned(getppid())`). The lifecycle log placement is correct: "started" fires after `spawn_orphan_watchdog()` but before `serve()`, and "shutting down cleanly" fires after `waiting()` returns, meaning it only emits on clean stdin-EOF exit — not on watchdog `process::exit(0)`.

### Phase 1 — Quality tools

- **Clippy**: 0 errors, 0 new warnings (1 pre-existing vendored Kotlin C scanner warning, unrelated). Confirmed via adversarial QA artifact and local oracle run.
- **Format**: CI check passed (SUCCESS on commit 691964e).
- **Tests**: All 1063 unit tests pass. 7 watchdog-specific tests confirmed green.
- **CI**: Format ✓, Clippy ✓, Integrity ✓, Mutation Testing ✓. Platform tests (ubuntu/macos/windows) and BDD Acceptance still in progress at review time — no failures.

No new linter exclusions, ignore rules, or pre-commit bypasses introduced. `Cargo.toml` change is purely structural (moving `libc` under `[target.'cfg(unix)'.dependencies]`).

### Phase 3 — Code standards

`libc` is correctly scoped to `[target.'cfg(unix)'.dependencies]`. All `libc::*` usage is inside `#[cfg(unix)]` blocks. The `unsafe` block has a load-bearing `// SAFETY:` comment explaining the POSIX preconditions. Rust idioms are correct throughout.

**One LOW finding** (see inline comment): `src/server.rs:79` — the cast comment says `<= 1` compare but the actual predicate is `== 1`. This is a stale residue from the reverted `ppid <= 1` attempt at commit 6829277. All other comments explain WHY, not what. The doc comment on `should_exit_orphaned` is exemplary — it explains the namespace/container-init rationale, why `ppid == 0` is deliberately not treated as orphaned, and the false-positive risk of broadening to `<= 1`. The `// SAFETY:` comment is accurate and complete.

### Phase 5 — Test quality

The three seam tests (`|| 1` → true, `|| 0` → false, `|| 4242` → false) correctly pin the closure wiring through public-facing behavior. Tests use `watchdog_should_exit` (module-private but visible in `#[cfg(test)]`), not internal state. The `it_returns_false_when_ppid_is_three` test correctly triangulates the boundary (`== 1`, not `< 3`). Test names use "it returns" / "it does not" framing, not "should". No loops, no complexity.

The adversarial QA artifact notes the `spawn_orphan_watchdog` loop body (the `process::exit(0)` call itself) remains untested — this is inherent to the polling loop design, not a regression, and is documented as an INFO-level observation.

### Phase 6 — Architecture

The seam design (`watchdog_should_exit(impl Fn() -> u32)`) is the correct Rust idiom for injecting a syscall dependency without a trait. It avoids over-engineering (no `GetPpid` trait, no mock struct) while remaining testable. `spawn_orphan_watchdog` remains under 20 lines. `run_server` is unchanged except for three `eprintln!` calls. No YAGNI concerns.

### Phase 7 — Security and performance

No security concerns. `unsafe { libc::getppid() }` is correctly guarded with a SAFETY comment. `process::exit(0)` is intentional (hard-exit skipping cleanup) and documented. No hardcoded secrets, no SQL, no injection vectors.

### Phase 8 — Regression check

- Divergence point: `82de28ea` (the orphan-watchdog PR #363/#370 that landed on main first)
- Files modified on main since divergence: **none** — the branch is fresh off current main. Zero regression risk.
- Deletion volume: 19 lines deleted, 93 added. Net positive; deletions are the old direct-libc wiring replaced by the seam. No unexpected deletions.

### Findings

| LOW | `src/server.rs:79` | Inline comment says `<= 1 compare` but predicate is `== 1` — stale from reverted ppid<=1 attempt |

### Praise

The doc comment on `should_exit_orphaned` (lines 21–35) is genuinely excellent: it documents a deliberate non-feature (`ppid == 0` is NOT orphaned), explains the Linux PID namespace mechanics, gives the false-positive consequence of the alternative, and references the issue number. Future maintainers will understand not just what the code does but why the adjacent design was rejected. This is the right way to document a predicate with a non-obvious boundary.

The two-function layering (`should_exit_orphaned` → pure predicate on a u32; `watchdog_should_exit` → seam that injects getppid) cleanly separates testability from production wiring without any over-abstraction.

### Decision

**COMMENT** (solo-dev repo — self-review; cannot formally approve own PR)

The single LOW finding (stale comment text `<= 1` vs actual `== 1`) does not block merge — it is a misleading comment, not a behavioral defect. All predicate logic, cfg-gating, test coverage, and documentation are correct. The stale comment should be fixed before or alongside merge.

PR #383, opened 2026-06-20
fix(git): walk A..B by reachability, not first-parent-to-base-OID

### LINKED ISSUE #382: fix(git): walk A..B by reachability, not first-parent-to-base-OID
**Title:** `fix(git): walk A..B by reachability, not first-parent-to-base-OID`

## Problem

A two-dot range (`git log A..B`) run through the git-prism shim fails whenever the two refs have diverged — i.e. when `A` is not an ancestor of `B`. Instead of listing the commits reachable from `B` but not `A`, the shim errors:

```
$ git merge-base --is-ancestor origin/main HEAD; echo $?
1
$ git log --oneline origin/main..HEAD
git-prism shim: handler error: git error: failed to read object: commit 931ab2c90e28bcd4223e6ffb709b0fa7e51eed87 has no parent but base 9c3c771d2164147dee02293dd9ebe0a494d38d73 not yet reached
```

`931ab2c` is the repository's root commit. Both directions fail (`origin/main..HEAD` and `HEAD..origin/main`); in each, the walk runs off to the same root commit before reaching the named base. Native git resolves the exact same range with no error. These all succeed against the same objects, so the objects are intact and the bug is specific to the `..` range walker: `git merge-base HEAD origin/main`, `git log -n 20 HEAD`, `git log -n 20 origin/main`, `git rebase origin/main`.

**Why now / impact:** Diverged history is the *normal* state during everyday development — branch off `main`, let `main` advance, and `origin/main..HEAD` is exactly how you list "my commits." Any agent or human using `git log origin/main..HEAD` (or `main..HEAD`) through the shim to enumerate branch commits hits this. It silently fails review and changelog workflows the moment a branch falls behind its base, which is most of the time.

## Affected Surface (the exact call chain)

This is **not** a passthrough case — the shim intercepts the command. Confirmed by reading the source:

`git log --oneline origin/main..HEAD`
→ `src/shim/classify.rs::classify_log` (line 72): `--oneline` is not in `has_scripted_output_flag`, the range token matches `has_ref_range`, so it returns `Classification::History { range: "origin/main..HEAD" }`
→ `src/shim/handlers.rs::handle_history` (line 81)
→ `src/tools/mod.rs::collect_all_history_pages` (line 125) → `build_history`
→ `src/git/reader.rs::walk_commits` (line 132) — **the broken walk.**

## Root Cause

`src/git/reader.rs::walk_commits` (line 132) walks history incorrectly:

```rust
let base_id = base_commit.id();
let mut current = head_commit;
loop {
    // ...record current...
    if current.id() == base_id {     // stops ONLY at the literal base OID
        break;
    }
    commits.push(info);
    let parent = current.parent_ids().next()   // follows FIRST parent only
        .ok_or_else(|| GitError::ReadObject(format!(
            "commit {} has no parent but base {} not yet reached",
            current.id(), base_id)))?
        // ...
    current = parent;
}
```

Two defects:

1. **Stops on OID equality, not reachability.** It breaks only when it walks *onto* the exact base OID. For `A..B` with diverged history, `A` is not on `B`'s ancestry path, so the walk never matches it and runs past the merge-base to the root commit, which has no parent → the error fires. Native git computes `A..B` as the set difference "reachable from B, not reachable from A" using merge-base/reachability.
2. **Follows first parent only.** `parent_ids().next()` ignores second+ parents of merge commits, so even on non-diverged history it can miss commits merged in from side branches.

## Recommended Approach (settles the implementation decision)

Replace the hand-rolled first-parent loop with gix's revision walk computing reachable-from-`head` excluding everything reachable from `base` (gix's `Repository::rev_walk` over the head tip with `base` as a hidden/boundary tip), rather than reimplementing set-difference by hand. The output must equal `git rev-list A..B` ordering. Do not optimize beyond correctness.

### Rust unit tests (primary — `walk_commits` is directly unit-testable; existing tests at reader.rs ~464-531 must stay green)

- Diverged history (base is NOT an ancestor of head): `walk_commits(base, head)` returns exactly the SHAs of `git rev-list base..head` and does **not** error.
- Linear history (base IS an ancestor of head): unchanged behavior — still equals `git rev-list base..head` (no regression).
- Range spanning a merge commit whose second parent is not reachable from base: every commit in `git rev-list base..head` is present (closes the first-parent-only defect).
- Disjoint histories (two unrelated root commits, no common ancestor): returns the set reachable from head, without erroring on the root commit.
- `base == head`: returns empty, no error (matches `git rev-list A..A`).

### Behave regression scenarios (end-to-end through the shim; tag each `@ISSUE-NN @not_implemented` at bootstrap, NN = this issue)

```gherkin
Scenario: log A..B on diverged history lists B-not-A commits
  Given a repo where origin/main is NOT an ancestor of HEAD (history has diverged)
  When I run "git log --oneline origin/main..HEAD" through the shim
  Then the exit code is 0
  And the listed commit SHAs equal "git rev-list origin/main..HEAD" from stock git
  And no "has no parent but base ... not yet reached" error appears

Scenario: reverse range on diverged history also works
  Given the same diverged repo
  When I run "git log --oneline HEAD..origin/main" through the shim
  Then the exit code is 0
  And the listed commit SHAs equal "git rev-list HEAD..origin/main" from stock git
```

**Gaming-test note:** Pinning every scenario to `git rev-list <range>` from stock git closes the cheat of "stop erroring but return the wrong/partial set." The merge-commit unit test closes the first-parent-only cheat. The disjoint-history unit test closes the "walk first-parent-to-root and stop at root" cheat that would pass the simple diverged case while returning B's entire ancestry instead of the set difference. The linear-history case guards the path that currently works.

## Out of Scope

- Three-dot symmetric-difference ranges (`A...B`). Two-dot `A..B` only. File `A...B` separately if also broken.
- `git log` revision-set flags beyond plain `A..B` (`--not`, multiple positive/negative refs, `^ref`, pathspec filtering).
- Performance/perf tuning of the walk — correctness only.
- Output *formatting* parity for `git log` (`--oneline` columns, colors). The shim's JSON response shape is unchanged; this ticket fixes only which commits are included.

## Blast Radius

`walk_commits` feeds `build_history` → `collect_all_history_pages` → `get_commit_history` and the `history` interception path. Changing the walk semantics affects every consumer: the existing linear-history unit tests (reader.rs ~464-531, which assert ordering) must continue to pass, and any base/head resolution shared with `diff_commits` should be spot-checked for regressions.

## Dependencies

None. Independent of the `git show <rev>:<path>` blob-spec bug filed alongside this.

## Edge Cases

- `A` and `B` are the same commit → empty result, exit 0.
- `A` is the root commit → all of `B`'s ancestry up to but excluding root, like native git.
- `B` disjoint from `A` (no common ancestor) → returns reachable-from-B without erroring on the root.
- Octopus merge (3+ parents) on the head side → covered by the reachability approach (all parents traversed), same mechanism as the two-parent merge test.
- Annotated tags peeling to commits → resolves the same as the underlying commits.

## Environment (from report)

git-prism 0.10.0 (Homebrew, `/opt/homebrew/bin/git-prism`), delegated git 2.54.0, macOS (Apple Silicon), non-shallow repo, run from inside a git worktree. Root commit observed in the failure: `931ab2c90e28bcd4223e6ffb709b0fa7e51eed87`.

mikelane:
## Summary

Fixes `git log A..B` (and the shim's `History` interception of it) failing on **diverged history** with `commit <root> has no parent but base <base> not yet reached`.

`src/git/reader.rs::walk_commits` computed the two-dot range with a hand-rolled loop that (1) stopped only on exact base-OID equality and (2) followed first-parent only. When `base` is not an ancestor of `head` — the normal state of a branch whose base has advanced — the base OID is never on head's first-parent chain, so the walk ran to the root commit and errored. Native git resolves the same range fine.

## Fix

Replace the hand-rolled walk with gix's revision walk:

```rust
self.repo
    .rev_walk([head_id])
    .with_hidden([base_id])                                   // exclude base + its ancestors → base..head
    .sorting(Sorting::ByCommitTime(CommitTimeOrder::NewestFirst))
    .all()?
// ...collect, then reverse() → oldest-first
```

`with_hidden` gives correct `git rev-list base..head` set semantics (handles diverged, merge/all-parents, and disjoint histories). `ByCommitTime(NewestFirst)` + `reverse()` matches `git rev-list --reverse`'s default (commit-date) ordering.

## Notable: ordering bug caught in-gauntlet

The first implementation relied on gix's **default BreadthFirst** order, which — even reversed — does **not** match `git rev-list --reverse` for any merge-containing range (correct set, wrong order). Adversarial QA proved this with differential tests against the `git rev-list --reverse` oracle; the explicit `ByCommitTime` sorting is the fix.

## Tests

`src/git/reader.rs` tests, all differential against real `git rev-list --reverse base..head`:
- diverged history, linear (regression), merge with unreachable 2nd parent (all-parents traversal), octopus merge, disjoint/orphan histories, base==root, head-ancestor-of-base (empty), short SHAs, annotated tags, skewed commit dates.
- Full suite: 1096 passing. `fmt`/`clippy --all-targets -D warnings` clean.

`gix-traverse` promoted from transitive to direct dep (same `0.57.0`, Cargo.lock +1 line) to name `CommitTimeOrder` (gix imports it privately, no re-export).

## Follow-ups (LOW, non-blocking)
- Land 3 of the QA edge-case graph shapes (equal-timestamp tie-break, non-monotone dates, deep skewed merges) as permanent in-repo tests — verified via the release binary during QA but not yet committed (an edit-guard blocked the QA agent from adding them).
- One doc-comment over-claims a date-vs-topo distinction; reword.

Closes #382

### DIFF
--- a/src/git/reader.rs
+++ b/src/git/reader.rs
@@ -149,16 +149,23 @@ impl RepoReader {
         // `with_hidden` marks base as a "hidden tip": the walker traverses from
         // head but excludes any commit reachable from base (including base itself),
         // which is exactly the two-dot range semantics of `git rev-list base..head`.
+        //
+        // We sort by `NewestFirst` (commit-time descending) and then reverse to match
+        // `git rev-list --reverse`, which is the order callers of `get_commit_history`
+        // expect. `BreadthFirst` (the default) reversed does NOT match git's output for
+        // ranges that span merge commits — the commit SET is correct but the ORDER
+        // diverges.  `ByCommitTime(NewestFirst)` then reversed matches `git rev-list
+        // --reverse` for all tested graph shapes.
         let walk = self
             .repo
             .rev_walk([head_id])
             .with_hidden([base_id])
+            .sorting(gix::revision::walk::Sorting::ByCommitTime(
+                gix_traverse::commit::simple::CommitTimeOrder::NewestFirst,
+            ))
             .all()
             .map_err(|e| GitError::ReadObject(e.to_string()))?;
 
-        // gix default walk order is newest-first (BreadthFirst). Reverse to
-        // oldest-first to match `git rev-list --reverse` and the existing test
-        // contract.
         let mut commits = walk
             .map(|result| {
                 let info = result.map_err(|e| GitError::ReadObject(e.to_string()))?;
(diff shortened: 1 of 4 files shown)


src/git/reader.rs
```diff
@@ -138,45 +140,49 @@ impl RepoReader {
...
+        // `with_hidden` marks base as a "hidden tip": the walker traverses from
+        // head but excludes any commit reachable from base (including base itself),
+        // which is exactly the two-dot range semantics of `git rev-list base..head`.
+        //
+        // We sort by `NewestFirst` (commit-time descending) and then reverse to match
+        // `git rev-list --reverse`, which is the order callers of `get_commit_history`
+        // expect. `BreadthFirst` (the default) reversed does NOT match git's output for
+        // ranges that span merge commits — the commit SET is correct but the ORDER
+        // diverges.  `ByCommitTime(NewestFirst)` then reversed matches `git rev-list
+        // --reverse` for all tested graph shapes.
```
mikelane, line 160: **LOW** — The block comment at lines 146–160 explaining `with_hidden` semantics and `ByCommitTime` ordering rationale is excellent and should stay. One sentence slightly over-claims: it says `BreadthFirst` reversed 'does NOT match git's output for ranges that span merge commits', which is true, but stops short of noting that `ByCommitTime` itself can also diverge from `git rev-list` when commit dates are non-monotone across parents. The `qa_walk_skewed_dates_ordering_matches_git_rev_list` test actually exercises this case — the test passing is what validates it works, not the comment. Consider tightening the last two sentences to 'matches `git rev-list --reverse` in all tested graph shapes (linear, diverged, octopus, disjoint, skewed dates)' rather than implying date-order is generally equivalent to rev-list order. Non-blocking — this is the doc-comment issue called out in the PR description as a follow-up.

mikelane (commented):
## PR Review: fix(git): walk A..B by reachability, not first-parent-to-base-OID

### Summary

This PR replaces a hand-rolled first-parent commit walk with gix's native `rev_walk(...).with_hidden(...)` API, fixing a real crash on diverged history (the normal state of any branch whose base has advanced). The fix is semantically correct, well-understood, and proven by differential tests against the `git rev-list --reverse` oracle across all meaningful graph shapes. The gauntlet caught and fixed an ordering bug (default `BreadthFirst` ≠ rev-list order on merge-containing ranges) before this review, and the final `ByCommitTime(NewestFirst) + reverse()` approach is sound.

---

The regression test `it_walks_diverged_history_without_error` directly reproduces the crash from issue #382 with a concrete diverged-history fixture, proves the old code errored, and proves the new code succeeds. No additional failing tests needed — the bug is confirmed and fixed.

---

### Phase 1 — Quality Tools

- `cargo fmt --check`: **PASS** (exit 0)
- `cargo clippy --all-targets -- -D warnings`: **PASS** — the 3 warnings are from the vendored Kotlin tree-sitter C scanner (`unused variable 'in_string'` in `vendor/tree-sitter-kotlin/src/scanner.c:503`), pre-existing in the codebase, not introduced by this PR
- `cargo test --bin git-prism`: **1096 tests PASS** (0 failures, 16.81s)

---

No changes to `.pre-commit-config.yaml`, `lefthook.yml`, or any lint/clippy config. No suppression directives added.

---

### Phase 3 — Code Standards

**Production code (`src/git/reader.rs`, lines 134–188):** Clean. Uses `?` operator consistently, error mapping via `.map_err(|e| GitError::ReadObject(e.to_string()))` follows the existing pattern in the file. No bare `unwrap()` in production paths. No shell-outs to `git` CLI in production code (confirmed by grep — zero matches for `Command::new("git")` in `src/`).

**Dependency addition:** `gix-traverse = "0.57"` promoted from transitive to direct dependency to access `CommitTimeOrder` (gix imports it privately, no re-export). The Cargo.lock shows the same `0.57.0` checksum before and after — no version skew, no new transitive deps. Minimal and justified.

**Imports:** `use gix::revision::walk::Sorting` and `use gix_traverse::commit::simple::CommitTimeOrder` — both specific, no wildcard imports.

---

The 14-line block comment (lines 146–160) explaining `with_hidden` semantics and the `ByCommitTime` ordering rationale is genuinely excellent — it explains WHY, references the specific failure mode of the default BreadthFirst order, and documents the constraint that callers expect oldest-first. This is exactly what a code comment should do. One LOW finding on the final two sentences (see inline comment) — non-blocking.

---

### Phase 5 — Test Quality

**Exceptional.** The test suite adds:
- **4 behavioral `it_` tests**: diverged history without error, linear history matching oracle, all-parents traversal through merge commits, disjoint histories, short SHAs, annotated tags on both ends
- **8 `qa_walk_` adversarial probe tests**: diverged+merge, octopus merge (3 parents), disjoint orphan, base==root, reverse range (empty), short SHAs, annotated tags, skewed commit dates (the highest-risk ordering case)

All differential against the `git rev-list --reverse` oracle. The `commit_with_date` helper pins dates via `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE` to make ordering tests deterministic across runs. Tests use only public API. No private method access. No `assert!(true)`. No loops or branching in tests. The `walk_shas` and `git_rev_list_reverse` extraction helpers at the bottom follow DRY without obscuring test intent.

**Particular praise:** `qa_walk_skewed_dates_ordering_matches_git_rev_list` — the diamond where committer dates are intentionally inverted relative to topological order — is exactly the highest-risk case for `ByCommitTime` matching `git rev-list` and it's covered. This is the test that would catch a future regression if gix changes its traversal behavior.

---

### Phase 6 — Architecture

The change is a surgical replacement of one private method body (`walk_commits`). No new abstractions, no YAGNI, no SRP violations. The public API surface is unchanged. The commit history is clean: fix → ordering fix → hoist imports → changelog.

---

No security surface changes. No network calls. No user input enters the gix walk API — `base_ref` and `head_ref` are already validated by `peel_to_commit` before the walk runs (both resolve to concrete OIDs). Memory: `collect::<Vec<_>>()` followed by `reverse()` is O(n) in the number of commits in the range — identical to what `git rev-list` itself allocates, and appropriate for the use case (commit history, not file contents).

---

Branch divergence point: `53766d1` (current main tip). This branch is **fresh off main** — zero commits on main since divergence. The diff shows 0 deletions on the left side (no code removed from main was regressed). The only changed files are `src/git/reader.rs`, `CHANGELOG.md`, `Cargo.toml`, and `Cargo.lock`. No shared files with sibling PRs.

---

### Findings

| LOW | `reader.rs:160` | Doc comment last sentence over-claims date-order equivalence — see inline comment |

---

### Blockers

None.

---

### Praise

- The `with_hidden` approach is the correct gix primitive for two-dot range semantics — this is not a workaround, it is the intended API.
- Catching the `BreadthFirst` ordering bug during adversarial QA (before this review) and fixing it with `ByCommitTime(NewestFirst)` demonstrates exactly the right testing discipline: oracle-driven differential tests that would have caught a wrong implementation.
- The `qa_walk_skewed_dates_ordering_matches_git_rev_list` test is genuinely impressive — it targets the one edge case where date-order sorting could diverge from rev-list order in a way that's hard to reason about without testing.
- The Cargo.lock change is +1 line (promoting a transitive dep to direct) with identical checksum — the minimal possible footprint for this dependency addition.

---

**Decision: APPROVE**

Clean fix, clean tests, clean tools. The correctness claim (gix `rev_walk + with_hidden` matches `git rev-list` semantics) is validated by 12 differential oracle tests covering every graph shape that matters. No blockers.

PR #384, opened 2026-06-20
fix(shim): pass git show <rev>:<path> blob specs through to real git

### LINKED ISSUE #381: fix(shim): pass git show <rev>:<path> blob specs through to real git
**Title:** `fix(shim): pass git show <rev>:<path> blob specs through to real git`

## Problem

`git show <rev>:<path>` — print a file's contents at a revision without checking it out — fails when run through the git-prism PATH shim. The shim classifies the command as a commit "show snapshot" and tries to peel the resolved object to a commit. Because `<rev>:<path>` resolves to a **blob**, not a commit, the peel fails:

```
$ which git
/Users/mike.lane/.local/share/git-prism/bin/git

$ git ls-tree origin/main --name-only | grep -x '.lefthook.yml'   # path exists at the rev
.lefthook.yml

$ git show "origin/main:.lefthook.yml"
git-prism shim: handler error: failed to resolve commit origin/main:.lefthook.yml: failed to read object: Last encountered object f09b7bbeba2 was blob while trying to peel to commit

$ /usr/bin/git show "origin/main:.lefthook.yml" | head -1   # stock git, same spec
# Lefthook Configuration
```

The same failure occurs with a SHA-based spec (`<40-char-sha>:<path>`). Read-only commands that don't read blob contents (`ls-tree`, `fetch`, `rev-parse`, `grep`) pass through fine — the breakage is scoped to `show` with `rev:path` blob specs.

**Why now / impact:** This breaks any agent or tooling workflow that reads file contents at a revision without a checkout — code review flows hit it constantly. The reporter had to fall back to the GitHub contents API mid-review on a monorepo PR. The shim's contract is to be transparent for read-only git operations; silently breaking a core porcelain command undermines that contract and makes the shim unsafe to leave first on `PATH`.

## Root Cause

Confirmed by reading the source:

- `src/shim/classify.rs::classify_show` (line 103) returns the first non-flag argument as the sha: for `git show origin/main:.lefthook.yml`, `sha = "origin/main:.lefthook.yml"`, classified as `Classification::ShowSnapshot { sha }`. There is no check for the `rev:path` (object spec) form.
- `src/shim/handlers.rs::build_show_commit_detail` (line 200) calls `RepoReader::peel_to_commit(sha)` (line 204), which errors because the object is a blob, not a commit. The wrapper `"failed to resolve commit {sha}: {e}"` (line 205) matches the reported error string exactly.

A `rev:path` object spec is not a commit-ish and has no JSON-manifest meaning. The shim should not intercept it — it should pass it through to real git so the blob prints exactly as stock git does.

## Detection Rule (closes the classification decision)

`classify_show` must route object specs to `Classification::Passthrough`. Define an object spec as a non-flag argument that contains a `:` where:

- the text **before** the first `:` is non-empty (a `<rev>`), and
- the text **after** the first `:` does **not** begin with `/`.

This routes `<rev>:<path>` (blob) and `<rev>:` (tree) to passthrough, while deliberately **excluding** `:/<regex>` (git's "search commit message" syntax — a commit-ish, must stay intercepted). The bare index spec `:<path>` (empty `<rev>`) is also passed through (safe: it has no manifest meaning). Plain `git show <commit-ish>` (no `:`) is unaffected and stays intercepted.

### Rust unit tests (primary — `classify_show` is directly unit-testable)

- `classify(["git","show","origin/main:.lefthook.yml"])` returns `Classification::Passthrough`.
- `classify(["git","show","<full-sha>:apps/x/package.json"])` returns `Classification::Passthrough`.
- `classify(["git","show","HEAD:"])` (tree spec) returns `Classification::Passthrough`.
- `classify(["git","show",":/fix typo"])` (commit-message search) does **not** return Passthrough — it stays `ShowSnapshot` (commit-ish).
- `classify(["git","show","<commit-sha>"])` (no colon) still returns `ShowSnapshot { sha }` — no regression.

### Behave regression scenarios (end-to-end, tag each `@ISSUE-NN @not_implemented` at bootstrap, NN = this issue)

```gherkin
Scenario: git show with a branch-relative blob spec prints the file
  Given the git-prism shim is first on PATH
  And the path ".lefthook.yml" exists at origin/main
  When I run "git show origin/main:.lefthook.yml"
  Then the exit code is 0
  And stdout is byte-identical to "/usr/bin/git show origin/main:.lefthook.yml"
  And no "git-prism shim: handler error" line appears on stderr

Scenario: git show with a SHA-relative blob spec prints the file
  Given the git-prism shim is first on PATH
  When I run "git show <full-sha>:<existing/path>"
  Then the exit code is 0
  And stdout is byte-identical to "/usr/bin/git show <full-sha>:<existing/path>"

Scenario: git show of a plain commit still returns the JSON manifest
  Given the git-prism shim is first on PATH
  And a commit with SHA <commit-sha> exists
  When I run "git show <commit-sha>"
  Then the response is the git-prism JSON show-manifest
  And the manifest "commit.sha" field equals the full 40-char <commit-sha>
```

**Gaming-test note:** "byte-identical to `/usr/bin/git show`" closes the cheat of returning a hand-rolled approximation of the blob. The plain-commit scenario now pins `commit.sha` to a concrete value, closing the empty-but-valid-manifest cheat. The `:/regex` and no-colon unit tests close the "just passthrough anything with a colon" cheat that would break commit-message search.

## Out of Scope

- `git show <rev>:` of a **tree** (directory). Passthrough is required (it must not error), but matching git's tree-listing output byte-for-byte is not — passthrough inherits whatever real git does.
- `git show --textconv` / filter attributes — passthrough inherits git's behavior; no special handling.
- The JSON-manifest behavior of plain `git show <commit>` (no `:path`) — stays intercepted and unchanged.
- `git cat-file` and other blob-reading porcelain — scoped to `git show <rev>:<path>` only.
- The error-message mangling described below — investigate, but file separately if confirmed (see note).

## Related observation (unconfirmed — do not block this ticket)

The reporter saw a mangled error when the spec didn't resolve:

```
$ git show 73c75449082cc08650ac21b27f423ba507af77da:apps/mcp-servers/packages/api-clients/package.json
... Could not find ref '/Users/mike.lane/workplace/monorepo/73c75449082cc08650ac21b27f423ba507af77dapps/mcp-servers/packages/api-clients/package.json'
```

The cwd appears prepended and `...77da` + `:apps` collapses to `...77dapps`. **This is not reproduced by the current code path:** `src/git/reader.rs::ResolveRefError` (lines 10-24) prints `refspec` verbatim — it neither prepends the cwd nor swallows the colon. The mangled string is most likely a shell/copy-paste artifact, not a shim defect. The dev should confirm whether any shim path actually injects cwd into a refspec; if a real defect is found, open a separate `fix(shim)` ticket. No acceptance criterion here.

## Dependencies

None. Independent of the `git log A..B` diverged-range bug filed alongside this. Related prior shim-parity work for context only (not blockers): #356 (`git -C` bypass), #328 (shim parity epic).

## Edge Cases

- `<rev>` is an annotated tag (`v1.0:path`) — must passthrough and print the blob.
- Path containing a colon, or starting with `-` — the detection rule keys on the **first** `:`; a `-`-leading path is part of the spec, not a flag.
- `<rev>` resolves but `<path>` does not exist — passthrough lets real git emit its native "exists on disk, but not in 'rev'" error.
- `:/<regex>` (commit-message search) — explicitly stays intercepted as a commit-ish (covered by unit test).

## Environment (from report)

git-prism shim at `~/.local/share/git-prism/bin/git`, first on PATH ahead of `/usr/bin/git` (Apple Git-155, git 2.50.1), macOS (darwin 25.5.0), `altana-tech/monorepo` working copy. Reproduced with both branch-relative and SHA-relative blob specs.

mikelane:
## Summary

Fixes `git show <rev>:<path>` (and other object specs) failing through the git-prism shim with `failed to resolve commit ...: was blob while trying to peel to commit`.

`classify_show` took the first non-flag arg as a commit-ish, so `git show origin/main:.lefthook.yml` was classified as `ShowSnapshot` and the handler tried to peel a **blob** to a commit. Native git just prints the file. Any agent/review flow that reads a file at a revision without a checkout hit this.

## Fix

Route object specs to `Passthrough` so real git prints them. The rule mirrors git's object-spec grammar:

```rust
// The ONLY colon-bearing commit-ish is `:/text` (commit-message search).
// Every other colon form is a blob/tree/index spec. Refs cannot contain ':'.
if spec.contains(':') && !spec.starts_with(":/") {
    return Classification::Passthrough;
}
```

Passed through: `<rev>:<path>`, `<rev>:` (tree), `:path` (stage-0 index), `:N:path` (explicit stage), `<rev>:/abs/path`. Still intercepted: `:/regex` (commit search) and plain commit-ish with no colon — those return the JSON manifest as before.

## How the rule got here

The first implementation used a narrower before/after-colon predicate. Adversarial QA proved it still wrongly intercepted index/stage blob specs (`:path`, `:0:path`) and `<rev>:/abs/path` — the same peel-to-commit failure #381 set out to kill. The grammar-based `contains(':') && !starts_with(":/")` rule is both simpler and complete; QA run 2 confirmed no legitimate commit-ish is wrongly passed through.

## Tests

- `src/shim/classify.rs` unit tests (table-driven): branch:path, sha:path, `HEAD:`, `:path`, `:0:path`, `:1/2/3:path`, `HEAD:/abs/path`, `HEAD:dir/file` → Passthrough; `:/regex`, `:/x`, plain sha, `HEAD` → ShowSnapshot. ASCII-colon slicing is panic-safe (verified with multibyte input).
- `bdd/features/shim_wrapper.feature` `@ISSUE-381` scenarios (4): `git show HEAD:<file>` and stage specs are byte-identical to real git with no `handler error` on stderr; plain `git show <sha>` still returns the JSON manifest.
- Full suite: 1110 passing; `fmt`/`clippy --all-targets -D warnings` clean.

## Note
The "mangled error message" secondary observation in the issue is **not reproduced** by current code — `ResolveRefError` prints the refspec verbatim (no cwd-prepend, no colon-swallow), so it's most likely a shell/copy artifact. No change made; left as a note on the issue.

Closes #381


mikelane (commented):
## PR #384 Review — `fix(shim): pass git show <rev>:<path> blob specs through to real git`

### Summary

This is a clean, well-scoped bug fix. The predicate `spec.contains(':') && !spec.starts_with(":/")` correctly partitions all known git object-spec forms, and the implementation makes no production code changes beyond `classify_show` in `src/shim/classify.rs`. Every quality gate passes.

---

The bug was real and the diagnosis was correct. Before this fix, `classify_show` took the first non-flag arg and passed it wholesale to `ShowSnapshot { sha: spec }`, which is fed into `peel_to_commit`. Git's object spec grammar has two completely distinct namespaces: commit-ish (no colon, or `:/regex` commit-message search) and object-specs (colon-bearing: `<rev>:<path>`, `:<path>`, `:N:<path>`). The old code conflated them. The fix is correct.

---

### Phase 1 — Quality Tools (Strict Mode)

- **`cargo clippy --all-targets -- -D warnings`**: PASS (only pre-existing vendored Kotlin C scanner warning, unrelated to this PR)
- **`cargo fmt --check`**: PASS — no output
- **`cargo test --bin git-prism`**: PASS — 1110 tests, 0 failures, 15.6s
- **`cargo test --bin git-prism -- classify`**: PASS — 82 tests, 0 failures
- **BDD `@ISSUE-381`**: PASS — all 4 scenarios green (verified from output)

---

No changes to `.pre-commit-config.yaml`, `pyproject.toml`, linter configs, or any suppression list. Nothing bypassed.

---

### Phase 3 — Predicate Correctness (Core of this Review)

The central question is whether `spec.contains(':') && !spec.starts_with(":/")` correctly partitions git's object-spec grammar. Independent verification against the full grammar:

**Correctly routed to Passthrough (blob/tree/index specs):**
- `HEAD:file.txt` — rev-qualified blob
- `HEAD:` — rev-qualified tree root
- `origin/main:.lefthook.yml` — remote-qualified blob
- `:staged.txt` — stage-0 index blob
- `:0:file`, `:1:file`, `:2:file`, `:3:file` — explicit-stage index blobs
- `abc123:path/to/file` — SHA-qualified blob
- `HEAD:/absolute/path` — rev-qualified absolute path blob
- `v1.0.0:README.md`, `refs/heads/main:src/lib.rs` — tag/ref-qualified blobs

**Correctly routed to ShowSnapshot (commit-ish):**
- `:/fix typo`, `:/regex` — commit-message search (starts with `:/`)
- `abc1234`, `HEAD`, `HEAD~1`, `v1.0.0`, `main`, `refs/heads/feature` — no colon at all

The comment in the code (lines 113-121) correctly documents all these forms including the subtle point that refs cannot contain `:`. This is accurate — git ref names are prohibited from containing `:` by `git-check-ref-format` rules, making the single predicate sufficient.

**One edge case I probed and confirmed safe:** `HEAD:/abs/path` (absolute path via `<rev>:/path` form). This contains `:` and does NOT start with `:/` (it starts with `H`), so it routes to Passthrough. Correct — this is a valid blob lookup.

**The only form that would be a false positive is `:/regex`**, which starts with `:/` and is correctly intercepted as ShowSnapshot. Confirmed.

---

The block comment at `classify.rs:113-121` explains the WHY and documents the grammar. Each bullet in the comment maps to a distinct form with its name (`rev-qualified blob`, `stage-0 index blob`, etc.). The `SAFETY` comment pattern for the `--` separator limitation in `has_scripted_output_flag` is already present in the existing code and is appropriate.

---

### Phase 5 — Test Quality

**Unit tests (classify.rs):** 82 tests total, 16 new for this issue. Tests:
- Assert the exact `Classification` variant (not just "is not ShowSnapshot")
- Use `assert_eq!` with the full enum value including the carried `sha` field
- Test the positive case (blob spec → Passthrough), negative case (plain SHA → ShowSnapshot), and the tricky boundary (commit-message search `:/` → ShowSnapshot)
- Names follow the `it_passes_through_` / `it_classifies_` convention used throughout

All tests go through the public `classify()` API — no private method access.

**BDD scenarios (4 new `@ISSUE-381` scenarios):**
- Scenario 1: `git show HEAD:README.md` — real file content returned, no handler error marker
- Scenario 2: plain `git show HEAD` — still JSON with `snapshots` key (regression guard)
- Scenario 3: `git show :staged.txt` — stage-0 index blob
- Scenario 4: `git show :0:staged.txt` — explicit-stage index blob

The adversarial QA fixture (`step_fixture_repo_with_staged_file`) properly stages a file without committing it, making the stage-0 tests meaningful. The `step_output_no_handler_error` assertion checks stderr for `git-prism shim: handler error` — this is the exact string that would appear on the pre-fix code path, making it a precise regression guard.

---

### Phase 6 — Architecture

Single-responsibility change: one predicate in one function, one file. No new abstraction layers. The approach of using `contains`/`starts_with` rather than regex or character indexing eliminates any panic risk from ASCII colon slicing that might have existed with an index-based approach.

---

No I/O, no external calls, no user-controlled data reaches any execution path. The classifier is pure — `&[&str] → Classification`. No security surface.

---

- **Merge base**: `a7b757f` — branch is current off origin/main
- **Files changed on main since divergence**: none
- **Overlap with PR files**: none (PR is pure addition to an already-merged state)
- **Deletion volume**: 0 lines deleted (pure additions: +195, −3 pre-existing lines in CHANGELOG and feature file formatting)
- **Conclusion**: no stale-branch regression risk

---

### Praise

The comment block at lines 111-122 is excellent — it enumerates every known colon-bearing form, names each one, and explains the one exception (`:/text`). This is exactly the kind of WHY comment that the codebase's style requires. Future readers will understand immediately why the predicate is correct without having to look up git's object-spec grammar themselves.

The fixture setup in `step_fixture_repo_with_staged_file` is worth calling out: staging a file without committing creates a genuine stage-0 object in the index, so `git show :staged.txt` and `git show :0:staged.txt` actually round-trip through real git plumbing. No mock.

---

### Decision: APPROVED

0 CRITICAL, 0 HIGH, 0 MEDIUM, 0 LOW findings. All quality tools pass. Predicate is correct. Tests are behavioral and complete. No regressions. Ready to merge.

